PR #2, opened 2019-11-25
Remove redundant final modifier and redundant variables

Gaur4vGaur:
Below is the gist of changes:

- Remove redundant final modifier in front of private methods. In java private already implies that a subclass can not override a method, declaring a private method to be final is redundant

- I found few redundant variables though out the code. I have performed a bit of code clean up. Let me know if you do not agree with any of the change

I have ran `make` command and made sure that the build is fine and unittests pass.

PS: IMO, it would be easier it you review each commit. Even though files changed are 34 but changes are really simple.

### DIFF
--- a/benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/BenchmarkDriver.java
+++ b/benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/BenchmarkDriver.java
@@ -71,17 +71,14 @@ public abstract class BenchmarkDriver {
     }
 
     private boolean skipGC() {
-        if (this instanceof BitsetTornado) {
-            return true;
-        }
-        return false;
+        return this instanceof BitsetTornado;
     }
 
     public void benchmark() {
 
         setUp();
 
-        validResult = (VALIDATE) ? validate() : true;
+        validResult = (!VALIDATE) || validate();
 
         int size = toIntExact(iterations);
 
--- a/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/OCLDevice.java
+++ b/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/OCLDevice.java
@@ -132,7 +132,7 @@ public class OCLDevice extends TornadoLogger implements TornadoTargetDevice {
         buffer.clear();
         clGetDeviceInfo(id, OCLDeviceInfo.CL_DEVICE_AVAILABLE.getValue(), buffer.array());
 
-        return (buffer.getInt() == 1) ? true : false;
+        return buffer.getInt() == 1;
     }
 
     @Override
@@ -407,7 +407,7 @@ public class OCLDevice extends TornadoLogger implements TornadoTargetDevice {
         clGetDeviceInfo(id, OCLDeviceInfo.CL_DEVICE_ENDIAN_LITTLE.getValue(), buffer.array());
         deviceEndianLittle = buffer.getInt();
 
-        return deviceEndianLittle == CL_TRUE ? true : false;
+        return deviceEndianLittle == CL_TRUE;
     }
 
     public int getWordSize() {
--- a/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/lir/OCLKind.java
+++ b/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/lir/OCLKind.java
@@ -327,18 +327,12 @@ public enum OCLKind implements PlatformKind {
     }
 
     public boolean isInteger() {
-        if (kind == ILLEGAL || isFloating()) {
-            return false;
-        }
-        return true;
+        return kind != ILLEGAL && !isFloating();
     }
 
     public boolean isFloating() {
         // TODO are vectors integers?
-        if (kind == FLOAT || kind == DOUBLE) {
-            return true;
-        }
-        return false;
+        return kind == FLOAT || kind == DOUBLE;
     }
 
     public boolean isVector() {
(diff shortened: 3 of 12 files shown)


benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/blackscholes/BlackScholes.java
```diff
@@ -1,218 +1,216 @@
...
+        final float oneBySqrt2pi = 0.398942280f;
+
+        final float absX = Math.abs(X);
+        final float t = one / (one + (temp4 * absX));
+
+        final float y = (one - (oneBySqrt2pi * TornadoMath.exp((-X * X) / two) * t * (c1 + (t * (c2 + (t * (c3 + (t * (c4 + (t * c5))))))))));
+
+        return (X < zero) ? (one - y) : y;
+    }
+
```
Gaur4vGaur: This is the only file I have changed, but git is highlighting the entire file because I am using Ubuntu. IntelliJ on ubuntu uses LF to end the file.
There are two ways to represent a line ending in a source code file, CRLF or just LF. Windows tends to default to CRLF, but the rest of the world uses just LF, so Git/Programming standards in general prefer LF.

benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/ComputeKernels.java
```diff
@@ -249,15 +246,12 @@ public static void mandelbrot(int size, short[] output) {
...
-            int indexIDX = i;
             for (@Parallel int j = 0; j < size; j++) {
 
-                int indexJDX = j;
-
                 float Zr = 0.0f;
                 float Zi = 0.0f;
-                float Cr = (1 * indexJDX * space - 1.5f);
-                float Ci = (1 * indexIDX * space - 1.0f);
+                float Cr = (1 * j * space - 1.5f);
```
jjfumero: Could you revert to `indexJDX` and `indexIDX`? This kernel comes from AMD OpenCL examples, so the indexing is to indicate the levels of parallelism.

benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/blackscholes/BlackScholes.java
jjfumero: I still don't get why this is like a new file. We also use Ubuntu and Red-Hat based. Are you using the TornadoVM code formatter? 

We use the Eclipse code formatter plugin, and the XML is in the TornadoVM repository.  Full explanation here:

https://github.com/beehive-lab/TornadoVM/blob/master/assembly/src/docs/3_INTELLIJ.md#required-plugins

examples/src/main/java/uk/ac/manchester/tornado/examples/compute/Mandelbrot.java
```diff
@@ -51,15 +51,12 @@ public MandelbrotImage() {
...
-                int indexIDX = i;
                 for (int j = 0; j < size; j++) {
 
-                    int indexJDX = j;
-
                     float Zr = 0.0f;
                     float Zi = 0.0f;
-                    float Cr = (1 * indexJDX * space - 1.5f);
-                    float Ci = (1 * indexIDX * space - 1.0f);
+                    float Cr = (1 * j * space - 1.5f);
```
jjfumero: Same comment that before. We prefer to keep this as `indexJDK` and `indexIDX`.

examples/src/main/java/uk/ac/manchester/tornado/examples/dynamic/MandelbrotFPGA.java
```diff
@@ -46,15 +46,12 @@ private static void mandelbrotTornado(int size, short[] output) {
...
-            int indexIDX = i;
             for (int j = 0; j < size; j++) {
 
-                int indexJDX = j;
-
                 float Zr = 0.0f;
                 float Zi = 0.0f;
-                float Cr = (1 * indexJDX * space - 1.5f);
-                float Ci = (1 * indexIDX * space - 1.0f);
+                float Cr = (1 * j * space - 1.5f);
```
jjfumero: Same here

runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskSchedule.java
```diff
@@ -899,12 +899,11 @@ private void runScheduleWithParallelProfiler(Policy policy) {
 
         // One additional threads is reserved for sequential CPU execution
         final int numThreads = numDevices + 1;
-        final int indexSequential = numDevices;
         Thread[] threads = new Thread[numThreads];
         long[] totalTimers = new long[numThreads];
 
         // Last Thread runs the sequential code
-        runParallelSequential(policy, threads, indexSequential, timer, totalTimers);
+        runParallelSequential(policy, threads, numDevices, timer, totalTimers);
```
jjfumero: This is not the number of devices. It is the index, within the array of all devices in which we keep a reference to the sequential execution. Please revert to `indexSequential`

benchmarks/src/main/java/uk/ac/manchester/tornado/benchmarks/blackscholes/BlackScholes.java
Gaur4vGaur: Sure, I will check it again. I am using IntelliJ on Ubuntu.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskSchedule.java
Gaur4vGaur: Sure, I will revert it but when I was going through the code, it seems we reassign `numDevices` to `indexSequential`.
`final int indexSequential = numDevices;`

stratika: > Sure, I will revert it but when I was going through the code, it seems we reassign `numDevices` to `indexSequential`.
> `final int indexSequential = numDevices;`

Hi @Gaur4vGaur, the variable `indexSequential` is the index of the sequential thread that resides in the matrix of threads. If we have _n_ devices the number of threads is _n+1_, because it includes the sequential execution. In our case the last index (_n_) is used for the sequential thread, as the first _n_ indexes (_0-(n-1)_) are used for the threads that are associated with the _n_ devices.

stratika:
I agree with the comments from @jjfumero. I have just had a look at the code changes. Once the comments are resolved, we can have a final look to approve the PR.

Issue #12, opened 2020-05-05
Regulate the amount of memory depending on the current target device present in the system for the unittests


jjfumero:
Some of the unittests consume a lot of memory (on purpose). Regulate the amount of memory depending on the current target device that is present in the system.

By default, TornadoVM allocates a heap of 1GB on the target device. Therefore, it assumes the device has at least 1GB. Regulate the space to be allocated if the memory available on the target device is less than the default value (e.g., by allocating 512MB). 

The device heap is controlled using the following flag:
`-Dtornado.heap.allocation=1GB`

jjfumero:
Merged in #18

PR #17, opened 2020-05-30
#12 regulate amount of allocated device heap depending on the device

dalexandrov:
I've examined the tests and it looks like TestBatches is the only candidate for this ticket.

I've added some checks before every test to watch for the allocated heap (-Dtornado.heap.allocation flag).
If there is not enough memory even for one chunk - UnsupportedConfigurationException is thrown , thus test will end up with "Unsupported Configuration" status.

I hope i've got the ticket #12 right :)

### DIFF
--- a/unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
+++ b/unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
@@ -1,33 +1,34 @@
 /*
  * Copyright (c) 2013-2020, APT Group, Department of Computer Science,
  * The University of Manchester.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *    http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- * 
+ *
  */
 
 package uk.ac.manchester.tornado.unittests.batches;
 
-import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+import uk.ac.manchester.tornado.api.TaskSchedule;
+import uk.ac.manchester.tornado.api.annotations.Parallel;
+import uk.ac.manchester.tornado.unittests.common.TornadoTestBase;
+import uk.ac.manchester.tornado.unittests.tools.Exceptions.UnsupportedConfigurationException;
 
 import java.util.Random;
 import java.util.stream.IntStream;
 
-import org.junit.Test;
-
-import uk.ac.manchester.tornado.api.TaskSchedule;
-import uk.ac.manchester.tornado.api.annotations.Parallel;
-import uk.ac.manchester.tornado.unittests.common.TornadoTestBase;
+import static org.junit.Assert.assertEquals;
+import static uk.ac.manchester.tornado.unittests.virtualization.TestsVirtualLayer.getTornadoRuntime;
 
 public class TestBatches extends TornadoTestBase {
 
@@ -76,8 +77,19 @@ public class TestBatches extends TornadoTestBase {
     @Test
     public void test100MB() {
 
+        long maxAllocMemory = getTornadoRuntime().getDefaultDevice().getDeviceContext().getMemoryManager().getHeapSize();
+
+        //check if there is enough memory for at least one chunk
+        if (maxAllocMemory < 100 * 1024 * 1024) {
+            throw new UnsupportedConfigurationException("Not enough memory to run the test");
+        }
+
         // Fill 800MB of float array
         int size = 200000000;
+        // or as much as we can
+        if (size * 4 > maxAllocMemory) {
+            size = (int) ((maxAllocMemory / 4 / 2) * 0.9);
+        }
         float[] arrayA = new float[size];
         float[] arrayB = new float[size];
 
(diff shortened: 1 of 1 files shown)


unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
```diff
@@ -76,8 +77,19 @@ public static void compute(short[] arrayA, short[] arrayB, short[] arrayC) {
     @Test
     public void test100MB() {
 
+        long maxAllocMemory = getTornadoRuntime().getDefaultDevice().getDeviceContext().getMemoryManager().getHeapSize();
+
+        //check if there is enough memory for at least one chunk
+        if (maxAllocMemory < 100 * 1024 * 1024) {
+            throw new UnsupportedConfigurationException("Not enough memory to run the test");
+        }
+
```
jjfumero, line 86: Create a method for this - You can pass the MBs required for each test. 

The method can be private and located in this class.

kotselidis: We can also make the MBs and GBs constants.

unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
```diff
@@ -76,8 +77,19 @@ public static void compute(short[] arrayA, short[] arrayB, short[] arrayC) {
...
         // Fill 800MB of float array
         int size = 200000000;
+        // or as much as we can
+        if (size * 4 > maxAllocMemory) {
+            size = (int) ((maxAllocMemory / 4 / 2) * 0.9);
```
kotselidis, line 91: I miss a bit of context here, what are the numbers we divide maxAllocMemory?

unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
dalexandrov: Yep, I put it this way, since it is a different setup for each test, but I will extract it

unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
dalexandrov: I had to do like this, since in some tests we fill one array, in some tests two, in some three.. And we have heap. I always multiply by 0.9 not to fill all the memory.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskSchedule.java
```diff
@@ -25,57 +25,15 @@
...
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task15;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task2;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task3;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task4;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task5;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task6;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task7;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task8;
-import uk.ac.manchester.tornado.api.common.TornadoFunctions.Task9;
+import uk.ac.manchester.tornado.api.common.*;
```
jjfumero, line 35: Use "Single class import"

runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskSchedule.java
```diff
@@ -104,65 +54,80 @@
...
+    private static final int PERFORMANCE_WARMUP = 3;
+    private final static boolean TIME_IN_NANOSECONDS = Tornado.TIME_IN_NANOSECONDS;
+    private static final String TASK_SCHEDULE_PREFIX = "XXX";
+    private static final ConcurrentHashMap<Policy, ConcurrentHashMap<String, HistoryTable>> executionHistoryPolicy = new ConcurrentHashMap<>();
+    private static final int HISTORY_POINTS_PREDICTION = 5;
+    private static final boolean USE_GLOBAL_TASK_CACHE = false;
+    private static ConcurrentHashMap<Integer, TaskSchedule> globalTaskScheduleIndex = new ConcurrentHashMap<>();
+    private static int baseGlobalIndex = 0;
+    private static AtomicInteger offsetGlobalIndex = new AtomicInteger(0);
+    private final CompileInfo COMPILE_ONLY = new CompileInfo(true, false);
```
jjfumero, line 96: These fields are not related to dynamic reconfiguration. Why is this reordering?

runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskSchedule.java
```diff
@@ -626,6 +640,12 @@ private AbstractTaskGraph analyzeSkeletonAndRun() {
 
     @Override
     public AbstractTaskGraph schedule() {
+
+        if (getTornadoRuntime().getNumDrivers() < 1 || getTornadoRuntime().getDriver(0).getNumPlatforms() < 1) {
+            runSequential();
```
jjfumero, line 645: Introduce a warning that the code was "deoptimize" and TornadoVM will execute the Java sequential code.

unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java
```diff
@@ -37,6 +38,8 @@ public static void compute(float[] array) {
         }
     }
 
+    ;
```
jjfumero, line 41: Remove this semicolon

Issue #21, opened 2020-06-25
[Support] Parallelize loops with constant values as loop-bounds 


jjfumero:
Currently, TornadoVM only parallelizes loops in which loop bounds are taken as array-length.
Add support for constant-values. 

```java
for (@Parallel i = 0; i < CONSTANT; i++) {
   ... 
}
```
To support this, the compiler phases `TornadoAutoParalleliser` and `TornadoParallelScheduler` need to be updated. 
This change also influences the batch processing, in which loop-bounds need to be updated according to the batch size.
This change will also provide a set of unittests to check this feature.

jjfumero:
The issue was that for loops such as:

```java
for (@Parallel int i = 0; i <= n; i++) {
      a[i] = 10;
}
```

The constant value is evaluated and rescheduled. The `TornadoTaskSpecialization` expects the loop bounds attached to the `ParallelRangeNode` instead of the conditional node. 

The proposed change is as follows:

![Screenshot from 2020-06-25 12-29-46](https://user-images.githubusercontent.com/5602565/85703480-b04cb200-b6df-11ea-93c1-9d165b86a205.png)


PR coming soon.

jjfumero:
Fixed in #22

PR #22, opened 2020-06-25
[TP-21] Parallelization of loops with constant loop-bounds

jjfumero:
The issue was that for expressions such as "<=", the `TornadoTaskSpecialization` phase was inserting the constant node outside the parallel range node.

```java
public static void forConstant01(int[] a, final int n) {
        for (@Parallel int i = 0; i < n; i++) {
            a[i] = 10;
        }
   }
```

The solution is that, for the cases in which the Parallel Range contains a PhiNode as value, we can inspect the condition of the loop-bound. If a constant node is found, we can perform node swapping between these two.

Github issue #21

#### How to test?

```bash
$ tornado-test.py -V -pk --debug  uk.ac.manchester.tornado.unittests.loops.TestLoops#testForConstant02 

task info: s0.t0
	platform          : NVIDIA CUDA
	device            : GeForce GTX 1050 CL_DEVICE_TYPE_GPU (available)
	dims              : 1
	global work offset: [0]
	global work size  : [257]
	local  work size  : [257, 1, 1]
```

### DIFF
--- a/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoTaskSpecialisation.java
+++ b/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoTaskSpecialisation.java
@@ -25,23 +25,21 @@
  */
 package uk.ac.manchester.tornado.drivers.opencl.graal.phases;
 
-import static uk.ac.manchester.tornado.api.exceptions.TornadoInternalError.unimplemented;
-import static uk.ac.manchester.tornado.runtime.TornadoCoreRuntime.getDebugContext;
-
-import java.lang.reflect.Array;
-import java.lang.reflect.Field;
-import java.lang.reflect.Modifier;
-
+import jdk.vm.ci.meta.JavaKind;
+import jdk.vm.ci.meta.ResolvedJavaField;
 import org.graalvm.compiler.core.common.type.ObjectStamp;
 import org.graalvm.compiler.debug.DebugContext;
 import org.graalvm.compiler.graph.Graph.Mark;
 import org.graalvm.compiler.graph.Node;
+import org.graalvm.compiler.graph.iterators.NodeIterable;
 import org.graalvm.compiler.nodes.ConstantNode;
 import org.graalvm.compiler.nodes.LogicConstantNode;
 import org.graalvm.compiler.nodes.NodeView;
 import org.graalvm.compiler.nodes.ParameterNode;
+import org.graalvm.compiler.nodes.PhiNode;
 import org.graalvm.compiler.nodes.PiNode;
 import org.graalvm.compiler.nodes.StructuredGraph;
+import org.graalvm.compiler.nodes.calc.IntegerLessThanNode;
 import org.graalvm.compiler.nodes.calc.IsNullNode;
 import org.graalvm.compiler.nodes.java.ArrayLengthNode;
 import org.graalvm.compiler.nodes.java.LoadFieldNode;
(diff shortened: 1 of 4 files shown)


drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoTaskSpecialisation.java
```diff
@@ -271,29 +272,49 @@ protected void run(StructuredGraph graph, TornadoHighTierContext context) {
...
 
-            hasWork = (lastNodeCount != graph.getNodeCount() || graph.getNewNodes(mark).isNotEmpty()) // ||
-                                                                                                      // hasGuardingPiNodes)
-                    && (iterations < MAX_ITERATIONS);
+            hasWork = (lastNodeCount != graph.getNodeCount() || graph.getNewNodes(mark).isNotEmpty()) && (iterations < MAX_ITERATIONS);
             lastNodeCount = graph.getNodeCount();
             iterations++;
         }
 
+        graph.getNodes().filter(ParallelRangeNode.class).forEach(range -> {
```
mikepapadim, line 282: In cases of 2D parallel loops with both having `constants` as bounds. Does it change the bounds correctly or copies the outter bound to the inner range?

jjfumero: +1 I will add a combination of those.

jjfumero:
2D tests added:

```bash
$ tornado-test.py -V -pk --debug -J"-Ds0.t0.device=0:0" uk.ac.manchester.tornado.unittests.loops.TestLoops#testForConstant04
$ tornado-test.py -V -pk --debug -J"-Ds0.t0.device=0:0" uk.ac.manchester.tornado.unittests.loops.TestLoops#testForConstant05
```

unittests/src/main/java/uk/ac/manchester/tornado/unittests/loops/TestLoops.java
```diff
@@ -26,10 +26,117 @@
...
+        for (@Parallel int i = 0; i <= n; i++) {
+            for (@Parallel int j = 0; j <= n; j++) {
+                m.set(i, j, 10);
+            }
+        }
+    }
+
+    @Test
+    public void testForConstant04() {
+        int size = 255;
```
stratika, line 104: We can extend the unit-tests with one test for 2 dimensional array in which each dimension has different size, I tested it and it seems to be working.
e.g. `Matrix2DFloat m = new Matrix2DFloat(256, 64);`

```
task info: s0.t0
	platform          : Intel(R) OpenCL HD Graphics
	device            : Intel(R) Gen9 HD Graphics NEO CL_DEVICE_TYPE_GPU (available)
	dims              : 2
	global work offset: [0, 0]
	global work size  : [64, 256]
	local  work size  : [16, 16, 1]
```

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoTaskSpecialisation.java
```diff
@@ -25,42 +25,45 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+
+import static uk.ac.manchester.tornado.api.exceptions.TornadoInternalError.unimplemented;
+import static uk.ac.manchester.tornado.runtime.TornadoCoreRuntime.getDebugContext;
+
 public class TornadoTaskSpecialisation extends BasePhase<TornadoHighTierContext> {
 
-    private static final int MAX_ITERATIONS = 10;
+    private static final int MAX_ITERATIONS = 15;
```
stratika, line 66: I am wondering why did we increase the number of maximum iterations.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/nodes/AbstractParallelNode.java
```diff
@@ -27,6 +27,7 @@
 import org.graalvm.compiler.core.common.type.StampFactory;
 import org.graalvm.compiler.graph.NodeClass;
 import org.graalvm.compiler.nodeinfo.NodeInfo;
+import org.graalvm.compiler.nodes.ConstantNode;
```
stratika, line 30: this package is not used

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/nodes/ParallelRangeNode.java
```diff
@@ -28,15 +28,18 @@
 import org.graalvm.compiler.graph.NodeClass;
 import org.graalvm.compiler.nodeinfo.InputType;
 import org.graalvm.compiler.nodeinfo.NodeInfo;
+import org.graalvm.compiler.nodes.ConstantNode;
```
stratika, line 31: this package is not used.

unittests/src/main/java/uk/ac/manchester/tornado/unittests/loops/TestLoops.java
```diff
@@ -26,10 +26,117 @@
...
+        Matrix2DFloat m = new Matrix2DFloat(size, size);
+        new TaskSchedule("s0") //
+                .task("t0", TestLoops::forConstant04, m, size) //
+                .streamOut(m) //
+                .execute();
+        for (int i = 0; i < m.M(); i++) {
```
stratika, line 110: This will run for 255 iterations, while the loop in method `forConstant04 ` line 95 will run for 256 iterations.
Am I missing something? I think it should be here the same condition as in the method (i.e. less equal).

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoTaskSpecialisation.java
jjfumero: Just to have the opportunity to further optimize the code.

unittests/src/main/java/uk/ac/manchester/tornado/unittests/loops/TestLoops.java
jjfumero: Good catch. We don't get the exceptions on GPUs. I fixed the test.

Issue #24, opened 2020-06-30
MonteCarloDynamic kernel failing on the Xilinx FPGA


stratika:
**Describe the bug**
I noticed a problem with the Montecarlo kernel in the dynamic package for all the sizes, when executing on the Xilinx KCU1500 FPGA. There is no error in the compilation, but the kernel does not finish and it causes failures at the driver level regarding the dma. The problem seems like this:
```
[  815.440478] xocl:engine_status_dump: SG engine 0-H2C1-MM status: 0x00000000:
[  815.440480] xocl:engine_status_dump: SG engine 0-H2C0-MM status: 0x00000001: BUSY
[  815.440483] xocl:transfer_abort: abort transfer 0x000000009584ae00, desc 11, engine desc queued 0.
[  815.440487] xocl:transfer_abort: abort transfer 0x00000000d2360335, desc 1, engine desc queued 0.
[  815.440505] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: DMA failed, Dumping SG Page Table
[  815.440508] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: DMA failed, Dumping SG Page Table
[  815.440516] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 0, 0xf3ce7c000
[  815.440521] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 1, 0xf3d800000
[  815.440526] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 2, 0xf3d400000
[  815.440531] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 3, 0xf3f000000
[  815.440536] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 4, 0xf7d000000
[  815.440540] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 5, 0xf4f800000
[  815.440545] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 6, 0xf54800000
[  815.440550] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 7, 0xf60400000
[  815.440554] xocl_mm_xdma mm_dma.v5.u.256: xdma_migrate_bo: 8, 0xf61c00000
[...]

PR #28, opened 2020-07-02
[fix] FMA optimizations can be disabled

jjfumero:
For Xilinx FPGAs running OpenCL 1.0 we need to disable FMA optimizations.
Use the flag `-Dtornado.enable.fma=False`

This flag is enabled by default.

Github issue #24 (Public)

### DIFF
--- a/assembly/src/docs/6_TORNADO_FLAGS.md
+++ b/assembly/src/docs/6_TORNADO_FLAGS.md
@@ -63,3 +63,9 @@ It enables profiler information such as `COPY_IN`, `COPY_OUT`, compilation time,
 * `-Dtornado.opencl.compiler.options=LIST_OF_OPTIONS`:  
 It allows to pass the compile options specified by the OpenCL ``CLBuildProgram`` [specification](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clBuildProgram.html) to TornadoVM at runtime. By default it doesn't enable any. 
 
+
+##### Optimizations
+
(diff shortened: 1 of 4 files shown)


stratika:
I confirm this works. Thanks Juan.

assembly/src/docs/6_TORNADO_FLAGS.md
```diff
@@ -63,3 +63,9 @@ It enables profiler information such as `COPY_IN`, `COPY_OUT`, compilation time,
 * `-Dtornado.opencl.compiler.options=LIST_OF_OPTIONS`:  
 It allows to pass the compile options specified by the OpenCL ``CLBuildProgram`` [specification](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clBuildProgram.html) to TornadoVM at runtime. By default it doesn't enable any. 
 
+
+##### Optimizations
+
+* `-Dtornado.enable.fma=True`:  
+It enables Fused-Multiply-Add optimizations. This option is enabled bu default. However, for some platforms, such as Xilinx with OpenCL 1.0, this option must be disabled.
```
stratika, line 70: bu -> by. I am not sure if it is generic to all Xilinx devices. I think it is a problem concerning the runtime so perhaps this is more accurate. "platforms, such as the Xilinx FPGAs using the SDAccel 2018.2, this option must be disabled as it causes runtime errors."

jjfumero: Done

Issue #25, opened 2020-07-01
Return flatten arrays from TornadoVM custom types


jjfumero:
Return the internal primitive array to the user for the custom types such as `ImageByte2`, `ImageByte3`, etc.

PR #29, opened 2020-07-02
[dev] TornadoVM collection types return backing array

jjfumero:
This PR addresses  Github issue #25 

This patch includes:
* Get backing array for wrap-types (e.g., `ImageByte3#getArray()`) 
* Clean-up TornadoVM Wrapper-APIs 

All unittests are passing as well as `kfusion-tornadovm`.

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Float6.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Float6.java
@@ -41,23 +41,11 @@
  */
 package uk.ac.manchester.tornado.api.collections.types;
 
-import static java.lang.Float.MAX_VALUE;
-import static java.lang.Float.MIN_VALUE;
-import static java.lang.String.format;
-import static java.nio.FloatBuffer.wrap;
-import static uk.ac.manchester.tornado.api.collections.types.FloatOps.fmt6;
-
 import java.nio.FloatBuffer;
 
 import uk.ac.manchester.tornado.api.collections.math.TornadoMath;
 import uk.ac.manchester.tornado.api.type.annotations.Payload;
 
-/**
- * Class that represents a vector of 3x floats e.g. <float,float,float>
- *
- * @author jamesclarkson
- *
- */
 public final class Float6 implements PrimitiveStorage<FloatBuffer> {
 
     public static final Class<Float6> TYPE = Float6.class;
@@ -65,7 +53,8 @@ public final class Float6 implements PrimitiveStorage<FloatBuffer> {
     /**
      * backing array
      */
-    @Payload final protected float[] storage;
+    @Payload
+    final protected float[] storage;
 
     /**
      * number of elements in the storage
@@ -88,6 +77,10 @@ public final class Float6 implements PrimitiveStorage<FloatBuffer> {
         setS3(s3);
         setS4(s4);
         setS5(s5);
+    }
+
+    public float[] getArray() {
+        return storage;
     }
 
     public void set(Float6 value) {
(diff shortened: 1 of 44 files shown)


tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Matrix2DFloat4.java
```diff
@@ -90,15 +90,19 @@ public Matrix2DFloat4(int width, int height, float[] array) {
...
      * @param width
+     *            number of columns
+     * @param height
      *            number of rows
      */
     public Matrix2DFloat4(int width, int height) {
         this(width, height, new float[width * height * VECTOR_ELEMENTS]);
     }
 
+    public float[] getFlattenMatrix() {
```
kotselidis, line 102: getFlatten or getFlattened?

kotselidis:
Looks good @jjfumero 
Couple points for discussion:
1) Depending on the class type we use either getStorage or getFlattenArray. Shall we use the same name convention for all?
2) I think we should start adding proper JavaDoc to our API so we can compile the HTML API

jjfumero:
@kotselidis , 
  `getStorage` is an old API. I will remove that.  I call `getArray` for types that are wraps-arrays, and `getFlattenArray` for types that wrap a matrix type. We can call all of them `getArray`. 

What do you think?

Concerning the Javadoc, I will open a task to work on that on a separate issue.

kotselidis:
I think the name convention you propose is good.

Regarding Javadoc, yes this shall be a different Issue

tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Matrix2DFloat4.java
jjfumero: Let's go with `getFlattenedMatrix`

tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Float6.java
```diff
@@ -175,15 +168,15 @@ public Float6 duplicate() {
...
     }
 
     @Override
     public String toString() {
-        return toString(fmt6);
+        return toString(FloatOps.fmt6);
     }
 
-    public static final Float6 loadFromArray(final float[] array, int index) {
+    public static Float6 loadFromArray(final float[] array, int index) {
```
stratika, line 179: This change is consistent across all classes. I have this question:

If the return value is a final Object, then why do we remove it?

jjfumero:
@stratika ,  I call `getArray` for types that are wraps-arrays, and `getFlattenArray` for types that wrap a matrix type. See previous comments

stratika:
> @stratika , I call `getArray` for types that are wraps-arrays, and `getFlattenArray` for types that wrap a matrix type. See previous comments

but the `getFlattenedMatrix` is used.

jjfumero:
Sorry, I meant `getFlattenMatrix` Shall we go for `FlattenedArray` instead?

stratika:
> Sorry, I meant `getFlattenMatrix`. Shall we go for `FlattenedArray` instead?

I do not mind. I pointed out because earlier it is mentioned as Array. I think both are correct. 
Perhaps `FlattenedArray` is more precise?

tornado-api/src/main/java/uk/ac/manchester/tornado/api/collections/types/Float6.java
jjfumero: We don't need the final since it is a `static` method.

jjfumero:
Changes done

PR #44, opened 2020-12-08
add clWaitForEvents to data_movers.c to avoid possible GC/OpenCL race

grfrost:
We always force a clWaitForEvent with enqueued data moves (read and write) whilst we are pinning the array. 

After the unpin GC is free to move the buffer, potentially before written (enqueueWrite), or after buffer has moved (enqueueRead)

### DIFF
--- a/drivers/opencl-jni/src/main/cpp/source/data_movers.c
+++ b/drivers/opencl-jni/src/main/cpp/source/data_movers.c
@@ -85,7 +85,9 @@ CREATE_ARRAY(Java_uk_ac_manchester_tornado_drivers_opencl_OCLContext, D, double)
             OPENCL_SOFT_ERROR("clEnqueueWriteBuffer (" #TYPE  ")", status, -1); \
             if(PRINT_DATA_TIMES) { \
                 long writeTime = getTimeEvent(event); \
-                printf("H2D time: %ld (ns) \n", writeTime); \
+                printf("H2D time: %ld (ns) \n", writeTime); /* clWaitForEvents call a side effect of this call so safe to not wait */ \
+            }else{ \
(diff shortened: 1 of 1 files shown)


jjfumero:
This will force the JNI code to be blocking for copies (in & out). We can force `clEnqueueReadBuffer` and `clEnqueueWriteBuffer` to be blocking instead in the JNI code, and avoid the `clWaitForEvents`.

Additionally, IMO since we are currently using a single command queue per device, this would not make too much of a difference in performance. 

@grfrost , what do you think?

grfrost:
@jjfumero we could absolutely block (by passing true arg to clEnqueueRead/Write). 
I was merely taking the lead from the current code.  One patch of the if(PRINT_DATA) branch calls getTime (or some such).  A side effect of which is that we call clWaitfOrEvent().  I merely balanced the if with an else. 
Let me know which you prefer.

jjfumero:
Ok. sure. Let's take this path (adding `clWaitForEvents`). Since I am developing a new design for events. it will be easy to modify this way. 

Thanks @grfrost

jjfumero (approved):

jjfumero:
Out of curiosity, did the error with ROCm driver go away with this patch?

grfrost:
> Out of curiosity, did the error with ROCm driver go away with this patch?

They went from 'always' to 'intermittent' on my newer device.  Always still on the older device.  There is something weird in this codebase that ROCM does not like. Neither handcrafted C++/OpenCL nor Aparapi solutions trigger these errors.     I am hopeful that  when we implement a strategy for releasing events vs clReleaseEvent() (when we call ContextDriver.sync()/flush()) these may well go away.  All the AMD_ROCM flag did was essentially make all event queues block (as if we called clWaitForEvents after every enqueueXXX), so there maybe some other enqueuing issues that I have not seen yet.  Will neet to monitor as we make incremental fixes to event handling.

mikepapadim (approved):

Issue #16, opened 2020-05-21
Plans for Windows support?


cosmicdan:
Hi,

Loving the project - something I've been looking for a long time to learn GPGPU programming. I'm very surprised it took me so long to hear about the work you guys have done.

I have a (long-term) goal to use TornadoVM for development of a videogame and simulations (I'm only a hobbyist for now, nothing commercial) so naturally I'm curious if there's any plans to bring Tornado to Windows since most of my intended audience would be using it.

Are there any known roadblocks for Windows or is it just something that hasn't been looked at yet?

Just to clarify - I only use Windows myself as a secondary OS so this is by no means urgent - just curious on if it's an option for future.

Thanks!

jjfumero:
Hi, 
  This is something we haven't looked yet.  We do support Linux and OSx. However, since TornadoVM 0.6, we can also run with Graal 19.3 and JDK 11, and I think that combination could allow us to run on Windows (with a bit of tune from the bash scripts), but we haven't checked. This is in our future plans (long term).

cosmicdan:
Thanks Juan, I'm quite satisfied with that as I wont targeting the Windows platform for a short while either.

If I do end up getting bored one day and experiment with Windows before you guys get to it, I'll report my findings here - since you made it sound quite trivial there. I was mostly worried about natives not being ported to Windows - not the case here? I admit I've not examined the TVM stack in detail, apologies.

Finally, GraalVM 20 has "proper" Windows support though, it's still considered early access in 19.3 from what I remember - so for sanity's sake I'd guess that getting TVM working under Graal 20 should be a prerequisite for Windows testing.

⁣Get BlueMail for Android ​

jjfumero:
Thank you, 
  there are few parts with some dependencies like `cmake` to compile the JNI part of TornadoVM. But I think there is a binary for Windows. 

We do plan to add support for Graal 20.0. Stay tuned!

vsilaev:
I was able to compile graal-jd8 and graal-jdk11 on Windows 10 Pro x64 with MingW64 and minor changes. I guess, any Win7+ x64 will work.

I've started with [v0.8](https://github.com/beehive-lab/TornadoVM/tree/v0.8), and noticed that code was upgraded to C++ version of JNI libs recently. So the patches should be cherry-picked later. Obviously, if project maintainers will find them useful. 

[...]

PR #46, opened 2021-01-16
OBSOLETE: Fix non-JNI types in PTXStream.cpp; fix posix_memalign usage in OCLCo…

vsilaev:
1. Root `pom.xml` is modified to use Unix `cmake.generator` on Windows (MingW64 GCC build is assumed)
2.  `CMakeLists.txt` is modified to replace `-rdynamic` with `-export-dynamic` flag (recognizable by MingW64 GCC) and to exclude any prefixes generated by MingW64 to target DLL names
3. In `opencl-jni` the file `OCLContext.cpp` is modified to use `_aligned_malloc` function instead of Unix-only `posix_memalign`
4. In `ptx-jni` the file `PTXStream.cpp` is fixed to use correct JNI types like `jint`, `jlong`, `jdouble` instead of native C types `int`, `long`, `double`
5. `tornado` in `assemby` now has no unused environment variable `EXTERNAL_DIRS` , otherwise external Java library path is set incorrectly on Windows

Signed-off-by: vsilaev <vsilaev@gmail.com>

### DIFF
--- a/assembly/src/bin/tornado
+++ b/assembly/src/bin/tornado
@@ -120,7 +120,7 @@ if [[ $JAVA_VERSION == "9.0" || $JAVA_VERSION == "10.0" ]]; then
   exit 0
 fi
 
-TORNADO_FLAGS="-Djava.library.path=${TORNADO_SDK}/lib:$EXTERNAL_DIRS "
+TORNADO_FLAGS="-Djava.library.path=${TORNADO_SDK}/lib"
 if [ "$JAVA_VERSION" = "1.8" ]; then
   TORNADO_FLAGS="${TORNADO_FLAGS} -Djava.ext.dirs=${TORNADO_SDK}/share/java/tornado"
 else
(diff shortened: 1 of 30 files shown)


jjfumero:
Thank you @vsilaev. I will test it on Linux and on the FPGAs. I need to check the alignment on the Xilinx and Altera FPGAs. 

What to expect from us? I can work on this after 31st Jan. So you can expect feedback/comments from there. 

Thanks

stratika (changes requested):

jjfumero:
This PR should only push the Windows support. Please, do not mix Windows support with the `async/await` proposal.

stratika:
In addition, it would be good to resolve the `make clean` in Windows, instead of deleting manually the `bin/bin`, `bin/sdk` and `dist` folders.

And include a documentation file in `assembly/src/docs` that will be a guideline of all the steps mentioned in the issue #16 for executing TornadoVM in Windows.

vsilaev:
My apologizes - I played too hard with original branch for pull request and now I can't revert commits.
It was simpler to re-start over (I have a separate PATCH files anyway) and create new pull request - https://github.com/beehive-lab/TornadoVM/pull/54

It contains same changes from my side with all your feedback applied.

Issue #57, opened 2021-02-01
TestArrays#testInitParallel on AMD - failed with '[REASON] [ERROR] reset() was called after warmup()'


vsilaev:
When running `TestArrays#testInitParallel`  I get '[REASON] [ERROR] reset() was called after warmup()' exception thrown from `TornadoVM.compileTaskFromBytecodeToBinary`.

In fact, `TornadoTestBase.before` indeed reset ALL devices, but `TornadoVM.warmup` validates only one (VM-default ???) device. And if this device is different from SCHEDULE.TASK-device than it leads to an error (task device is changed via `s0.t0.device=0:1` command arg).

jjfumero:
This is a bug. In fact, it happens with any device, not only for AMD. 

We will look into it. 

How to reproduce it? 

```bash 
tornado-test.py -V -pk --debug -J"-Ds0.t0.device=0:1" uk.ac.manchester.tornado.unittests.arrays.TestArrays#testInitParallel
```

Thank you for the report.

vsilaev:
Btw, `-J` option is undocumented, and I had to alter `tornado` bash...

jjfumero:
This PR also fixes this issue: https://github.com/beehive-lab/TornadoVM/pull/82 

You are right, the issue is that the default device is was not the selected device for the task.

PR #82, opened 2021-04-12
[fix] Device Information from MxM examples fixed

jjfumero:
### Summary 
This fix also provides a patch for querying the device type of the latest
execution.

### Problem description

The problem is that, when setting the device name for a specific task, the
device requested is not the same as the device being queried. This is because
the query happens at the task-schedule level instead of the task level.

This patch provides a fix for querying the device type by returning the device
assigned to the first task. This is currently legal since TornadoVM executes
all task on the same device.

For a multi-task configuration, we will need to pass a task index when querying
a device associated with a task within a task-scheduler.

This fixes Github issue #539 (from the Internal Repository)

Issue first reported by @kotselidis 
```
When running an example e.g.

tornado uk.ac.manchester.tornado.examples.compute.MatrixMultiplication1D

with the default device (0.0) TornadoVM outputs the following output

Computing MxM of 512x512
CPU Execution: 1.27 GFlops, Total time = 211 ms
GPU Execution: 2.74 GFlops, Total Time = 98 ms
Speedup: 2.1530612244897958x
Verification true

However, device 0.0 is a CPU and not a GPU.

I propose to change the above to:

CPU execution->Single Threaded CPU execution
GPU execution->Accelerated execution ()
```

### Backend/s tested

- [X] OpenCL
- [ ] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ tornado -Ds0.t0.device=0:2 uk.ac.manchester.tornado.examples.compute.MatrixMultiplication1D
```
Assuming device 2 is a CPU, the output will be similar to this:

```
Computing MxM of 512x512
	Single Threaded CPU Execution: 0.96 GFlops, Total time = 280 ms
	TornadoVM Execution on CPU (Accelerated): 3.49 GFlops, Total Time = 77 ms
	Speedup: 3.6363636363636362x
	Verification true
```


jjfumero:
Update: this patch also solves https://github.com/beehive-lab/TornadoVM/issues/57

stratika (approved):

jjfumero:
Updated with `develop`

kotselidis (approved):

Issue #52, opened 2021-01-22
uk.ac.manchester.tornado.drivers.opencl.CLEvent incorrectly uses static field "buffer"


vsilaev:
**Describe the bug**
uk.ac.manchester.tornado.drivers.opencl.CLEvent incorrectly uses static field `buffer`.

This static field is used from multiple instance methods like `readEventTime` and `getCLStatus`.
When multiple threads trying to execute separate `TaskSchedule` events are conflicting accessing this field. The race leads to clearing buffer right before reading value returned from JNI and code fails with empty buffer error.

The `buffer` should be a local variable of method rather than static field.

**How To Reproduce**
Run several Java threads each executing own `TaskSchedule.execute()`

**Expected behavior**
Running several separate `TaskSchedule`-s from different threads should work ok.

**Computing system setup (please complete the following information):**
 - OS: any
 - OpenCL Version 1.2
 - TornadoVM commit id: `develop` branch of 0.9 version

jjfumero:
Thanks @vsilaev . Yes, that's a bug. The `buffer` variable should not be static.
Feel free to push a PR with this change.

vsilaev:
Will do after previous PR review (windows support). It's hard for me to use separate Ubuntu instance -- my primary development currently is on Windows, and TornadoVM doesn't works on VM-s like Oracle VirtualBox or Microsoft Hyper-V due to lack of OpenCL devices (couldn't figure out how to set them up).

jjfumero:
@vsilaev , are you providing this fix?

PR #83, opened 2021-04-12
[fix] ByteBuffer in the OCL event from static to non-static

jjfumero:
Minor fix. As reported in #52, the `ByteBuffer` in `OCLEvent` should be a non-static field.


gigiblender:
LGTM.

Not related to the issue reported, but I can see that in the same class `OCLEvent` there is the `internalBuffer` declared as a static variable. I think that it might trigger a similar problem and the fix is simple.

Should we include that in this PR ?

jjfumero:
Yes @gigiblender.  good catch.

gigiblender (approved):

stratika (approved):

PR #85, opened 2021-04-13
Runtime error when two tasks have the same name

jjfumero:
Throw an exception when at least two tasks within the same task-scheduler have the same name. 

Developers much provide a unique task-id per task within the same task-scheduler, otherwise, TornadoVM will fail to obtain the right binary from the code cache. However, prior to this PR there was no exception or help to guide the user. 

Github issue #569 (Internal)

#### OS tested

This is at the API level. So it should not matter the backend and the OS. 

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ tornado-test.py -V -pk --debug uk.ac.manchester.tornado.unittests.fails.RuntimeFail#test01
```

The test launches an exception `TornadoTaskRuntimeException` that is watched in the unit test.

### DIFF
--- a/assembly/src/docs/15_FAQ.md
+++ b/assembly/src/docs/15_FAQ.md
@@ -2,10 +2,12 @@
 
 ## 1. What can TornadoVM do?
 
-TornadoVM accelerates parts of your Java applications on heterogeneous hardware devices such as multicore CPUs, GPUs, and FPGAs.
+TornadoVM accelerates parts of your Java applications on heterogeneous hardware devices such as multicore CPUs, GPUs,
+and FPGAs.
 
-TornadoVM is currently being used to accelerate machine learning and deep learning applications, computer vision, physics simulations, financial applications, computational photography, natural language processing and signal processing.
-
+TornadoVM is currently being used to accelerate machine learning and deep learning applications, computer vision,
+physics simulations, financial applications, computational photography, natural language processing and signal
+processing.
 
 ## 2. How can I use it?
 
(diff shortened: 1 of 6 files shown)


stratika (approved):
@jjfumero, it looks fine to me.

assembly/src/docs/15_FAQ.md
```diff
@@ -26,78 +29,84 @@ Note: To run TornadoVM on ARM Mali, install TornadoVM with GraalVM and JDK 11. M
...
+heterogeneous devices (multi-core CPUs, GPUs, and FPGAs). We sometimes refer to a VM inside a VM (nested VM). The main
+VM is the Java Virtual Machine, and TornadoVM sits on top of that.
 
-TornadoVM makes use of the Java Virtual Machine Common Interface (JVMCI) that is included from Java 9 to compile Java bytecode to OpenCL C / PTX at runtime. As a JVMCI implementation, TornadoVM uses Graal (it extends the Graal IR and includes new backends for OpenCL C and PTX code generation).
+You can find more information
+here: [https://dl.acm.org/doi/10.1145/3313808.3313819](https://dl.acm.org/doi/10.1145/3313808.3313819)
 
-## 12. How do I know which parts of my application are suitable for acceleration?
-Workloads with for-loops that do not have dependencies between iterations are very good candidates to offload on accelerators. Examples of this pattern are NBody computation, Black-scholes, DFT, KMeans, etc.
+## 10. How it interacts with OpenJDK?
```
stratika, line 83: How does TornadoVM interact with OpenJDK?

kotselidis (approved):

PR #90, opened 2021-04-21
Hotfix/673/thanos

stratika:
#### Description

This PR addresses an internal issue (673). This issue is that for FPGA compilation all the HLS logs and error logs are being dumped to the terminal. The proposal is to dump them in two log files that will be stored in the user-defined directory where the bitstream will be stored. 

This PR addresses the issue by writing these dumps in two files: a) `output.log` that stores all the information about HLS compilation, and b) `error.log` that stores any error coming from the HLS compilation. In addition, a new flag is introduced that can be used by users to also dump the logs in the terminal (`-Dtornado.fpgaDumpLog`).

#### Backend/s tested

- [x] OpenCL
- [ ] PTX

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [x] Yes
- [ ] No

Here follows an example for compiling `DFT` for FPGA execution:
- Command that dumps the logs in terminal. The output of compilation in FPGA emulation mode: 
[dumps_in_terminal.txt](https://github.com/beehive-lab/TornadoVM/files/6349333/dumps_in_terminal.txt)

```
tornado --printKernel --debug -Dtornado.fpgaDumpLog=True -Ds0.t0.device=0:1 uk.ac.manchester.tornado.examples.dynamic.DFTDynamic 1024 default 1
```
- Command that does not dump the logs in terminal. The output of compilation in FPGA emulation mode: 
[dumps_in_file.txt](https://github.com/beehive-lab/TornadoVM/files/6349339/dumps_in_file.txt)
[output.log](https://github.com/beehive-lab/TornadoVM/files/6349340/output.log)
[error.log](https://github.com/beehive-lab/TornadoVM/files/6349341/error.log)

```
tornado --printKernel --debug -Ds0.t0.device=0:1 uk.ac.manchester.tornado.examples.dynamic.DFTDynamic 1024 default 1
```

### DIFF
--- a/runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/RuntimeUtilities.java
+++ b/runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/RuntimeUtilities.java
@@ -33,6 +33,7 @@ import static uk.ac.manchester.tornado.runtime.common.TornadoOptions.PRINT_SOURC
 import static uk.ac.manchester.tornado.runtime.common.TornadoOptions.PRINT_SOURCE_DIRECTORY;
 
 import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.FileWriter;
@@ -57,6 +58,9 @@ public class RuntimeUtilities {
     public static final int ONE_GIGABYTE = 1 * 1024 * 1024 * 1024;
     public static final int ONE_MEGABYTE = 1 * 1024 * 1024;
     public static final int ONE_KILOBYTE = 1 * 1024;
+
+    public static final String FPGA_OUTPUT_FILENAME = "output.log";
+    public static final String FPGA_ERROR_FILENAME = "error.log";
 
     public static long parseSize(String size) {
         if (size.endsWith("B")) {
@@ -371,7 +375,7 @@ public class RuntimeUtilities {
         }
     }
 
-    public static void systemCall(String[] command, boolean printStandardOutput) throws IOException {
+    public static void systemCall(String[] command, boolean printStandardOutput, String loggingDirectory) throws IOException {
         String stdOutput;
         StringBuffer standardOutput = new StringBuffer();
         StringBuffer errorOutput = new StringBuffer();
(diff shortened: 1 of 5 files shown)


runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/RuntimeUtilities.java
```diff
@@ -58,6 +59,9 @@
     public static final int ONE_MEGABYTE = 1 * 1024 * 1024;
     public static final int ONE_KILOBYTE = 1 * 1024;
 
+    public static final String FPGA_OUTPUT_FILENAME = "output.log";
```
jjfumero, line 62: Can these files be configured? We can put a name by default `outputFPGA.log` and `errorFPGA.log`. But we can also allow to obtain a name using the command line parameter. 

`-Dtornado.fpga.output.log=file.log`

`-Dtornado.fpga.error.log=fileError.log`

jjfumero (changes requested):

runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/RuntimeUtilities.java
stratika: I will change the default names, as it is suggested.

In my opinion, I do not see the point of having the name of the files being configurable by the user. The directory is configurable via the configuration file.

jjfumero: M... true. I missed that. Ok, let's only update the names.

gigiblender (approved):

runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/RuntimeUtilities.java
```diff
@@ -416,6 +426,19 @@ public static void writeStreamToFile(File file, byte[] source, boolean append) {
 
     }
 
+    public static void writeStringToFile(String filename, String source, boolean append) {
+        try (FileWriter fw = new FileWriter(filename, append)) {
```
gigiblender, line 430: From the FileWriter docs --> `Whether or not a file is available or may be created depends upon the underlying platform.`

I suppose the file is created if missing for both Windows and Linux

stratika: good catch, we do not have an FPGA on Windows to test it. But I will try to call the `FileWriter` and see if it is ok.

stratika: I can confirm that in bot OS, the file is created if it does not exist.

jjfumero (approved):

PR #96, opened 2021-04-29
Fix dataflow analysis to retrieve data accesses

gigiblender:
#### Description

This PR fixes the following:
- dataflow analysis to retrieve data accesses when not fully inlining.
- parameter load order in the callee when some of the parameters of the callee are not used.
- NPE in the scheduler when performing task migration with a CPU device.
- Methods/functions invoked from different places in the call graph do not get compiled twice.

I also included some cleanup of unused methods (`OCLCompilationResultsBuilder` and `OCLCompiler`)  and classes.

#### Problem description

In the Sketcher, Tornado runs a `TornadoDataflowAnalysis` phase that looks at the dataflow and assigns an `Access` for each parameter. Currently, we only analyze the method passed to the task, but not the non-inlined methods in the call graph. This can cause wrong accesses to be assigned to the parameters. Accesses of tasks are then used to generate nodes in the `TornadoGraph`.

With this PR, we merge the accesses of the caller with the accesses of the callees, obtaining the right accesses for the method.

Additionally, with the call graph below, method D could get compiled twice. This PR fixes this in the `OCL/PTXCompiler` and tests it in `TestMultipleFunctions::testNoDoubleCompilation` 

```
A → B → D
  ↘ C ↗
```
#### Backend/s tested

- [X] OpenCL
- [X] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

Run all the unit tests. 
`TestMultipleFunctions` should check for the task accesses and parameter order in non-inlined callees.


jjfumero (approved):

runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java
```diff
@@ -158,6 +158,12 @@ public static boolean isProfilerEnabled() {
      */
     public static final int TORNADO_SKETCHER_THREADS = Integer.parseInt(getProperty("tornado.sketcher.threads", "4"));
 
+    /**
+     * It enables automatic discovery and parallelisation of loops.
+     * Please note that this option experimental and may cause issues if enabled.
```
stratika, line 163: Add "is" -> Please note that this option is experimental and may cause issues if enabled.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/phases/TornadoSketchTierContext.java
```diff
@@ -28,28 +28,30 @@
...
 
 public class TornadoSketchTierContext extends HighTierContext {
 
-    protected final ResolvedJavaMethod method;
-    protected final TaskMetaData meta;
+    private final ResolvedJavaMethod method;
 
-    public TornadoSketchTierContext(Providers providers, PhaseSuite<HighTierContext> graphBuilderSuite, OptimisticOptimizations optimisticOpts, ResolvedJavaMethod method, TaskMetaData meta) {
+    /**
+     * Contains the argument accesses for this method. The array gets populated in the {@link TornadoDataflowAnalysis} phase.
```
stratika, line 38: Which method is the one referred as "this" method?

tornado-api/src/main/java/uk/ac/manchester/tornado/api/common/Access.java
```diff
@@ -41,13 +41,23 @@
  */
 package uk.ac.manchester.tornado.api.common;
 
+/**
+ * Enum used to specify the accesses of parameters passed on {@link uk.ac.manchester.tornado.api.common.TornadoFunctions.Task}
```
stratika, line 45: missing full stop at the end of line 45.

unittests/src/main/java/uk/ac/manchester/tornado/unittests/tasks/TestMultipleFunctions.java
```diff
@@ -0,0 +1,231 @@
+/*
+ * Copyright (c) 2013-2021, APT Group, Department of Computer Science,
```
stratika, line 2: This is a new file, so the year should be 2021.

stratika (changes requested):

stratika (approved):

PR #112, opened 2021-05-26
Change data transfers to be blocking by passing CL_TRUE instead of calling clWaitForEvents

gigiblender:
#### Description

We have observed that for some drivers (Intel iGPU), having a read/write call to the driver, and then calling `clWaitForEvents` can be much slower than just passing `CL_TRUE` always to the read/write.
The two solutions are equivalent but the latter can provide some speed-ups.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

`make && make tests`

### DIFF
--- a/drivers/opencl-jni/src/main/cpp/source/OCLCommandQueue.cpp
+++ b/drivers/opencl-jni/src/main/cpp/source/OCLCommandQueue.cpp
@@ -212,16 +212,14 @@ jlong transferFromHostToDevice(JNIEnv * env, jclass javaClass,
         std::cout << "[TornadoVM JNI] transferFromHostToDevice from " << deviceOffset << " (" << numBytes << ") from buffer: " << buffer << std::endl;
     }
     cl_event event;
-    cl_int status = clEnqueueWriteBuffer((cl_command_queue) commandQueue, (cl_mem) devicePtr, blocking_write,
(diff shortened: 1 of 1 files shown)


jjfumero:
Thanks @gigiblender . Have you measure the % slowdown when we have the clWaitForEvents?

jjfumero (approved):

stratika (approved):

gigiblender:
> Thanks @gigiblender . Have you measure the % slowdown when we have the clWaitForEvents?

I have not measured the end to end performance in Tornado. I have a simple C test that calls `clEnqueueWrite/ReadBuffer` and simulates the same behavior. 


```
OpenCL profile time when using CL_FALSE and `clWaitForEvents`
start - queued : 478664
end - start : 2583

OpenCL profile time when using CL_TRUE and no `clWaitForEvents`
start - queued : 2335
end - start : 834
```
 The `end - start` time will increase as I increase the data size, while the `start - queued` time will stay the same regardless the data size.

Overall, it looks there is extra `DISPATCH_TIME (start - queued`) if we use `CL_FALSE`

PR #113, opened 2021-05-27
Fix loop conditions when data flow is attached to the LoopBegin node 

gigiblender:
#### Description

This PR fixes the case when there is data flow attached to the `LoopBegin`. This data flow must execute regardless the loop condition is true or false.

#### Problem description
Give the test in `uk.ac.manchester.tornado.unittests.loops.TestLoops#testSingleThreadLoopCondition`: 
```
private static void testSingleThreadLoopCond(int[] in, int[] out) {
        int otherCompVal = in[0];
        int i = 0;
        for (; i < in.length / 4 - 1; i++) {
            int someNumber = getNumber(in, i, i % 4, 4);
            in[i] = someNumber + i;
        }

        if (i == otherCompVal) {
            int someNumber = getNumber(in, i, i % 4, 4) + 1000;
            out[i] = someNumber;
        }
    }

    private static int getNumber(int[] in, int base, int offset, int multiplier) {
        // Perform some address computation
        return in[base * multiplier + offset];
    }
```

It is possible to have data flow attached to the `LoopBegin` node, as shown in the IR graph below. This happens because the induction variable `i` is used in the comparison after the `for` loop:

image

In order to get the right results, the data flow attached to the `LoopBegin` node must be executed regardless the loop condition is evaluated to true or false.

With this PR, in cases when such dataflow is found, instead of generating 
```
for(;loopCondition;) {
  // instructions for the data flow attached to loop begin 
  // other instructions
}
``` 
we will generate
```
for(;;) {
  ...
  // instructions for the data flow attached to loop begin
 ...
  if (!(loopCondition)) {
     break;
   }
  // other instructions
}
```
If no extra dataflow attached to the `LoopBegin` is present, then we generate the normal `for(;loopCondition;)`.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [X] Yes
- [ ] No

Run the test 
```
tornado-test.py --printKernel --verbose uk.ac.manchester.tornado.unittests.loops.TestLoops#testSingleThreadLoopCondition
``` 
on the OpenCL backend


stratika (commented):
In the example and the screenshot. Which node is the one attached to `LoopBegin`?

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/lir/OCLControlFlow.java
```diff
@@ -133,49 +133,73 @@ public LoopPostOp() {
...
         private final Value condition;
 
+        private boolean generateIfBreakStatement = true;
+
         public LoopConditionOp(Value condition) {
             super(TYPE);
             this.condition = condition;
         }
 
+        public void generateIfBreakStatement(boolean value) {
```
stratika, line 166: rename to `setGenerateIfBreakStatement`

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java
```diff
@@ -179,6 +179,30 @@ public Value doValue(LIRInstruction instruction, Value value, OperandMode mode,
...
+            }
+            if (instruction instanceof LoopPostOp) {
+                loopPostOpIndex = index;
+            }
+            if (instruction instanceof LoopConditionOp) {
+                loopConditionOpIndex = index;
+            }
+        }
+
+        return (loopConditionOpIndex - 1 == loopPostOpIndex) && (loopConditionOpIndex - 2 == loopInitOpIndex);
```
stratika, line 203: would it make sense to use two constants for the subtraction value? Can you explain why `loopConditionOpIndex` is subtracted by 1, while `loopConditionOpIndex` is by 2?

jjfumero:
I don't see the issue. Was the problem because of the inlining? If you prevent inlining, is this happening? Also, it seems only for sequential for loops.

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLBlockVisitor.java
```diff
@@ -132,17 +132,7 @@ public Block enter(Block block) {
 
         if (block.isLoopHeader()) {
             loopCount++;
-            openclBuilder.emitLoopHeader(block);
-            // Temporary fix to remove the end scope of the most outer loop
-            // without changing the loop schematics in IR level.
-            if (openclBuilder.shouldRemoveLoop()) {
```
jjfumero, line 138: @gigiblender , did you check on FPGAs? We have this condition for optimizing some type of code on FPGAs.

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java
```diff
@@ -236,23 +263,6 @@ void emitBlock(Block block) {
...
-            if (isLoopDependencyNode(lir.getLIRforBlock(block).get(i))) {
-                for (int j = i; j < lir.getLIRforBlock(block).size(); j++) {
-                    if (!isLoopDependencyNode(lir.getLIRforBlock(block).get(j))) {
-                        emitOp(this, lir.getLIRforBlock(block).get(j));
-                        opPreEmit = lir.getLIRforBlock(block).get(j);
-                    }
-                }
-                break;
-            }
-        }
```
stratika, line 255: Why is this part removed?

jjfumero (changes requested):

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java
gigiblender: I could use a constant but I can't come up with a suitable name. 

If there are no nodes (fixed or floating) between `LoopBegin` and the `If` corresponding to the loop condition, then in the LIR, the `LoopInitOp` and the `LoopPostOp` will be immediately followed by a `LoopConditionOp`

That is, the `emitLoopBegin` will be called right before the `emitIf` is called.
https://github.com/beehive-lab/TornadoVM/blob/202a27b0ee1f58368feab95619d9e108e7b52cda/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLNodeLIRBuilder.java#L505
https://github.com/beehive-lab/TornadoVM/blob/202a27b0ee1f58368feab95619d9e108e7b52cda/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLNodeLIRBuilder.java#L471

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java
gigiblender: This part tried to fix a similar issue but was only moving a single instruction outside the loop. 
I think that my fix covers this case

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLBlockVisitor.java
gigiblender: Yes, the outer most `@Parallel` loop is not generated on FPGAs

gigiblender: https://github.com/beehive-lab/TornadoVM/blob/202a27b0ee1f58368feab95619d9e108e7b52cda/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java#L263
This condition prevents it from being generated

gigiblender:
> In the example and the screenshot. Which node is the one attached to `LoopBegin`?

Node 8, the `Phi` node is attached to the `LoopBegin`. To the `PhiNode, there is additional data flow attached (operations to compute the `OCLAddress`)

gigiblender:
> I don't see the issue. Was the problem because of the inlining? If you prevent inlining, is this happening? Also, it seems only for sequential for loops.

I have not tested with inlining turned off. 
The issue is also for `@Parallel` loops. I have now added a similar test that uses a `@Parallel` loop. The condition of the `if` outside the loop will hold for a single thread.

The kernel generated for the single-threaded test before the change:  
```
// BLOCK 0
  ul_0  =  (ulong) _frame[3];
  ul_1  =  (ulong) _frame[4];
  ul_2  =  ul_0 + 24L;
  i_3  =  *((__global int *) ul_2);
  // BLOCK 1 MERGES [0 2 ]
  l_5  =  (long) i_4;
  l_6  =  l_5 << 2;
  l_7  =  l_6 + 24L;
  i_8  =  i_4 << 2;
  i_9  =  i_4 % 4;
  i_10  =  i_8 + i_9;
  l_11  =  (long) i_10;
  l_12  =  l_11 << 2;
  l_13  =  l_12 + 24L;
  ul_14  =  ul_0 + l_13;
  i_4  =  0;
  for(;i_4 < 255;)  l_5  =  (long) i_4;
  l_6  =  l_5 << 2;
  l_7  =  l_6 + 24L;
  i_8  =  i_4 << 2;
  i_9  =  i_4 % 4;
  i_10  =  i_8 + i_9;
  l_11  =  (long) i_10;
  l_12  =  l_11 << 2;
  l_13  =  l_12 + 24L;
  {
    // BLOCK 2
    i_15  =  *((__global int *) ul_14);
    ul_16  =  ul_0 + l_7;
    i_17  =  i_15 + i_4;
    *((__global int *) ul_16)  =  i_17;
    i_18  =  i_4 + 1;
    i_4  =  i_18;
  }  // B2
  // BLOCK 3
  z_19  =  i_4 == i_3;
  if(z_19)
  {
    // BLOCK 4
    i_20  =  *((__global int *) ul_14);
    ul_21  =  ul_1 + l_7;
    i_22  =  i_20 + 1000;
    *((__global int *) ul_21)  =  i_22;
    return;
  }  // B4
  else
  {
    // BLOCK 5
    return;
  }  // B5
}  //  kernel

```
The reason we have duplicated instructions in `BLOCK 1` is because [this code](https://github.com/beehive-lab/TornadoVM/blob/202a27b0ee1f58368feab95619d9e108e7b52cda/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java#L245) only works for a single instruction.

The kernel generated after the change:
```
// BLOCK 0
  ul_0  =  (ulong) _frame[3];
  ul_1  =  (ulong) _frame[4];
  ul_2  =  ul_0 + 24L;
  i_3  =  *((__global int *) ul_2);
  // BLOCK 1 MERGES [0 2 ]
  i_4  =  0;
  for(;;)
  {
    l_5  =  (long) i_4;
    l_6  =  l_5 << 2;
    l_7  =  l_6 + 24L;
    i_8  =  i_4 << 2;
    i_9  =  i_4 % 4;
    i_10  =  i_8 + i_9;
    l_11  =  (long) i_10;
    l_12  =  l_11 << 2;
    l_13  =  l_12 + 24L;
    ul_14  =  ul_0 + l_13;
    if(!(i_4 < 255))
    {
      break;
    }
    // BLOCK 2
    i_15  =  *((__global int *) ul_14);
    ul_16  =  ul_0 + l_7;
    i_17  =  i_15 + i_4;
    *((__global int *) ul_16)  =  i_17;
    i_18  =  i_4 + 1;
    i_4  =  i_18;
  }  // B2
  // BLOCK 3
  z_19  =  i_4 == i_3;
  if(z_19)
  {
    // BLOCK 4
    i_20  =  *((__global int *) ul_14);
    ul_21  =  ul_1 + l_7;
    i_22  =  i_20 + 1000;
    *((__global int *) ul_21)  =  i_22;
    return;
  }  // B4
  else
  {
    // BLOCK 5
    return;
  }  // B5
}  //  kernel
```
On the final iteration, regardless the loop condition is true or false, we compute the address `ul_14` and the value of the variable `l_7`, which are later used in the `if` statement outside the loop.

jjfumero:
@gigiblender , is there any decrease in performance. Let's measure the kfusion with this patch.

jjfumero (approved):

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLCompilationResultBuilder.java
stratika: Thanks @gigiblender, lets move the condition to a new method and add a Javadoc description.

stratika (approved):

PR #122, opened 2021-06-11
[refactor] GridTask renamed to GridScheduler

stratika:
#### Description

This PR applies a change that we have discussed internally regarding the name of the `GridTask`. The fact that we can have multiple tasks that use the same `GridTask` object, has made us think that the naming is not appropriate as it makes developers think that one task corresponds to a `GridTask`.

Therefore we have agreed to rename the `GridTask` to `GridScheduler`. Here is an example of usage:
```java
        WorkerGrid worker = new WorkerGrid1D(size);
        GridScheduler gridScheduler = new GridScheduler();
        gridScheduler.setWorkerGrid("s02.t0", worker);
        gridScheduler.setWorkerGrid("s02.t1", worker);
        gridScheduler.setWorkerGrid("s02.t2", worker);
        KernelContext context = new KernelContext();

        TaskSchedule s0 = new TaskSchedule("s0") //
                .streamIn(a, b) //
                .task("t0", TestCombinedTaskSchedule::vectorAddV2, context, a, b, cTornado) //
                .task("t1", TestCombinedTaskSchedule::vectorMulV2, context, cTornado, b, cTornado) //
                .task("t2", TestCombinedTaskSchedule::vectorSubV2, context, cTornado, b, cTornado) //
                .streamOut(cTornado);
        s0.execute(gridScheduler);
```

No functional code has changed in this PR.

#### Backend/s tested

- [x] OpenCL
- [x] PTX

#### OS tested

- [ ] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

No need to test. For sanity check, you can use:
```bash
grep -nr "GridTask" .
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ public class Compute {
         // When using the kernel-parallel API, we need to create a Grid and a Worker
 
         WorkerGrid workerGrid = new WorkerGrid2D(size, size);    // Create a 2D Worker
-        GridTask gridTask = new GridTask("s0.t0", workerGrid);   // Attach the worker to the Grid
+        GridScheduler gridScheduler = new GridScheduler("s0.t0", workerGrid);  // Attach the worker to the Grid
         KernelContext context = new KernelContext();             // Create a context
         workerGrid.setLocalWork(32, 32, 1);                      // Set the local-group size
 
@@ -123,7 +123,7 @@ public class Compute {
                 .streamIn(A, B)                                 // Stream data from host to device
                 .task("t0", Compute::mxmKernel, context, A, B, C, size)  // Each task points to an existing Java method
                 .streamOut(C);                                  // sync arrays with the host side
-        ts.execute(gridTask);   // Execute with a GridTask
+        ts.execute(gridScheduler);   // Execute with a GridScheduler
     }
 }
 ```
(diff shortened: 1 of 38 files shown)


jjfumero (approved):

gigiblender (approved):

vsilaev:
https://github.com/beehive-lab/TornadoVM/blob/develop/runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/meta/AbstractMetaData.java

Now there are 2 methods with overloaded params:
```
public void setGridScheduler(boolean use)
public void setGridScheduler(GridScheduler gridScheduler) 
```

I guess, the one with `boolean use` should be refactored to `setGridSchedulerEnabled` while you already have
`isGridSchedulerEnabled`

stratika:
> https://github.com/beehive-lab/TornadoVM/blob/develop/runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/meta/AbstractMetaData.java
> 
> Now there are 2 methods with overloaded params:
> 
> ```
> public void setGridScheduler(boolean use)
> public void setGridScheduler(GridScheduler gridScheduler) 
> ```
> 
> I guess, the one with `boolean use` should be refactored to `setGridSchedulerEnabled` while you already have
> `isGridSchedulerEnabled`

Thanks @vsilaev, good point. This was due to having the same name of method prior to this PR. I will update it in a later PR.

PR #125, opened 2021-06-15
[fix] FMA LIR Type fixed for ternary nodes

jjfumero:
Ternary nodes should infer the LIR type using the three operands. LIR types for `FMA` instructions were getting the LIR with the first two values, instead of the 3 operands. This patch fixes that. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash 
make tests
tornado-test.py -pk --debug -V -J"-Ds0.t0.device=0:0" --fast uk.ac.manchester.tornado.unittests.matrices.TestMatrices#testMatrixMultiplication
```

### DIFF
--- a/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/lir/OCLArithmeticTool.java
+++ b/drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/lir/OCLArithmeticTool.java
@@ -462,7 +462,7 @@ public class OCLArithmeticTool extends ArithmeticLIRGenerator {
     }
 
     public Value emitFMAInstruction(Value op1, Value op2, Value op3) {
-        LIRKind resultKind = LIRKind.combine(op1, op2);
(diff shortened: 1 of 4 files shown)


gigiblender (approved):

stratika (commented):
Can we enhance the unit-test to reviel the problem that this PR sorts out? Because the current unit-test in `develop` results in success.

jjfumero:
I added a few tests:

```bash
tornado-test.py --debug -pk -V --fast  uk.ac.manchester.tornado.unittests.math.TestMath#testFMA
tornado-test.py --debug -pk -V --fast  uk.ac.manchester.tornado.unittests.math.TestMath#testFMA2
```

jjfumero:
Graal will perform a typecast before the FMA operation, in the case the data types differ. However, still the LIR should be the combination of all operands.

stratika (approved):

Issue #79, opened 2021-04-08
Enhance Reduction Operators for TornadoMath functions


stratika:
I tried in [`TestReductionsDoubles`](https://github.com/beehive-lab/TornadoVM/blob/master/unittests/src/main/java/uk/ac/manchester/tornado/unittests/reductions/TestReductionsDoubles.java) to replace `Math.max()` with `TornadoMath.max()`.

```java
private static void maxReductionAnnotation(double[] input, @Reduce double[] result) {
        for (@Parallel int i = 0; i < input.length; i++) {
            result[0] = Math.max(result[0], input[i]);
        }
}
```

```java
private static void maxReductionAnnotation(double[] input, @Reduce double[] result) {
        for (@Parallel int i = 0; i < input.length; i++) {
            result[0] = TornadoMath.max(result[0], input[i]);
        }
}
```
This replacement resulted in the following error.

```
Test: class uk.ac.manchester.tornado.unittests.reductions.TestReductionsDoubles#testMaxReduction
	Running test: testMaxReduction           ................  [FAILED] 
		\_[REASON] [ERROR] Automatic reduce operation not supported yet: 17|OCLFPBinaryIntrinsic
	uk.ac.manchester.tornado.api.exceptions.TornadoRuntimeException: [ERROR] Automatic reduce operation not supported yet: 17|OCLFPBinaryIntrinsic
	at uk.ac.manchester.tornado.runtime.analyzer.ReduceCodeAnalysis.getReduceOperation(ReduceCodeAnalysis.java:117)
```

The same issue is shown for other operations (e.g. min, pow...)

So, the feature is to **update the [`ReduceCodeAnalysis`](https://github.com/beehive-lab/TornadoVM/blob/master/runtime/src/main/java/uk/ac/manchester/tornado/runtime/analyzer/ReduceCodeAnalysis.java) to support the operators in TornadoMath package**.

**How to test:**

Change the `Math.max()` with `TornadoMath.max()` and run:

```
$ make
$ tornado -ea -Dtornado.unittests.verbose=True -Xmx6g -Dtornado.recover.bailout=False uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  uk.ac.manchester.tornado.unittests.reductions.TestReductionsDoubles#testMaxReduction
```

jjfumero:
Thank you for the report @stratika. Let's schedule this enhancement for TornadoVM 0.10 .

PR #127, opened 2021-06-22
Support for `getArray` for vector types in TornadoVM

jjfumero:
[OCL][PTX] Support for the `getArray` for vector types;
    
The method `getArray` from the vector type classes was not supported. This patch introduces a new plugin in the TornadoVM JIT compiler to deal with get-subarrays within vectors of Type FloatX (X >= 2).
    
Github issue #79 (Internal). 

This PR also introduces a new node in the TornadoVM IR, `GetArrayNode`, which obtains a subarray within a larger array to be loaded with `vloadX` OpenCL instructions. 

The following Control Flow Graph shows an example of this new node. The result of a `getArray` is a load of vector type in private memory. 

Screenshot from 2021-06-22 11-01-14

#### Backend/s tested

- [X] OpenCL
- [X] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

All supported unittests are passing with both backends.

Additionally:

```bash
tornado-test.py -V -pk --igv uk.ac.manchester.tornado.unittests.vectortypes.TestFloats#testVectorFloat8_Storage
```


jjfumero:
This PR is ready for review

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/vector/VectorValueNode.java
```diff
@@ -160,30 +159,53 @@ private void generateVectorAssign(NodeLIRBuilderTool gen, LIRGeneratorTool tool,
...
-            case 8:
+            }
+            case 8: {
                 final OCLOp8 op8 = VectorUtil.resolveAssignOp8(getOCLKind());
                 s0 = getParam(gen, tool, 0);
+                OCLKind kindOp = (OCLKind) s0.getPlatformKind();
+                if (kindOp.isVector()) {
+                    gen.setResult(this, s0);
+                    return;
+                }
```
stratika, line 208: Can you provide some explanation here? 

Why is only `s0` used to set the result in the `NodeLIRBuilderTool`?

jjfumero: Yes. So what happens is that the JIT compiler tried to create a new vector type from scalar values. However, we need to cover the case in which we just make a copy of a vector type (without invoking the subcomponents)

```
vector4 a = b;
```

Otherwise, the compiler will generate the following sequence (which is incorrect)

```
vector4 a = (vector4) (b, 0, 0, 0);  // b is vector4
```

For those cases, we need to control de `PlatformKind` of the first operand. In the case is a vector, we just pass the result to the next `LIROp` operation.

stratika: yes, it makes sense. Thanks for the info.

drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/vector/VectorValueNode.java
```diff
@@ -160,30 +159,53 @@ private void generateVectorAssign(NodeLIRBuilderTool gen, LIRGeneratorTool tool,
...
         switch (kind.getVectorLength()) {
-            case 2:
+            case 2: {
                 final OCLOp2 op2 = VectorUtil.resolveAssignOp2(getOCLKind());
                 s0 = getParam(gen, tool, 0);
+                OCLKind kindOp = (OCLKind) s0.getPlatformKind();
+                if (kindOp.isVector()) {
+                    gen.setResult(this, s0);
+                    return;
+                }
```
stratika, line 169: Can we move this in a method since it is used in every case?
I am not sure about the naming, but it could be something like `setResultIfVector(NodeLIRBuilder gen, Value s0)`.

drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/nodes/vector/VectorValueNode.java
```diff
@@ -150,26 +150,49 @@ private void generateVectorAssign(NodeLIRBuilderTool gen, LIRGeneratorTool tool,
...
         Value s0,s1,s2,s3,s4,s5,s6,s7;
         switch (kind.getVectorLength()) {
-            case 2:
+            case 2: {
                 s0 = getParam(gen, tool, 0);
+                PTXKind kindOp = (PTXKind) s0.getPlatformKind();
+                if (kindOp.isVector()) {
+                    gen.setResult(this, s0);
+                    return;
+                }
```
stratika, line 159: Same comment regarding adding a method for this.

jjfumero:
All changes suggestion from @stratika addressed

stratika (approved):

gigiblender (approved):

PR #132, opened 2021-07-21
[feat] TornadoVM support for a single VM per device

jjfumero:
#### Description

This fixes dynamic device switching when having multiple devices of different types (e.g., CPU, GPU, HD Graphics).
The issue was the number of threads deployed was not the correct one, getting only the previous configuration.

#### Solution provided     

This patch extends TornadoVM with a VM instance per device. There is a mapping between the `contextID` from TornadoVM bytecodes and the device selection.
    
The TornadoTaskScheduler keeps a mapping between VM instances and devices.
```
    [Device01] -> vm01
    [Device02] -> vm02
    ....
    [DeviceNN] -> vmNN
```
    
Additionally, TornadoVM checks the device using the `contextID` from the bytecodes. Besides, the device list is a one-to-one mapping between the physical information stored in the TornadoBackend and the VM. This way, it is easier to switch devices based on an index.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

* 100% Test passing 
```
make tests
```

* Live Task Migration: https://github.com/jjfumero/qconlondon2020-tornadovm 

* Slambench passing.


runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
```diff
@@ -52,10 +52,11 @@
 
 public class TornadoExecutionContext {
 
-    private final String name;
     private final int MAX_TASKS = 128;
-    private final ScheduleMetaData meta;
+    private final int INITIAL_DEVICE_CAPACITY = 16;
```
stratika, line 56: Any justification of the number 16?

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
```diff
@@ -81,7 +82,10 @@ public TornadoExecutionContext(String id, TornadoProfiler profiler) {
         objectMap = new HashMap<>();
         objects = new ArrayList<>();
         objectState = new ArrayList<>();
-        devices = new ArrayList<>();
+        devices = new ArrayList<>(INITIAL_DEVICE_CAPACITY);
+        // IntStream.range(0,
+        // INITIAL_DEVICE_CAPACITY).<TornadoAcceleratorDevice>mapToObj(i ->
+        // null).forEach(devices::add);
```
stratika, line 88: remove comments

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
```diff
@@ -204,36 +211,41 @@ public void mapAllTo(TornadoDevice mapping) {
...
         info("assigning %s to %s", id, target.getDeviceName());
 
-        int deviceIndex = devices.indexOf(target);
         if (deviceIndex == -1) {
-            deviceIndex = devices.size();
-            devices.add(accelerator);
+            // deviceIndex = devices.size();
+            deviceIndex = task.meta().getDeviceIndex();
+            checkDeviceListSize(deviceIndex);
+            devices.set(deviceIndex, accelerator);
```
stratika, line 246: you can call the `setDevice()` method here.

jjfumero: `devices` is a linkedList.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
jjfumero: No, random value. Usually, we have [1-5] devices, depending on the system and number of drivers installed.

runtime/src/main/java/uk/ac/manchester/tornado/runtime/TornadoVM.java
```diff
@@ -156,7 +158,9 @@ public TornadoVM(TornadoExecutionContext graphContext, byte[] code, int limit, T
             final int deviceIndex = buffer.getInt();
             debug("loading context %s", contexts.get(deviceIndex));
             final long t0 = System.nanoTime();
-            contexts.get(deviceIndex).ensureLoaded();
+            if (contexts.get(deviceIndex) != null) {
+                contexts.get(deviceIndex).ensureLoaded();
```
gigiblender, line 162: Since now there is a mapping between <device, TornadoVM object>, is there any reason to keep a list of `contexts` inside each TornadoVM object?

jjfumero: The mapping is between a logic device ID (abstract representation in TornaoVM) and the physical device. Yes, we need to keep this high-level mapping for the concurrent BC interpreter.

jjfumero:
All changes done

gigiblender (approved):

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
stratika: I said that because the `setDevice()` method does this functionality:
```
public void setDevice(int index, TornadoAcceleratorDevice device) {
        checkDeviceListSize(index);
        devices.set(index, device);
}
```

stratika (approved):

runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java
jjfumero: Ok, I got what you meant. Yes, you are right.

PR #200, opened 2022-07-25
[PTX] Fix #195: PTX Kernel Signature match with arguments size during kernel launch

### LINKED ISSUE #195: PTX Backend reporting error code 701 when dispatching cuLaunchKernel on NVIDIA 30XX GPUs
### Describe the bug 

When running with the PTX backend on NVIDIA 30XX GPUs (e.g., 3070), kernel dispatch fails due to error 701.

```bash
[TornadoVM-PTX-JNI] ERROR : cuLaunchKernel -> Returned: 701
```

Following the NVIDIA documentation, this error means `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES` 

[link](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDA__TYPES_g0cdead942fd5028d157641eef6bdeeaa.html#gg0cdead942fd5028d157641eef6bdeeaab5da09cc5697599a56a71a04184ffdaa)

> This indicates that a launch did not occur because it did not have appropriate resources. This error usually indicates that the user has attempted to pass too many arguments to the device kernel, or the kernel launch specifies too many threads for the kernel's register count. Passing arguments of the wrong size (i.e. a 64-bit pointer when a 32-bit int is expected) is equivalent to passing too many arguments and can also result in this error.

Possibly this is related to the number if threads in the case of TornadoVM. Note that this error cannot be reproduced on older GPUs (e.g., 2060, 1050, etc), or the OpenCL backend running on the same 30XX GPU. 

NVIDIA Driver used: `510.60.02` 

### How To Reproduce

```bash
$ make BACKEND=ptx
$ tornado -Ds0.t0.device=0:0 --printKernel -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixMultiplication2D
```

### Computing system setup (please complete the following information):

 - OS: Fedora 36, NVIDIA Driver 510.60.02
 - OpenCL Version: 3.0 
 - 7.6
 - TornadoVM commit id: https://github.com/beehive-lab/TornadoVM/commit/5e5fbbc8cc2e79f3ecb87c32dedfec48d0f6b44d

---------------------------------------------------------------------

jjfumero:
#### Description

This PR provides a fix for #195. The issue is that, from the NVIDIA Driver 510.60, the driver checks for the argument size of each parameter passed to the kernel. This was, in fact, a bug in the TornadoVM runtime. This patch provides a fix by using all buffers of the same size, and then padding when necessary. 

With this patch, TornadoVM can run many unittests, including Matrix Multiplication benchmarks. However, there are still some errors during the kernel launch only when using the TornadoVM Kernel API. The fix for this part will be provided in a separate PR. Additionally, some of the reductions produce not-correct results (only from 30XX GPUs with NVIDIA driver >= 510.60). 
This will be also patched in a separated PR. 

The TornadoVM backend for GPUs using NVIDIA Drivers <= 510.54 is still working as normal. 

#### Backend/s tested

This PR only affects the PTX-backed, 

- [ ] OpenCL
- [X] PTX
- [ ] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ make BACKEND=ptx
$ tornado -Ds0.t0.device=0:0 -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixMultiplication2D
$ tornado --threadInfo --printKernel -ea -Dtornado.unittests.verbose=True -Xmx6g -Dtornado.recover.bailout=False  -m  tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  uk.ac.manchester.tornado.unittests.reductions.TestReductionsFloats#testComputeAverage
```

### DIFF
--- a/drivers/ptx-jni/src/main/cpp/source/ptx_utils.cpp
+++ b/drivers/ptx-jni/src/main/cpp/source/ptx_utils.cpp
@@ -29,9 +29,6 @@
 CUresult record_events_create(CUevent* beforeEvent, CUevent* afterEvent) {
     CUresult result = cuEventCreate(beforeEvent, CU_EVENT_DEFAULT);
     LOG_PTX_AND_VALIDATE("cuEventCreate (beforeEvent)", result);
-    if (result == CUDA_ERROR_OUT_OF_MEMORY) {
-
-    }
     result = cuEventCreate(afterEvent, CU_EVENT_DEFAULT);
     LOG_PTX_AND_VALIDATE("cuEventCreate (afterEvent)", result);
     return result;
--- a/drivers/ptx-jni/src/main/cpp/source/ptx_utils.cpp
+++ b/drivers/ptx-jni/src/main/cpp/source/ptx_utils.cpp
@@ -28,14 +28,17 @@
 
 CUresult record_events_create(CUevent* beforeEvent, CUevent* afterEvent) {
     CUresult result = cuEventCreate(beforeEvent, CU_EVENT_DEFAULT);
-    LOG_PTX_AND_VALIDATE("cuEventCreate", result);
+    LOG_PTX_AND_VALIDATE("cuEventCreate (beforeEvent)", result);
+    if (result == CUDA_ERROR_OUT_OF_MEMORY) {
+
+    }
     result = cuEventCreate(afterEvent, CU_EVENT_DEFAULT);
-    LOG_PTX_AND_VALIDATE("cuEventCreate", result);
+    LOG_PTX_AND_VALIDATE("cuEventCreate (afterEvent)", result);
     return result;
 }
 
-CUresult record_event(CUevent* beforeEvent, CUstream* stream) {
-    CUresult result = cuEventRecord(*beforeEvent, *stream);
+CUresult record_event(CUevent* event, CUstream* stream) {
+    CUresult result = cuEventRecord(*event, *stream);
     LOG_PTX_AND_VALIDATE("cuEventRecord", result);
     return result;
 }
--- a/drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/mm/PTXKernelArgs.java
+++ b/drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/mm/PTXKernelArgs.java
@@ -57,45 +57,6 @@ public class PTXKernelArgs extends PTXByteBuffer implements KernelArgs {
         return callArguments;
     }
 
-    public int getArgumentsTotalSizeInBytes() {
-        int size = 0;
-        for (CallArgument argument : callArguments) {
-            if (argument.isReferenceType()) {
-                size += 8; // Reference is 8 bytes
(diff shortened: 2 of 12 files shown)


drivers/ptx-jni/src/main/cpp/source/ptx_utils.cpp
```diff
@@ -28,14 +28,17 @@
 
 CUresult record_events_create(CUevent* beforeEvent, CUevent* afterEvent) {
     CUresult result = cuEventCreate(beforeEvent, CU_EVENT_DEFAULT);
-    LOG_PTX_AND_VALIDATE("cuEventCreate", result);
+    LOG_PTX_AND_VALIDATE("cuEventCreate (beforeEvent)", result);
+    if (result == CUDA_ERROR_OUT_OF_MEMORY) {
+
+    }
```
kotselidis, line 34: dead code

stratika, line 34: Remove this code. I believe that if result is not Success, the `LOG_PTX_AND_VALIDATE` will display the error code.

drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/mm/PTXKernelArgs.java
```diff
@@ -58,6 +57,45 @@ public List<CallArgument> getCallArguments() {
...
+                    size += 2;
+                } else if (klass == double.class) {
+                    size += 8;
+                } else if (klass == long.class) {
+                    size += 8;
+                } else if (klass == byte.class) {
+                    size += 1;
+                } else if (klass == char.class) {
+                    size += 1;
+                }
```
stratika, line 93: is there any probability to have `else`? have you covered all types?

stratika (changes requested):

drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/mm/PTXKernelArgs.java
jjfumero: What do you mean?

jjfumero: We have covered all primitives. This is detected as a changed because we moved code due to the code formatter. Not really a change.

jjfumero: Sorry, as it is now, this code is never invoked. We can remove this function.

jjfumero:
All changes done

stratika (approved):

kotselidis (approved):

Issue #179, opened 2022-03-07
Problematic frame:  C  0x00007f0f2f6af230


NicevilleWest:
Thank you for taking the time to submit a bug report. Please fill the following template with your report to help our team to fix and/or propose solutions. 

---------------------------------------------------------------------

### Describe the bug 
Environment 👍 
```
Linux MININT-AO4NFSP 5.10.60.1-microsoft-standard-WSL2 #1 SMP Wed Aug 25 23:20:18 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

```bash
Installation with script successful (See below)
[INFO] Reactor Summary for tornado 0.13-dev:
[INFO]
[INFO] tornado ............................................ SUCCESS [  0.869 s]
[INFO] tornado-api ........................................ SUCCESS [  2.268 s]
[INFO] tornado-runtime .................................... SUCCESS [  1.766 s]
[INFO] tornado-matrices ................................... SUCCESS [  0.757 s]
[INFO] tornado-drivers .................................... SUCCESS [  0.378 s]
[INFO] tornado-drivers-common ............................. SUCCESS [  0.607 s]
[INFO] tornado-drivers-opencl ............................. SUCCESS [  1.772 s]
[INFO] tornado-drivers-opencl-jni ......................... SUCCESS [  5.096 s]
[INFO] tornado-drivers-spirv .............................. SUCCESS [  1.461 s]
[INFO] tornado-drivers-spirv-levelzero-jni ................ SUCCESS [  4.781 s]
[INFO] tornado-drivers-ptx ................................ SUCCESS [  1.426 s]
[INFO] tornado-drivers-ptx-jni ............................ SUCCESS [  5.262 s]
[INFO] tornado-examples ................................... SUCCESS [  1.854 s]
[INFO] tornado-benchmarks ................................. SUCCESS [  4.731 s]
[INFO] tornado-unittests .................................. SUCCESS [  2.596 s]
[INFO] tornado-annotation ................................. SUCCESS [  0.602 s]
[INFO] tornado-assembly ................................... SUCCESS [  4.416 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15.099 s (Wall Clock)
[INFO] Finished at: 2022-03-07T01:07:26+01:00
[INFO] ------------------------------------------------------------------------

##########################################################
Tornado build success
[...]

PR #209, opened 2022-10-11
[tooling] tornado script for Win11 fixed

jjfumero:
`tornado` script fixed for Windows 11. The problem was that the PATH was using the Linux directory separator (`/`) instead of the Windows separator (`\`). 

This change has been tested for Windows 11 using MSYS64 for the OpenCL and PTX backends. 

Issue related #179

### DIFF
--- a/assembly/src/bin/tornado
+++ b/assembly/src/bin/tornado
@@ -30,6 +30,7 @@ import shlex
 import sys
 import re
 import argparse
+import platform
 
 # ########################################################
 # FLAGS FOR TORNADOVM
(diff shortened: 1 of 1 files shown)


stratika (approved):
hi @jjfumero, I tested it in Windows 10 with OpenCL and it builds and passes all unit-tests. It looks good to me.

mairooni (approved):
@jjfumero I also built the code and run the unittests, on the OpenCL backend, on Windows 10 and I verify that it works.

PR #279, opened 2023-11-20
[fix] fix --profiler flag in tornado-benchmarks.py script

ShAlireza:
#### Description

The --profiler flag didn't work, since when enabled by the user, the flag was added to JVM options, however, it should be placed in front of the tornado command and not JVM options.

Hence, the options variable is updated to two separate variables: jvm_options and tornado_options, so every option will be placed in its right place.

#### Problem description

JVM didn't recognize the --enableProfiler flag and showed the following error:

```
tornado --jvm="-Xms24G -Xmx24G -server -Dtornado.recover.bailout=False  -Dtornado.benchmarks.skipserial=True  --enableProfiler silent "  -m tornado.benchmarks/uk.ac.manchester.tornado.benchmarks.BenchmarkRunner --p
arams="saxpy"
Unrecognized option: --enableProfiler
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
You can regenerate the error by running the following command:

```
tornado-benchmarks.py --skipSequential --profiler
```

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [ ] No

You can run the tornado-benchmarks.py script just like the following example:

tornado-benchmarks.py --skipSequential --profiler console

And profiler output will be displayed on the console.

### DIFF
--- a/tornado-assembly/src/bin/tornado-benchmarks.py
+++ b/tornado-assembly/src/bin/tornado-benchmarks.py
@@ -231,7 +231,7 @@ def runBenchmarksFullCoverage(args):
 
 def runMediumConfiguration(args):
     jvm_options, tornado_options = composeAllOptions(args)
-    print(options)
+    print(tornado_options, jvm_options)
     for key in mediumSizes.keys():
         for size in mediumSizes[key][0]:
             numIterations = eval(mediumSizes[key][1][0])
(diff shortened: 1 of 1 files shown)


jjfumero:
Thank you for your contributions. LGTM. 

The way to check now is as follows:

```bash
$ tornado-benchmarks.py --skipSequential --profiler silent
```

It seems we also need to update the documentation here:https://tornadovm.readthedocs.io/en/latest/benchmarking.html#how-to-run  

We will do that.

jjfumero (approved):

mikepapadim (approved):
LGTM

PR #286, opened 2023-11-27
Move input sizes check to also check for tornado arrays

SirYwell:
#### Description

TornadoVM currently requires all inputs to have the same size when using batching.
However, the BatchConfiguration only checked that for java arrays before. That seems unintended.

The patch moves the check until after the loop to also cover TornadoVM array types.

I ran the tests locally and saw some failures, but from inspecting them, none of the failing tests uses batching, so the failures are likely unrelated.

#### Problem description

From reading the code, I noticed that the check only covers Java arrays. I wrote a test that ensures that mismatching sizes are discovered early instead of crashing the JVM.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run the added tests (potentially with the changes to BatchConfiguration reverted).

### DIFF
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
@@ -26,6 +26,7 @@ package uk.ac.manchester.tornado.runtime.common;
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 
 import uk.ac.manchester.tornado.api.types.arrays.ByteArray;
@@ -36,6 +37,7 @@ import uk.ac.manchester.tornado.api.types.arrays.IntArray;
 import uk.ac.manchester.tornado.api.types.arrays.LongArray;
 import uk.ac.manchester.tornado.api.types.arrays.ShortArray;
 import uk.ac.manchester.tornado.api.exceptions.TornadoRuntimeException;
+import uk.ac.manchester.tornado.api.types.arrays.TornadoNativeArray;
 import uk.ac.manchester.tornado.runtime.graph.TornadoExecutionContext;
 
 /**
@@ -69,56 +71,38 @@ public class BatchConfiguration {
         // Get the size of the batch
         List<Object> inputObjects = context.getObjects();
         long totalSize = 0;
-        DataTypeSize dataTypeSize = null;
 
-        HashSet<Class<?>> classObjects = new HashSet<>();
         HashSet<Long> inputSizes = new HashSet<>();
+        LinkedHashSet<Byte> elementSizes = new LinkedHashSet<>();
 
         for (Object o : inputObjects) {
             if (o.getClass().isArray()) {
                 Class<?> componentType = o.getClass().getComponentType();
-                dataTypeSize = findDataTypeSize(componentType);
+                DataTypeSize dataTypeSize = findDataTypeSize(componentType);
                 if (dataTypeSize == null) {
                     throw new TornadoRuntimeException("[UNSUPPORTED] Data type not supported for processing in batches");
                 }
                 long size = Array.getLength(o);
                 totalSize = size * dataTypeSize.getSize();
 
-                classObjects.add(componentType);
+                elementSizes.add(dataTypeSize.getSize());
                 inputSizes.add(totalSize);
-                if (classObjects.size() > 1) {
-                    throw new TornadoRuntimeException("[UNSUPPORTED] Input objects with different data types not currently supported");
-                }
-            } else if (o instanceof IntArray) {
-                totalSize = ((IntArray) o).getNumBytesWithoutHeader();
+            } else if (o instanceof TornadoNativeArray tornadoNativeArray) {
+                totalSize = tornadoNativeArray.getNumBytesWithoutHeader();
                 inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(int.class);
-            } else if (o instanceof FloatArray) {
-                totalSize = ((FloatArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(float.class);
-            } else if (o instanceof DoubleArray) {
-                totalSize = ((DoubleArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(double.class);
-            } else if (o instanceof LongArray) {
-                totalSize = ((LongArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(long.class);
-            } else if (o instanceof ShortArray) {
-                totalSize = ((ShortArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(short.class);
-            } else if (o instanceof ByteArray) {
-                totalSize = ((ByteArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(byte.class);
-            } else if (o instanceof CharArray) {
-                totalSize = ((CharArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(char.class);
+                byte elementSize = switch (tornadoNativeArray) {
+                    case IntArray _ -> DataTypeSize.INT.getSize();
+                    case FloatArray _ -> DataTypeSize.FLOAT.getSize();
+                    case DoubleArray _ -> DataTypeSize.DOUBLE.getSize();
+                    case LongArray _ -> DataTypeSize.LONG.getSize();
+                    case ShortArray _ -> DataTypeSize.SHORT.getSize();
+                    case ByteArray _ -> DataTypeSize.BYTE.getSize();
+                    case CharArray _ -> DataTypeSize.CHAR.getSize();
+                    default -> throw new TornadoRuntimeException("Unsupported array type: " + o.getClass());
+                };
+                elementSizes.add(elementSize);
             } else {
-                throw new TornadoRuntimeException("Unsupported type: ");
+                throw new TornadoRuntimeException("Unsupported type: " + o.getClass());
             }
         }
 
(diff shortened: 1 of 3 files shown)


SirYwell:
I also noticed that the type mismatch also only covers Java arrays, but I didn't address that for now (I assume this should also cover the Tornado array types?). Please let me know if you want me to address that too, and whether you want this to be a separate PR.

jjfumero:
Hi @SirYwell . Can you expand on the issue you are facing? 
So, by design, when using batches, TornadoVM splits in chunks of equal size. If this is possible (taken the total input sizes), then TornadoVM will run the application. Otherwise, only the last chunk gets resized. 

So, can you expand on the error you are getting? Do you have a usecase?

jjfumero:
Ok, by looking at the PR, I think I know what you meant. So you introduced a check to guarantee that the parameter arrays have the same size. Good catch. 

Can you please add the check also for the arrays? 
Thank you!

jjfumero:
I appreciate the clarification. I do agree with you that having different types with the same number of byte representations could potentially work (e.g., `int` and `float`). 

So the check must be for all input/output parameters using the same number of bytes and the same number of elements. 

Note that, although we still support Java arrays, we will deprecate this, so I will focus only on Tornado Arrays (E.g. `IntArray` etc). 
Internally, the layout representation of `int[]` and `IntArray` is the same for the TornadoVM JIT Compiler and runtime. So, I am expecting also this to work. There are minor differences (E.g., array header) but this logic is not included to process the batch size. 

So it would be great if we can fix this. To sum up:
- The check should be per size and type-compatibility (as in the number of bytes representation). 
- For the time being, we should allow both, primitive arrays and TornadoVM arrays. 

I suggest starting with more unit-tests to cover all these cases.

SirYwell:
> I suggest starting with more unit-tests to cover all these cases.

I added more tests. As expected, `testSameInputTypeRestriction` and `testSameInputSizeJavaArrays` fail. 

- The first one fails because it does not throw an exception although we have mismatching element sizes (IntArray vs LongArray, with the same total input size)
- The second one fails because of the check for the exact class type when using Java arrays (int[] and float[]).

Please let me know what you think. Also getting these tests succeed shouldn't be difficult.

If we support different element types, we also need to change the documentation I linked before, right?

jjfumero:
Thank you @SirYwell . This is very helpful. If you want, you can add the patch as well and update the `docs` with the updated list of limitations. 

One question, which backend/s are you using?

SirYwell:
> Thank you @SirYwell . This is very helpful. If you want, you can add the patch as well and update the `docs` with the updated list of limitations.

I updated the code and the docs now.
 
> One question, which backend/s are you using?

I'm currently running things via WSL2 with the PTX backend (from what I found, OpenCL doesn't work in WSL2 for now).

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
```diff
@@ -69,63 +71,48 @@ public static BatchConfiguration computeChunkSizes(TornadoExecutionContext conte
...
-                dataTypeSize = findDataTypeSize(short.class);
-            } else if (o instanceof ByteArray) {
-                totalSize = ((ByteArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(byte.class);
-            } else if (o instanceof CharArray) {
-                totalSize = ((CharArray) o).getNumBytesWithoutHeader();
-                inputSizes.add(totalSize);
-                dataTypeSize = findDataTypeSize(char.class);
+                byte elementSize = switch (tornadoNativeArray) {
```
SirYwell, line 93: This can be further simplified, e.g. I saw that the arrays all have the `SegmentElementSize` annotation on their type. Would it make sense to use it here?

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
```diff
@@ -69,63 +71,48 @@ public static BatchConfiguration computeChunkSizes(TornadoExecutionContext conte
...
+                    case IntArray _ -> DataTypeSize.INT.getSize();
+                    case FloatArray _ -> DataTypeSize.FLOAT.getSize();
+                    case DoubleArray _ -> DataTypeSize.DOUBLE.getSize();
+                    case LongArray _ -> DataTypeSize.LONG.getSize();
+                    case ShortArray _ -> DataTypeSize.SHORT.getSize();
+                    case ByteArray _ -> DataTypeSize.BYTE.getSize();
+                    case CharArray _ -> DataTypeSize.CHAR.getSize();
+                    default -> throw new TornadoRuntimeException("Unsupported array type: " + o.getClass());
```
SirYwell, line 101: This is needed as TornadoNativeArray isn't sealed. Not sure if you considered it before, but I don't know if there is any reason for custom implementations of either TornadoNativeArray or any of the special array types.

jjfumero:
Testing this PR on Apple M2:

```bash
 tornado -ea  --jvm "-Xmx6g -Dtornado.recover.bailout=False -Dtornado.unittests.verbose=True "  -m  tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  --params "uk.ac.manchester.tornado.unittests.batches.TestBatches"

Test: class uk.ac.manchester.tornado.unittests.batches.TestBatches
	Running test: test50MBInteger            ................  [PASS] 
	Running test: testSameInputSize          ................  [PASS] 
	Running test: test100MBSmall             ................  [PASS] 
	Running test: test100MB                  ................  [PASS] 
	Running test: test300MB                  ................  [PASS] 
	Running test: test512MB                  ................  [PASS] 
	Running test: test50MB                   ................  [PASS] 
	Running test: test50MBShort              ................  [PASS] 
	Running test: test50MBDouble             ................  [FP64 UNSUPPORTED FOR CURRENT DEVICE] 
	Running test: test50MBLong               ................  [FAILED] 
		\_[REASON] Unable to allocate 50000024 bytes of memory.
	Running test: testSameInputSizeAndTypeRestriction ................  [PASS] 
	Running test: testSameInputTypeRestrictionJavaArrays ................  [PASS] 
	Running test: testSameInputSizeAndTypeJavaToTornado ................  [PASS] 
	Running test: testSameInputSizeAndTypeTornadoToJava ................  [PASS] 
	Running test: testSameInputTypeRestriction ................  [PASS] 
	Running test: testSameInputSizeJavaArrays ................  [PASS] 
	Running test: testSameInputSizeJavaToTornado ................  [PASS] 
	Running test: testSameInputSizeTornadoToJava ................  [PASS] 
	Running test: testSameInputSizeAndTypeRestrictionJavaArrays ................  [PASS] 
```

Errors are not related to this PR, but rather unsupported features from this  OpenCL implementation.

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
jjfumero: We only want to use the annotation part in the compiler side, not in the runtime. We try to avoid reflection as much as possible.

jjfumero:
This PR looks good to me. I will check on Linux with the OpenCL backend before merge.

jjfumero:
Related to Windows WSL, I could not make it work with NVIDIA and OpenCL. What I do is to build with OpenCL from the Intel oneAPI runtime. This gives me access to the OpenCL implementation using the Intel HD Graphics. If you have one, this could be an option.

jjfumero:
Tests on Linux with the OpenCL backend passing:

```bash
tornado -ea  --jvm "-Xmx6g -Dtornado.recover.bailout=False -Dtornado.unittests.verbose=True "  -m  tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  --params "uk.ac.manchester.tornado.unittests.batches.TestBatches"

Test: class uk.ac.manchester.tornado.unittests.batches.TestBatches
	Running test: testSameInputSizeAndTypeRestriction ................  [PASS] 
	Running test: testSameInputTypeRestriction ................  [PASS] 
	Running test: testSameInputSizeJavaToTornado ................  [PASS] 
	Running test: testSameInputSizeJavaArrays ................  [PASS] 
	Running test: testSameInputSizeTornadoToJava ................  [PASS] 
	Running test: test100MBSmall             ................  [PASS] 
	Running test: test100MB                  ................  [PASS] 
	Running test: test300MB                  ................  [PASS] 
	Running test: test512MB                  ................  [PASS] 
	Running test: test50MB                   ................  [PASS] 
	Running test: test50MBInteger            ................  [PASS] 
	Running test: test50MBShort              ................  [PASS] 
	Running test: test50MBDouble             ................  [PASS] 
	Running test: test50MBLong               ................  [PASS] 
	Running test: testSameInputSizeAndTypeRestrictionJavaArrays ................  [PASS] 
	Running test: testSameInputTypeRestrictionJavaArrays ................  [PASS] 
	Running test: testSameInputSize          ................  [PASS] 
	Running test: testSameInputSizeAndTypeJavaToTornado ................  [PASS] 
	Running test: testSameInputSizeAndTypeTornadoToJava ................  [PASS] 
Test ran: 19, Failed: 0, Unsupported: 0

```

jjfumero (approved):

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
SirYwell: Sorry, I was referring to the `default` case. The type-specific implementations make sense, but they are extendable (I could write `MyCustomIntArray extends IntArray`) which most likely doesn't make sense, and TornadoNativeArray is extendable as well (I could write `MyCustomWhateverArray extends TornadoNativeArray`) which most likely doesn't make sense either.

My suggestion was to make TornadoNativeArray `sealed`, only allow the type-specific implementations as subtypes, and make the subtypes final. This way, the `default` case here could be removed, and introducing a new type-specific implementation (e.g. `HalfFloatArray`) would be a java compiler error in this switch rather than only failing at runtime as soon as someone wants to use it. But this is something that can be addressed separately from this PR I guess.

SirYwell:
> Related to Windows WSL, I could not make it work with NVIDIA and OpenCL. What I do is to build with OpenCL from the Intel oneAPI runtime. This gives me access to the OpenCL implementation using the Intel HD Graphics. If you have one, this could be an option.

I don't have Intel HD Graphics, but I could probably try to get it running on Windows directly.

But it looks like the current tests are looking good enough for this PR?

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/BatchConfiguration.java
jjfumero: This is great feedback. I think makes sense. Let me analyze it first with the current plans and use cases for TornadoVM.

mikepapadim (approved):

PR #288, opened 2023-11-29
Fix aws s3 configuration usage in aws_post_processing.sh

ShAlireza:

#### Description

Currently, tornado gets a configuration file for running codes on fpga.
On AWS instances, the configuration needs some metadata about S3 bucket
and related folders to upload files on S3, however, aws_post_processing.sh
is not using these metadata properly. This fix uses the defined variables
to run create_vitis_afi.sh

#### Problem description

If user has different bucket and folder names in their S3, tornado 
will fail to upload generated files to S3. So, this will prevent create_vitis_afi.sh 
script to generate AFI image for FPGA on AWS and the image will not be 
available for tornadovm to run on the FPGA device.  Also, any additional
white space after the equal sign and the value, will be propagated to the
create_vitis_afi.sh and results in errors.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [x] Yes
- [ ] No

Compiling the tornadovm should generate the correct form of aws_post_processin.sh
in bin/bin and bin/sdk/bin folders. Then using any name for S3 bucket and 
dcp/logs folder keys should work properly and the AMI image should be generated.

### DIFF
--- a/tornado-assembly/src/bin/aws_post_processing.sh
+++ b/tornado-assembly/src/bin/aws_post_processing.sh
@@ -4,9 +4,9 @@ configurationFilePath=$1
 directoryBitstream=$2
 entryPoint=$3
 
-export S3_BUCKET=$(grep -w "AWS_S3_BUCKET" ${configurationFilePath} | awk -F'[[:space:]]=[[:space:]]' '{print $2}')
-export S3_DCP_KEY=$(grep -w "AWS_S3_DCP_KEY" ${configurationFilePath} | awk -F'[[:space:]]=[[:space:]]' '{print $2}')
-export S3_LOGS_KEY=$(grep -w "AWS_S3_LOGS_KEY" ${configurationFilePath} | awk -F'[[:space:]]=[[:space:]]' '{print $2}')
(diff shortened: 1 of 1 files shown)


jjfumero:
Thank you @ShAlireza for digging into this. We will test this as well.

stratika:
good catch, I will test it and will let you know.

stratika:
@ShAlireza it appears to be an error in the Vitis script regarding the passing parameters. A space is added for each of the three arguments (`"$S3_BUCKET"`, `"$S3_DCP_KEY"`, `"$S3_LOGS_KEY"`), and this causes the script to not be able to upload the generated bitstream to S3. See below:
```bash
aws s3  cp 23_11_29-121533_Developer_Vitis_Kernel.tar s3:// tornadovm-fpga-bucket/ outputfolder/

Unknown options: tornadovm-fpga-bucket/,outputfolder/
```

Can you please confirm whether you have encountered that issue?

stratika (approved):
After the latest patch from @ShAlireza, I confirm that it works. We can proceed and merge.

Issue #296, opened 2023-12-12
Separate the Level Zero JNI lib from the internals of TornadoVM


jjfumero:
The JNI Level Zero should be separated from the TornadoVM main repo and pull from the official Beehivelab repo instead.
We can do this since the Level Zero JNI library is sufficient and changes are rare. Thus, we maintain the level zero lib using the official repo and pull the corresponding version in the TornadoVM project. 

This task involves:
a) Remove the library from the TornadoVM internals
b) Update the installation script to pull the necessary directories and build the Java project
c) Update the maven assembler module to obtain the `.so` library and copy to the TornadoVM SDK along with the JAR files.

jjfumero:
This has been merged: https://github.com/beehive-lab/TornadoVM/pull/297

PR #297, opened 2023-12-12
[dependencies] Decouple of the Level Zero JNI lib

jjfumero:
#### Description

This PR decouples the internal Level Zero JNI library.  Instead, it build the library by pulling the latest changes from the official repo. This change makes easier to maintain the external library from now on, since the JNI lib contain the method methods required by  the TornadoVM SPIR-V runtime to operate.
    
Related issue: #296 (Public)

#### Backend/s tested

This PR apples only to the SPIR-V backend

- [ ] OpenCL
- [ ] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
## 1. Clean the m2 repos 
rm -Rf ~/.m2/repositories 

## Build SPIR-V Backend of TornadoMV 
make BACKEND=spirv

## Pass the unittests
make tests
```


mikepapadim (approved):
LGTM, only one nit. Maybe, we could use `stdout=PIPE` on subprocess.run to get the process output when running these commands.

We could do that, in other iterration.

mairooni (approved):
LGTM

PR #300, opened 2023-12-19
[fix][install] Python module deps fixed. SPIR-V installer improved

jjfumero:
#### Description

This PR improves the installation process for all backends and includes a cache compilation of all SPIR-V package/module dependencies. 
We do not need to pull and rebuild all SPIR-V dependencies with every TornadoVM build. 
We can force to rebuild the dependencies by running:

```bash
make rebuild-deps-jdk21 BACKEND=spirv
```

Otherwise, if the depdenceis are not present, it will only download and build the first time.
This speeds up the TornadoVM build process for SPIR-V. 

#### Problem description

Related issue #299 

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
## remove the dependencies to trigger the automatic download and build
$ rm -Rf levelzero-jni/ level-zero/ spirv-beehive-toolkit/
$ make BACKEND=spirv
```

To force rebuild all dependencies:

```bash
$ make rebuild-deps-jdk21 BACKEND=spirv
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ BACKENDS := --backends $(BACKEND)
 build jdk21:
 	bin/compile --jdk jdk21 $(BACKENDS)
 
+rebuild-deps-jdk21:
+	bin/compile --jdk jdk21 --rebuild $(BACKENDS)
+
 graal-jdk-21:
 	bin/compile --jdk graal-jdk-21 $(BACKENDS)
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -20,9 +23,6 @@ ptx:
 spirv:
 	bin/compile --jdk jdk21 --backends spirv,ptx,opencl
 
-offline:
```
mikepapadim, line 23: why we removed the offline build?

mikepapadim (commented):
LGMT
One thing, if we removed the offline option from the Makefile. 
Should we clean it also from the compile.py? cause it expect it as an argument.

Makefile
jjfumero: This rule is not used now. Even without connection, and assuming all dependencies are already installed, we can still build.

mikepapadim: thanks
should we clean it from the compile script as well?

jjfumero: Yes, good catch.

mikepapadim (approved):
LGTM, thanks

stratika (approved):

PR #313, opened 2024-01-23
Support half precision floating point types 

mairooni:
#### Description

This PR provides support for half-float types, i.e. float values represented by 16 bits instead of 32. Since Java does not currently offer a half-float type, in this implementation it is represented by a new class, `HalfFloat`.
The `HalfFloat` class offers two constructors. The first receives a float value, which is internally converted to float-16,  using the `Float.floatToFloat16` function. The second directly receives a short value, which represents a half-float. 
The `HalfFloat` class exposes a set of operations that can be performed between half-float values, specifically addition, subtraction, multiplication, and division.  
Finally, a new off-heap array type for half-float values (`HalfFloatArray` class) has been included. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Tests have been added for the following scenarios:

1) Initialization of a new `HalfFloat` instance in the kernel:
`tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestArrays#testHalfFloatInitialization`
2) Addition of two HalfFloatArrays:
`tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestArrays#testVectorAdditionHalfFloat`
3) Subtraction of two HalfFloatArrays:
`tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestArrays#testVectorSubtractionHalfFloat`
4) Multiplication of two HalfFloatArrays:
`tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestArrays#testVectorMultiplicationHalfFloat`
5) Division between two HalfFloatArrays: 
`tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestArrays#testVectorDivisionHalfFloat`
6) Test the `fromElements` and` fromArray` functions of the `HalfFloatArray` class:
`tornado-test -V uk.ac.manchester.tornado.unittests.api.TestAPI#testSegmentsHalfFloats`
7) Test the `fromSegment` function of the `HalfFloatArray` class:
`tornado-test -V uk.ac.manchester.tornado.unittests.api.TestAPI#testBuildWithSegmentsHalfFloat`


bin/compile
```diff
@@ -191,7 +191,7 @@ def build_spirv_toolkit_and_level_zero(rebuild=False):
 
     if (rebuild or build):
         os.chdir(spirv_tool_kit)
-        subprocess.run(["git", "pull", "origin", "master"])
```
mikepapadim, line 194: I think this needs to be reverted

tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/HalfFloat.java
```diff
@@ -0,0 +1,190 @@
...
+     * Takes two {@code HalfFloat} objects and returns a new {@HalfFloat} instance
+     * that contains the results of the multiplication.
+     *
+     * @param a
+     *     The first {@code HalfFloat} input for the multiplication.
+     * @param b
+     *     The second {@code HalfFloat} input for the multiplication.
+     * @return A new {@HalfFloat} containing the results of the multiplication.
+     */
+    public static HalfFloat mult(HalfFloat a, HalfFloat b) {
```
mikepapadim, line 155: nit: any reason we called it mult and mul

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/asm/OCLAssembler.java
```diff
@@ -74,6 +74,8 @@ public OCLAssembler(TargetDescription target) {
             emitLine("#pragma OPENCL EXTENSION cl_khr_fp64 : enable  ");
         }
 
+        emitLine("#pragma OPENCL EXTENSION cl_khr_fp16 : enable  ");
```
mikepapadim, line 77: so, now we are going to emit this in all opencl kernels we generate, regardless if it uses half floating point?

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/asm/OCLAssembler.java
```diff
@@ -74,6 +74,8 @@ public OCLAssembler(TargetDescription target) {
...
+
         if (((OCLTargetDescription) target).supportsInt64Atomics()) {
```
mikepapadim, line 79: can we have a similar condition instead?

bin/compile
jjfumero: We keep this until we merge the code from the Beehive SPIR-V Toolkit

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/asm/OCLAssembler.java
jjfumero: I think we should emit this line only if the platform supports it.

jjfumero (changes requested):
Let's add the check for FP16 in OpenCL (pragma)

jjfumero:
Unit-test are passing for SPIR-V and PTX. For OpenCL, depending on the device, this passes. 

In my case, using RTX 2060 and the NVIDIA Drivers 510.54 do not pass.

mairooni:
I included a function that checks if FP16 is supported. However, since in my machine I do not have this capability but I can pass the tests for OpenCL, we decided with @jjfumero to keep the printing of the OpenCL pragma for the time being, regardless of whether it is supported, until we figure out why this is happening.

jjfumero:
Yes, we can query the capability. The pragma can always be generated. We need to investigate how to run OpenCL FP16 on modern NVIDIA GPUs.

jjfumero (approved):

mikepapadim:
Ok, lets merge it then and we can iterate for a fix in codegen when they add the extension in the driver

mikepapadim (approved):

jjfumero:
It appears a new conflict. @mairooni, can you take a look?

PR #332, opened 2024-02-08
Batch Processing with Multiple tasks in a task-graph

jjfumero:
#### Description

This fixes and adds support for batch processing when having a `TaskGraph` with multiple tasks. 
Besides, it removes old/unreachable methods that are not needed anymore. 

However, this PR still has the issue of incorrect results when the task-graph does not perform a copy-in. 
I added two more unit-tests, one with copy in and the other without. It seems that without copy-in, the batch processing works (compiler and internals of the runtime), but the results are not correct. 

For task-graphs with copy-in/out (which was the original intended use), multiple tasks in a taskgraph with batch processing now works. 
We can go ahead this is partial fix, and add a new extension on top of this to process also for only outputs. 

#### Problem description

The issue was that, when having multiple tasks in a task-graph, the TornadoVM interpreter was not invalidating the code cache (to force recompilation). 

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ make
$ tornado-test -V --fast --debug --threadInfo --printKernel --printBytecode uk.ac.manchester.tornado.unittests.batches.TestBatches#testBatchNotEven
$ tornado-test -V --fast --debug --threadInfo --printKernel --printBytecode uk.ac.manchester.tornado.unittests.batches.TestBatches#testBatchNotEven2
```

### DIFF
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
@@ -542,22 +542,6 @@ public class TornadoTaskGraph implements TornadoTaskGraphInterface {
 
     }
 
-    private void triggerRecompile() {
-        // 1. Force to recompile the task-sketcher
-        int i = 0;
-        for (TaskPackage tp : taskPackages) {
-            updateTask(tp, i);
-            i++;
-        }
-
-        // 2. Clear the code caches in every instance of a TornadoVMInterpreter that
-        // TornadoVM instantiated
-        if (vm != null) {
-            vm.clearInstalledCode();
-            vm.setCompileUpdate();
-        }
-    }
-
     @Override
     public TornadoAcceleratorDevice getDeviceForTask(String id) {
         return executionContext.getDeviceForTask(id);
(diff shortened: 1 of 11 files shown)


tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
```diff
@@ -542,22 +542,6 @@ public void setDevice(String taskName, TornadoDevice device) {
 
     }
 
-    private void triggerRecompile() {
```
mikepapadim, line 545: What happend with this ?

mikepapadim (approved):
LGMT

mairooni (approved):

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
jjfumero: This method was not used.

mairooni: This was never used

PR #346, opened 2024-03-07
[fix][spirv] Reset LocalWorkSize when the device is changed

jjfumero:
#### Description

This PR solves the problem of getting the wrong local work group sizes when the device is changed from the CPU using the OpenCL backend to the GPU using the SPIR-V backend. 

#### Problem description

The problem was that, in the SPIR-V backend, once the number of blocks is computed, is then stored in the `TaskMetaData` object. However, while the global work size for the same task will not change, there is no guarantee that the local work group size is the same. This patch solves the issue by resetting the local work size only when the device is updated. 
 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

The use case for this is the task-migration in Client/Server setup:

https://github.com/jjfumero/tornadovm-examples?tab=readme-ov-file#live-task-migration-client-server-app

```bash
## Run Server in one terminal
./runServer.sh

## Client in another terminal
./runClient.sh  ## Change device during runtime 

## Note: the application selects the backend 0:0 (default backend)

# type different <backend:deviceNumber> version from the client. 
# Examples:
# 0:1 
# 1:0 
## etc
```

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/memory/TaskMetaDataInterface.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/memory/TaskMetaDataInterface.java
@@ -55,4 +55,6 @@ public interface TaskMetaDataInterface {
     boolean isPrintKernelEnabled();
 
     void setPrintKernelFlag(boolean printKernelEnabled);
+
+    void resetThreadBlocks();
 }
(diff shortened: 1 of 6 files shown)


mairooni (approved):
LGTM

stratika (approved):
LGTM

PR #357, opened 2024-03-20
Extension of the TornadoExecutionPlan API to allow backend and device filters from the API

jjfumero:
#### Description

This PR adds an extension to the `TornadoExecutionPlan` API in TornadoVM to allow developers to query backends and devices by applying filters. This API is designed as a consequence of the inconsistencies when obtaining a device using different APIs (more details in the next Section). 

This API fixes this inconsistencies by giving the user a new Data Structure, called `TornadoDeviceMap` in which developers can query all backends, devices and apply filters. 

Examples of use:

```java
// Obtain an instance of the TornadoDeviceMap via the execution plan
TornadoDeviceMap tornadoDeviceMap = TornadoExecutionPlan.getTornadoDeviceMap();

// Obtain the number of backends installed
int numBackends = tornadoDeviceMap.getNumBackends();


// Query all backends installed
List<TornadoBackend> backends = tornadoDeviceMap.getAllBackends();

// Obtain the number of devices for the first backend
int numDevicesBackendZero = backends.getFirst().getDeviceCount();
```

We can apply more complex queries as follows:

```java
TornadoDeviceMap tornadoDeviceMap = TornadoExecutionPlan.getTornadoDeviceMap();

// Obtain the OpenCL Backend
List<TornadoBackend> openCLBackend = tornadoDeviceMap.getBackendWithPredicate(backend -> backend.getBackendType() == TornadoVMBackendType.OPENCL);

// Obtain Backends with at least, two devices
List<TornadoBackend> multiDeviceBackends = tornadoDeviceMap.getBackendWithPredicate(backend -> backend.getDeviceCount() > 1);


// Obtain All backends that can access to an NVIDIA GPU
List<TornadoBackend> backendsWithNVIDIAAccess = tornadoDeviceMap.getBackendWithDevicePredicate(device -> device.getDeviceName().toLowerCase().contains("nvidia"));
```

As part of this API, we needed to refactor some internal classes. The concept of Driver is not the same as we have in TornadoVM compared to the OS driver. Thus, all old driver classes became the Backend classes. 

#### Problem description

The main problem is that to query the number of devices, developers need to access through the `TornadoRuntime.getRuntime` class.  However, to select a device or a backend, developers need to the ExecutionPlan.

This PR fixes this inconsistencies by providing a more concise and expressive API to filter devices and backends without the need to instantiate or interact with the TornadoVM Core Runtime (internal class). 

#### Backend/s tested

Mark the backends affected by this PR.

- [X] OpenCL
[...]

PR #358, opened 2024-03-20
Fix Native Test to pass on modern platforms with no support of FP64

jjfumero:
#### Description

The purpose of the native tests is to check that the middle-ware API works as expected, not to check if we can run FP64. Since many OpenCL platforms these days do not support FP64, we do not need to check this in this unit-tests. This PR just replaces the `double` with `float` and test the interaction with the JIT compiler and runtime using the internal APIs. 

Note that this PR is built in top of #357 .

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [X] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ make 
$ test-native.sh 

Testing the Native OpenCL API
Running Native: uk.ac.manchester.tornado.drivers.opencl.tests.TestOpenCLJITCompiler ................ [PASS]
```


mairooni (approved):

jjfumero:
This is ready for review

stratika (approved):
LGTM As you said, the purpose of the native tests is not to test precision, but the functionality of the low-level APIs.

mikepapadim (approved):

PR #369, opened 2024-03-29
Introduce Tensor API, Tensor Utilities and compatiblity with ONNX RT

mikepapadim:
#### Description

This pull request introduces a new Tensor API v0.1, providing a comprehensive set of classes and utilities for working with tensors of various data types along with compatibility with ONNX RT Java bindings. 

**The primary changes include:**
- **AbstractTensor Interface**: An abstract base class that defines the common contract for all tensor implementations.
- **Tensor Classes**: A set of concrete tensor classes for different data types, including TensorFloat32, TensorFloat16, TensorInt32, TensorInt16, TensorInt64, TensorFloat64, and TensorByte. These classes extend the AbstractTensor interface and provide type-specific functionality.
- **DType Class**: A utility class that encapsulates data type information for tensors, facilitating type-safe operations and conversions.
- **Shape Class**: A class representing the shape (dimensions) of a tensor.

Also, this PR adds test to show compatibility with ONNX RT Java bindings->

```java
  public void testOnnxCompatibility() throws OrtException {
        Shape shape = new Shape(1, 3, 224, 224);
        TensorFloat32 tornadoTensor = new TensorFloat32(shape);

        tornadoTensor.init(2f);

        OnnxTensor outputTensor = null;

        try (OrtEnvironment env = OrtEnvironment.getEnvironment()) {
            // Load the MobileNet V2 ONNX model
            OrtSession session = env.createSession(MODEL_PATH, new OrtSession.SessionOptions());

            OnnxTensor inputTensor = OnnxTensor.createTensor(env, tornadoTensor.getFloatBuffer(), shape.dimensions());

            Map<String, OnnxTensor> inputMap = new HashMap<>();
            inputMap.put(INPUT_TENSOR_NAME, inputTensor);

            // Run the model inference
            try (OrtSession.Result outputMap = session.run(inputMap)) {
                Optional<OnnxValue> optionalOutputTensor = outputMap.get(OUTPUT_TENSOR_NAME);
                if (optionalOutputTensor.isEmpty()) {
                    throw new IllegalArgumentException("Output tensor not found in model output.");
                }
                outputTensor = (OnnxTensor) optionalOutputTensor.get();

            }
        } finally {
            Assert.assertNotNull(outputTensor);
        }
    }
```

**Overview of the Tensor API Architecture:**
TensorInt64

**Notes:**
It adds a dependency to ONNX runtime in the `tornado-unittests` packages, as an alternative we could move the compatibility tests into a seperate repo or submodule.

[...]

PR #373, opened 2024-04-09
Handle cases where the class field is a half float array

mairooni:
#### Description

This PR handles the cases where a class field is an instance of a `HalfFloatArray`. This case was highlighted by the new Tensor implementation (PR #369). The patch is applied for all three backends. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

Run `tornado-test -V uk.ac.manchester.tornado.unittests.tensors.TestTensorTypes`


mikepapadim (approved):
LGTM, thanks @mairooni

jjfumero (approved):

PR #378, opened 2024-04-16
Fix for the free memory marker of TornadoVM when running multi-threaded execution plans

jjfumero:
#### Description

This PR fixes the multi-threaded deallocation marker within the TornadoVM runtime. This PR also fixes the code in the GAIA project. 

#### Problem description

The issue was that we need to guarantee that the deallocator marker is executed in sychronized mode, and guarantee that the `unused` buffer item in the list was not `null`.  It can be `null` if it references the same host memory object (e.g., the same Panama-based array). 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
$ make 
$ tornado-test -V --fast uk.ac.manchester.tornado.unittests.multithreaded.TestMultiThreadedExecutionPlans
```

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/memory/ObjectState.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/memory/ObjectState.java
@@ -21,7 +21,7 @@ import uk.ac.manchester.tornado.api.common.TornadoDevice;
 
 public interface ObjectState {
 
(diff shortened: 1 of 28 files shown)


mairooni (approved):
LGTM

mikepapadim (approved):
Thanks

PR #400, opened 2024-04-30
Fix batch computation when the global index is written in the output buffer

mairooni:
#### Description
This PR provides a fix for a corner case in batch processing, which was showcased by the  `uk.ac.manchester.tornado.unittests.batches.TestBatches.testBatchNotEven` unittest. The issue occurs when the loop index is directly written in the output buffer, e.g.: 
```
for (@Parallel int i = 0; i < data.getSize(); i++) {
      data.set(i, i);
}
```

Since, in batch processing, the loop bound of each kernel is equal to the batch size, the `i` written in the output buffer is correct only for the first batch. To solve this, the value of `i` is offseted based on the number of the batch. 
For instance, the generated code for the example shown above is changed as follows: 
`data.set(i, i) -> data.set(i, i + batchNumber * BatchSize)`

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run `tornado-test -V --fast uk.ac.manchester.tornado.unittests.batches.TestBatches`


tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/phases/TornadoHighTierContext.java
```diff
@@ -37,15 +37,19 @@ public class TornadoHighTierContext extends HighTierContext {
     protected final TaskMetaData meta;
     protected final boolean isKernel;
     private long batchThreads;
+    private int batchNumber;
+    private long batchSize;
 
     public TornadoHighTierContext(Providers providers, PhaseSuite<HighTierContext> graphBuilderSuite, OptimisticOptimizations optimisticOpts, ResolvedJavaMethod method, Object[] args,
-            TaskMetaData meta, boolean isKernel, long batchThreads) {
+            TaskMetaData meta, boolean isKernel, long batchThreads, int batchNumber, long batchSize) {
```
mikepapadim, line 44: Can we have a  `BatchConfig` class or maybe a record to pass around long batchThreads, int batchNumber, long batchSize?

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/sketcher/Sketch.java
```diff
@@ -50,4 +52,12 @@ public Access[] getArgumentsAccess() {
         return argumentsAccess;
     }
 
+    public boolean isIndexInWrite() {
+        return indexInWrite;
+    }
+
+    public void indexInWrite() {
```
jjfumero, line 59:
```suggestion
    public void getIndexInWrite() {
```

jjfumero: If you do this change, you need to refactor all dependencies.

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/sketcher/TornadoSketcher.java
```diff
@@ -180,7 +180,11 @@ private static Sketch buildSketch(ResolvedJavaMethod resolvedMethod, Providers p
                 mergeAccesses(methodAccesses, invoke.callTarget(), sketch.getArgumentsAccess());
             });
 
-            return new Sketch(graph.copy(TornadoCoreRuntime.getDebugContext()), methodAccesses);
+            Sketch sketch = new Sketch(graph.copy(TornadoCoreRuntime.getDebugContext()), methodAccesses);
```
jjfumero, line 183: Pass the `indexInWrite` to the `Sketch` constructor. Also, I think the name should be changed. This is more like `batchWriteThreadIndex` and the method for get and set should have `get` and `set` as prefix.

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/CompilableTask.java
```diff
@@ -144,6 +146,26 @@ public void setBatchThreads(long batchThreads) {
...
+        this.batchNumber = batchNumber;
+    }
+
+    @Override
+    public int getBatchNumber() {
+        return this.batchNumber;
+    }
+
+    @Override
+    public void setBatchSize(long batchSize) {
```
jjfumero, line 160: What is the difference between batchSIze and batchThread

```java
void setBatchThreads(long batchThreads);

void setBatchSize(long batchSize);
```

jjfumero (changes requested):
I agree with @mikepapadim . Let's create a class called `BatchCompilationConfig` and store all the related fields here. Then, what we pass around in is an object of type `BatchCompilationConfig`.

jjfumero:
What happens when you have an expression that uses the thread-ID to store a value?

For example:

```java
for (@Parallel int i = 0; i < data.getSize(); i++) {
      data.set(i, i * 20 + beta);
}
```

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/CompilableTask.java
mairooni: The value of `batchThreads` changes when the chunks are not even, the `batchSize` is always the initial batch size. For example, if we have 3 batches, 60000, 60000 and 20000, the batchSize will be 60000 throughout the computation, even though the batch threads will change from 60000 to 20000 when computing the final chunk.

mairooni:
> What happens when you have an expression that uses the thread-ID to store a value?
> 
> For example:
> 
> ```java
> for (@Parallel int i = 0; i < data.getSize(); i++) {
>       data.set(i, i * 20 + beta);
> }
> ```

This case also works, I just included a unittest to check. To test it, run 
`tornado-test -V --fast uk.ac.manchester.tornado.unittests.batches.TestBatches#testBatchThreadIndex`

mairooni:
All comments have been applied. This PR is ready for another review.

jjfumero (changes requested):
Missing license header

mairooni:
> Missing license header

Done

jjfumero (approved):

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/phases/sketcher/TornadoBatchFunctionAnalysis.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2024 APT Group, Department of Computer Science,
```
stratika, line 2:
```suggestion
 * Copyright (c) 2024, APT Group, Department of Computer Science,
```

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoBatchGlobalIndexOffset.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright (c) 2024 APT Group, Department of Computer Science,
```
stratika, line 2:
```suggestion
 * Copyright (c) 2024, APT Group, Department of Computer Science,
```

stratika: We forgot the comma after the year. Please append all files that have new headers.

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLHighTier.java
```diff
@@ -77,8 +78,8 @@ public OCLHighTier(OptionValues options, TornadoDeviceContext deviceContext, Can
                 appendPhase(new IterativeConditionalEliminationPhase(canonicalizer, false));
             }
         }
-
         appendPhase(new TornadoTaskSpecialisation(canonicalizer));
+        appendPhase(new TornadoBatchGlobalIndexOffset());
```
stratika, line 82: You mentioned that this phase is batch-specific. However, it is invoked also when batch processing is not used. Right?

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/compiler/TornadoSketchTier.java
```diff
@@ -83,6 +84,7 @@ public TornadoSketchTier(OptionValues options, CanonicalizerPhase.CustomSimplifi
         appendPhase(new TornadoAutoParalleliser());
         appendPhase(new TornadoDataflowAnalysis());
         appendPhase(new TornadoPanamaPrivateMemory());
+        appendPhase(new TornadoBatchFunctionAnalysis());
```
stratika, line 87: This phase will be invoked for all cases (even when batch processing is not enabled), right?

mikepapadim (approved):

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/compiler/TornadoSketchTier.java
mikepapadim: It seems there is a check inside the phase to prevent from running it everytime.

stratika:
From my side, the suggestion for the header parts of the new files and some comments that have not been answered.

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/compiler/TornadoSketchTier.java
stratika: which check is this?

mairooni: This phase in the sketcher just checks if the loop index is directly written into the output buffer and it does not perform any changes in the graph. Since we do not pass any information related to batches to the sketches phases, this check is always performed. However, the phase that actually alters the graph (`TornadoBatchGlobalIndexOffset`) checks if there is batch processing before applying any changes 

```
if (batchSize == 0) {
            return;
}
```
If the `batchSize` has not been set, there is no batch processing.

mairooni:
I added the commas in the headers.

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graal/compiler/TornadoSketchTier.java
jjfumero: The check inside the phase is correct. We want to append all phases to the SketchTier, and then select it using runtime information.

stratika: excellent, thanks @mairooni. I was mislead and looking into the `TornadoBatchFunctionAnalysis` class.

PR #422, opened 2024-05-14
[fix] Exclude the nvml library in cmake if platform is jetson nano as…

stratika:
#### Description

This PR aims to fix the execution with `PTX` backend on `NVIDIA Jetson Nano`. Since the initial integration with the NVIDIA Management Library (NVML) in the native OpenCL and PTX backends for the monitoring of power usage, we have had the linking of the `nvml` library as default for when we build with the PTX backend. This was because, it requires the NVIDIA CUDA Toolkit to be installed.

However, we figured it out that for NVIDIA Jetson Nano the `nvidia-smi` and the `nvml` capabilities are not supported.

For instance an example is as follows:
```bash
tornado --devices

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING:

You should always run with libnvidia-ml.so that is installed with your
NVIDIA Display Driver. By default it's installed in /usr/lib and /usr/lib64.
libnvidia-ml.so in GDK package is a stub library that is attached only for
build purposes (e.g. machine that you build your application doesn't have
to have Display Driver installed).
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
[TornadoVM-PTX-NVML-JNI] ERROR : nvmlInit -> Returned: 9

Number of Tornado drivers: 1
Driver: PTX
  Total number of PTX devices  : 1
  Tornado device=0:0  (DEFAULT)
	PTX -- PTX -- NVIDIA Tegra X1
		Global Memory Size: 3.9 GB
		Local Memory Size: 48.0 KB
		Workgroup Dimensions: 3
		Total Number of Block Threads: [2147483647, 65535, 65535]
		Max WorkGroup Configuration: [1024, 1024, 64]
		Device OpenCL C version: N/A
``` 

Therefore this PR provides a patch that will ignore the `nvml` functions if the platform is Jetson Nano.

#### Backend/s tested

- [ ] OpenCL
- [x] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=ptx
tornado --enableProfiler console -m tornado.examples/uk.ac.manchester.tornado.examples.VectorAddInt --params="100000"
```


jjfumero:
I confirm it works.

stratika:
> I confirm it works on the Jetson nano.

Thanks, I have also tested it on my Dell laptop that it does not break that setup.

mikepapadim (approved):

PR #425, opened 2024-05-16
Vector types for Float<With> with Int<Width> supported at the API-level

jjfumero:
#### Description

This PR adds Float<Size> vector operations to be operate with Int<Size> vector types.
Due to OpenCL/SPIR-V restrictions, the actual operation in native code can't be done using explicit vector types. But still it is possible to operate by accessing the individual fields of the vector.

#### Problem description

Example could not use vector operations because of mismatch types. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

Compile and run for all backends:

```bash
tornado-test --printKernel -V uk.ac.manchester.tornado.unittests.vectortypes.TestFloats
```

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/vectortypes/TestFloats.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/vectortypes/TestFloats.java
@@ -1071,9 +1071,10 @@ public class TestFloats extends TornadoTestBase {
         VectorFloat2 output = new VectorFloat2(size);
         VectorFloat2 seq = new VectorFloat2(size);
 
-        for (int x = 0; x < b.getLength(); x++) {
-            a.set(x, new Float2(x, x));
-            b.set(x, new Int2(1, 2));
+        Random r = new Random();
+        for (int i = 0; i < b.getLength(); i++) {
+            a.set(i, new Float2(r.nextFloat(), r.nextFloat()));
+            b.set(i, new Int2(r.nextInt(), r.nextInt()));
         }
 
         TaskGraph graph = new TaskGraph("s0") //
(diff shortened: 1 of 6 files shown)


mairooni (approved):
LGTM

laa:
Hi guys.
Would be good also provide documentation updates. 

Docs are scarce as of now and you need to guess features by inspecting the code. 
Not the best way for quick immersion into the project.

jjfumero:
Hi @laa , thanks for your feedback. Sure, the main reason is that TornadoVM is evolving really fast now and it is driven by the 5 EU projects we are working on. But I do agree with you that the documentation can be improved. 

Is there anything specific you would like to to cover first? Myself, I am preparing new tutorials on vector types, if that helps.

laa:
@jjfumero thank you :-) That is exactly what I wanted :-)

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/vectortypes/TestFloats.java
```diff
@@ -986,4 +1030,165 @@ public void testAllocationIssue() {
...
+    public void testVectorFloat16Int16() throws TornadoExecutionPlanException {
+        int size = 8192;
+        VectorFloat16 a = new VectorFloat16(size);
+        VectorInt16 b = new VectorInt16(size);
+        VectorFloat16 output = new VectorFloat16(size);
+        VectorFloat16 seq = new VectorFloat16(size);
+
+        for (int x = 0; x < b.getLength(); x++) {
+            a.set(x, new Float16(x, x, x, x, x, x, x, x, x, x, x, x, x, x, x, x));
+            b.set(x, new Int16(0, 1, 2, 3, 3, 2, 1, 0, 0, 1, 2, 3, 3, 2, 1, 0));
```
stratika, line 1172: I have a suggestion for the initialization of the tested vectors. I would suggest to use different values for each field of the vectors, in order to ensure that indexing is correct and we do not have any accidental matches. What do you think?

jjfumero: For this particular case it will have the same effect, because other tests already check with different values, but we can add random values instead.

jjfumero:
In general, we should work to improve the APIs for vectors. There is no other way to initialize a vector 16 with random numbers than this at the moment:

```java
arrayOfFloat16.set(i, new Float16(r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat(), //
                    r.nextFloat()));
```

The original design is given by the OpenCL/SPIR-V interfaces, but for Java we need to improve it. That's for a separate PR.

stratika (approved):
Thanks, I tested also on Apple M1. LGTM.

PR #426, opened 2024-05-16
Fix logger for all backends

jjfumero:
#### Description

Fix logger initialization after the Tornado class refactoring. 

#### Problem description

The issue was a missing initialisation for the factory logging methods. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
tornado --debug -ea  --jvm "-Xmx6g -Dtornado.recover.bailout=False -Dtornado.unittests.verbose=True "  -m  tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  --params "uk.ac.manchester.tornado.unittests.foundation.TestIntegers"
```


mairooni (approved):

jjfumero:
I have a better solution for the Logging. I will provide a new patch in a bit

jjfumero:
Let's have a new look. The reason behind this new design is that we are forced to declare the class and fine tune from which class we are enabling the logger instead of having the "Anonymous" option on all the time.

mairooni (approved):
LGTM

mikepapadim (approved):

PR #441, opened 2024-06-03
Fix failing unit-tests for PTX in Jenkins

stratika:
#### Description

This PR fixes some failing unit-tests for `PTX`.

#### Problem description

Here is a description of the reason of failing for every fixed test:
- `testTornadoMathCosPIDouble` and `testTornadoMathSinPIDouble`: `cos` and `sin` instructions support `f32` precision. So, we had to apply casting from `f64` to `f32` for those operators. For reference, see [here](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#floating-point-instructions).
- `testCopyInWithDevice`: This is a test that asserts whether there is a difference in the copy in data transfer timers, if the taskgraph is executed with the `withDevice` function, or not. This was due to a problem that was observed when the `withDevice` function was used, the behavior of an array that was defined with DataTransfer mode `FIRST_EXECUTION` was as if it was defined with `EVERY_EXECUTION`. Thus, we assess if the aggregated time is significantly higher. I increased the data size and number of iterations to make the case easier to track, and increased the delta for both metrics to be within a range of 25%.

#### Backend/s tested

- [ ] OpenCL
- [x] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=ptx
make tests
```

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestIO.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestIO.java
@@ -163,14 +163,14 @@ public class TestIO extends TornadoTestBase {
      * {@link uk.ac.manchester.tornado.api.TornadoExecutionPlan} API to execute on specific device.
      *
      * <p>
-     * The following test reproduces a bug, that when the user sets an array as DataTransferMode.FIRST_EXECUTION,
-     * when using the withDevice API the runtime overrides this setting and copies all arrays on every execution.
+     * The following test reproduces a bug, that when the user sets an array as DataTransferMode.FIRST_EXECUTION
+     * via the withDevice method, the runtime overrides this setting and copies all arrays on every execution.
      * </p>
      */
     @Test
     public void testCopyInWithDevice() {
-        final int N = 8096;
-        final int ITERATIONS = 20;
+        final int N = 16384;
+        final int ITERATIONS = 40;
 
         FloatArray arrayA = createAndInitializeArray(N);
         FloatArray arrayB = createAndInitializeArray(N);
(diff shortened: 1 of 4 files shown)


jjfumero:
It is not clear to me why we need to narrow the type from FP64 to FP32 if the operation selected is `double`. The PTX instruction set supports FP64. Can you expand on this?

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestIO.java
```diff
@@ -163,14 +163,14 @@ public void testLockObjectsInMemory() {
      * {@link uk.ac.manchester.tornado.api.TornadoExecutionPlan} API to execute on specific device.
      *
      * <p>
-     * The following test reproduces a bug, that when the user sets an array as DataTransferMode.FIRST_EXECUTION,
-     * when using the withDevice API the runtime overrides this setting and copies all arrays on every execution.
+     * The following test reproduces a bug, that when the user sets an array as DataTransferMode.FIRST_EXECUTION
+     * via the withDevice method, the runtime overrides this setting and copies all arrays on every execution.
```
jjfumero, line 167: Let's open an issue to solve this.

stratika:
> It is not clear to me why we need to narrow the type from FP64 to FP32 if the operation selected is `double`. The PTX instruction set supports FP64. Can you expand on this?

Yes of course, the operations `cos` and `sin` in PTX support only FP32. 
Screenshot 2024-06-04 at 10 14 42

This is something that we have already in develop when we run `cos` and `sin` operations with `double` types. However, the `cospi` and `sinpi` operations were escaping and producing compilation problems regarding mismatched types of the operands.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestIO.java
stratika: This is just a clarification of the comment that described the unit-test.

jjfumero:
Thanks @stratika  for the clarification. I saw how the NVIDIA compile handles when using OpenCL. It seems they use an intermediate table with pre-compute values called `__cudart_sin_cos_coeffs` . Let's continue with the narrow option and investigate this type of optimizations in the future.

jjfumero (approved):

jjfumero:
As a side note, can we document somewhere that we narrow these operations in the context of NVIDIA PTX?

stratika:
> As a side note, can we document somewhere that we narrow these operations in the context of NVIDIA PTX?

Do we have such documentation for the other narrowed operations? Any suggestions? Adding a Javadoc?

jjfumero:
I think the best way is to add Javadoc in the the TornadoMath API for those operations.

mikepapadim (approved):

jjfumero (approved):

PR #458, opened 2024-06-20
[fix] Profiler CopyIns fixed - READ_ONLY data fixed

jjfumero:
#### Description

This patch fixes the copy in events to be used in the profiler when the profiler option is on. Consecuently, it fixes a bug triggering more copies that needed when the device selected was not the default one. 

#### Problem description

The profiler used to print COPY_IN for read_only data if the execution plan was executed multiple times. The problem was that the internal copy functions were not returning the list of events properly. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
make BACKEND=spirv,ptx,opencl
make tests
```


mairooni (approved):
LGTM

jjfumero:
Updated with the latest develop

PR #460, opened 2024-06-24
Refactor tests and update to use the API try-with-resources

jjfumero:
#### Description

Refactor tests and update to use the API try-with-resources

#### Problem description

n/ a.

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

For each backend:

```bash
make tests
```


mikepapadim (approved):
LGTM, one thing that is not related with the PR directly.
We should update the Tornado Insight plugin to promote the use of try with resources.

PR #494, opened 2024-07-09
Add test to trigger unsupported input when user try to offload directly `MemorySegments` 

mikepapadim:
#### Description

Following the discussion in #491, currently there is no proper error and exception handling when the user tries to run TornadoVM with a `MemorySegment` directly without using any of the `TornadoNativeTypes`.

This PR adds a test to trigger this behaviour, so we can discuss at which level we should throw the appropriate exception to prevent end-users to try to use `MemorySegments` directly.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [x] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
tornado-test -V --debug uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType 
```

### DIFF
--- a/tornado-assembly/src/bin/tornado-test
+++ b/tornado-assembly/src/bin/tornado-test
@@ -123,6 +123,7 @@ __TEST_THE_WORLD__ = [
     TestEntry("uk.ac.manchester.tornado.unittests.tasks.TestMultipleTasksMultipleDevices"),
     TestEntry("uk.ac.manchester.tornado.unittests.vm.concurrency.TestConcurrentBackends"),
     TestEntry("uk.ac.manchester.tornado.unittests.api.TestDevices"),
+    TestEntry("uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType"),
     TestEntry("uk.ac.manchester.tornado.unittests.runtime.TestRuntimeAPI"),
     TestEntry("uk.ac.manchester.tornado.unittests.tensors.TestTensorTypes"),
     TestEntry("uk.ac.manchester.tornado.unittests.tensors.TestTensorAPIWithOnnx"),
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java
@@ -23,38 +23,9 @@
  */
 package uk.ac.manchester.tornado.runtime.tasks;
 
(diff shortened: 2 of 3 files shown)


andrii0lomakin:
@mikepapadim I do not use MemorySegments directly. I used them over the Float/Native arrays.

stratika:
thanks, I am testing it on `Linux` and `macOS` and got this:

```bash
tornado-test -V uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType

WARNING: Code Bailout to Java sequential. Use --debug to see the reason
Test: class uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType
	Running test: testMemorySegmentAsInput   ................  [FAILED] 
		\_[REASON] Unable to build sketch for method: getMemorySegment(Cannot invoke "jdk.vm.ci.meta.ResolvedJavaType.getAnnotation(java.lang.Class)" because the return value of "org.graalvm.compiler.core.common.type.ObjectStamp.type()" is null)
Test ran: 1, Failed: 1, Unsupported: 0
```

mikepapadim:
> thanks, I am testing it on `Linux` and `macOS` and got this:
> 
> ```shell
> tornado-test -V uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType
> WARNING: Using incubator modules: jdk.incubator.vector
> 
> WARNING: Code Bailout to Java sequential. Use --debug to see the reason
> Test: class uk.ac.manchester.tornado.unittests.api.TestMemorySegmentsAsType
> 	Running test: testMemorySegmentAsInput   ................  [FAILED] 
> 		\_[REASON] Unable to build sketch for method: getMemorySegment(Cannot invoke "jdk.vm.ci.meta.ResolvedJavaType.getAnnotation(java.lang.Class)" because the return value of "org.graalvm.compiler.core.common.type.ObjectStamp.type()" is null)
> Test ran: 1, Failed: 1, Unsupported: 0
> ```

Thats expected. Lets wait for @jjfumero to discuss at which level we should throw an appropriate exception if someone tries to use directly a `MemorySegment` and not the wrapped types.

jjfumero:
We can't use a segment directly due to the internal object header, which is custom for the GPU/FPGA. Also, with JDK22, we can't invoke it directly. So the way to go is to throw an exception during the Task-Graph built, since we get all input/output objects to create the internal GPU buffers.

mikepapadim:
This is fixed now, it throws a `TornadoRuntimeException` when user passes a MemorySegment as Task parameter

stratika (approved):
LGTM

PR #502, opened 2024-07-19
Too strict barrier

andrii0lomakin:
I believe we need localBarrier here because we do not modify global memory.


jjfumero:
It makes sense. We do not need a global barrier for this operation.

jjfumero (approved):

stratika (approved):

PR #509, opened 2024-07-25
Add `quickPass` option to skip heavy tests in tornado-test script

mikepapadim:
#### Description

Testing script runs the complete unit-test suite when is invoked. The test suite contains some test that they try to stress the device memory or rely on stable internet connection (high timeout). Thus, testing the whole suite during local development can cause  overhead.

This PR add an option in the test script to do a quick pass, by skipping these tests that cause they delay. This option by default is disabled and they user needs to explictly add it.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [ ] No

```bash
make jdk21 
tornado-test --ea --verbose --quickPass
```

### DIFF
--- a/tornado-assembly/src/bin/tornado-test
+++ b/tornado-assembly/src/bin/tornado-test
@@ -217,6 +217,13 @@ __TORNADO_TESTS_WHITE_LIST__ = [
 
 ]
 
+## List of tests to be excluded when running with te quick pass argument as they cause delays
+__TORNADO_HEAVY_TESTS__ = [
+    "uk.ac.manchester.tornado.unittests.memory.TestStressDeviceMemory",
+    "uk.ac.manchester.tornado.unittests.profiler.TestProfiler",
+    "uk.ac.manchester.tornado.unittests.tensors.TestTensorAPIWithOnnx",
+]
+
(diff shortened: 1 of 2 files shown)


jjfumero:
Add also a new rule in the `Makefile`: `make quick-tests` something like that..

tornado-assembly/src/bin/tornado-test
```diff
@@ -217,6 +217,13 @@ __TORNADO_TESTS_WHITE_LIST__ = [
 
 ]
 
+## List of tests to be excluded when running with te quick pass argument as they cause delays
```
stratika, line 220:
```suggestion
## List of tests to be excluded when running with the quick pass argument as they cause delays
```

stratika (commented):
LGTM, just a typo fix and please add the rule as suggested by @jjfumero.

mikepapadim:
Done
The option is make fast-tests

stratika (approved):

jjfumero (approved):

Issue #506, opened 2024-07-25
Double addition of inputs in PrintfNode prevents inlining and as result debuggin


andrii0lomakin:
During the addition of PrintfNode, all its input usages are also registered in the constructor because of the calling of `NodeList#set` method.
After that, it obviously added to the graph of operations by calling ` b.add(b.append(print mode));.`

That causes a second attempt to add printf inputs to the graph. This causes side effect - argument is used as a parameter of printf registers printf extra usage twice, so as a result during inlining, inputs of two printf nodes, instead of one, will be tried to be replaced by the new inlined node that causes an exception during compilation and as result inability to use printf for debugging.

Let us consider the example:

```
        var currentInputOffset = inputOffset + context.globalIdy * tensorLength + context.globalIdx;
        var localSnippet = context.allocateFloatLocalArray(64);

        float value = inputTensor.get(currentInputOffset);

        var localOffset = context.localGroupSizeX * context.localIdy;
        localSnippet[localOffset + context.localIdx] = value * value;

        if (context.localIdx == 0) {
            Debug.printf("%f\n", value);
        }
```

In the graph printf argument will be `FloatArray#get` method, which will be extracted in the plugin and then:
During construction, FloatArray#get will be added as input and then PrintfNode will be added as extra usage of FloatArray#get
![image](https://github.com/user-attachments/assets/6cd0aafe-c97a-4b3b-8200-1ad1f1ce1cdc)
But then after the call of ` b.add(b.append(print mode));.` , one more extra usage of `FloatArray#get` is added
![image](https://github.com/user-attachments/assets/3e44407a-234e-42b0-a010-70191280d0ac)
That causes an exception during `FloatArray#get` inlining because second attempt to replace the input of `second` printf node fails (because those nodes are the same instances):

```
Unable to build sketch for method: squareSumKernel(not found in inputs, usage: 64|printf)
uk.ac.manchester.tornado.api.exceptions.TornadoBailoutRuntimeException: Unable to build sketch for method: squareSumKernel(not found in inputs, usage: 64|printf)
	at tornado.runtime@1.0.6/uk.ac.manchester.tornado.runtime.sketcher.TornadoSketcher.buildSketch(TornadoSketcher.java:190)
	at tornado.runtime@1.0.6/uk.ac.manchester.tornado.runtime.sketcher.TornadoSketcher$TornadoSketcherCallable.call(TornadoSketcher.java:262)
[...]

PR #513, opened 2024-07-27
Issue #506 has been fixed.

andrii0lomakin:
Issue #506 has been fixed.
Not used nodes, not related to any API or code, has been removed. 
GraalVM JIT has been added as provided dependency to improve IDE support.

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [x] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

`tornado-test -V uk.ac.manchester.tornado.unittests.TestHello`


jjfumero:
Thanks for the contribution. It looks much cleaner. 
This PR applies for OpenCL backend. I will remove the `PTX` mark from this PR. 

Before we merge, @mikepapadim , can you merge this into a new branch that includes `JDK22`? I want to see how these plugins will work as they are. 

It works on Linux. I will check on OSx.

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/plugins/OCLGraphBuilderPlugins.java
```diff
@@ -430,39 +366,6 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
     }
 
     private static void registerTornadoVMIntrinsicsPlugins(InvocationPlugins plugins) {
-        final InvocationPlugin tprintfPlugin2 = new InvocationPlugin("tprintf", String.class, Object[].class) {
```
jjfumero, line 433: I see the `tprinf` intrinsic is removed, which makes sense. We can debug directly using the `Debug.printf` method.

jjfumero (approved):
LGTM

jjfumero:
For completeness: printf function does not work (does not print anything) on Apple M1-M3 OpenCL implementations, despite that it is OpenCL 1.2. This is a driver issue, and nothing related to this PR.

andrii0lomakin:
@jjfumero For completeness, I also observed issues with PTX and vprintf, but because I have not made a deep investigation, I will provide a separate PR + Issue once we start to work with it closely. Right now, we are concentrating on OpenCL kernels. Because the OpenCL backend is the most widely provided, and then we will work on performance optimizations for specific backends.

mikepapadim (approved):
LGTM, thanks

I also tested with the new port JDK22 and it works.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/TestHello.java
```diff
@@ -94,6 +103,26 @@ public void testHello() throws TornadoExecutionPlanException {
         }
     }
 
+    @Test
+    public void testPrintIntArray() throws TornadoExecutionPlanException {
+        assertNotBackend(TornadoVMBackendType.SPIRV);
```
stratika, line 108: Please add also the `assertNotBackend` check for the PTX type.

jjfumero: Yes, good catch

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/TestHello.java
```diff
@@ -50,6 +50,15 @@ private static void printHello(final int n) {
...
 
+    private static void printIntArray(IntArray a) {
+        int firstValue = a.get(0);
+        int secondValue = a.get(1);
+
+        if (a.getSize() > 1) {
+            Debug.printf("First value %d, second value %d\n", firstValue, secondValue);
+        }
+    }
+
```
stratika, line 61: I would suggest to complement this test, or add a second one that prints the value of each element in the input array, and uses the @Parallel annotation, so that to test the indices passed from the `GlobalThreadIdNode`.

e.g., 
```bash
private static void printIntArray(IntArray a) {
        for (@Parallel int i = 0; i < a.getSize(); i++) {
            Debug.printf("value a[%d] = %d\n", i, a.get(i));
        }
    }
```

Ideally, we can extend the test to assess all 3 dimensions.

stratika (commented):
thanks @andrii0lomakin, it looks really cool to have this feature supported. The code looks good to me. Just one comment regarding the test. I would suggest to extend it to assess also that the printing works with the parallel loop (I already tested it in my laptop and it works).

I will test it in a bit in Windows for sanity check and will let you know.

stratika:
It works on my laptop with Windows 11.

andrii0lomakin:
Hi @stratika 
Thank you for your help.
I have added new tests.

jjfumero (approved):
LGTM

stratika (approved):
LGTM, thanks

jjfumero:
Thanks all, I will merge this PR

PR #531, opened 2024-08-13
[deps] Level Zero JNI v0.1.4 updated

jjfumero:
#### Description

Update the LevelZero JNI dependency with the new version 0.1.4. This new version brings passing constant parameters to Level Zero Kernels.

#### Problem description

n/ a.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

To use the new version:

```bash
$ rm -Rf levelzero-jni/ level-zero/
$ source setvars.sh
$ make BACKEND=spirv
```

### DIFF
--- a/tornado-drivers/spirv/pom.xml
+++ b/tornado-drivers/spirv/pom.xml
@@ -23,7 +23,7 @@
         <dependency>
             <groupId>beehive-lab</groupId>
             <artifactId>beehive-levelzero-jni</artifactId>
-            <version>0.1.3</version>
+            <version>0.1.4</version>
         </dependency>
 
         <dependency>


mairooni (approved):

stratika (commented):
I think it would be nice, if we can also change the `compile` [script](https://github.com/beehive-lab/TornadoVM/blob/68abbcd386e82a51714a1104e837a4a89fe508cb/bin/compile#L160) to pull the branch that follows the same version. Because as it is right now, we can not revert to a previous version, and we are forcing using the master. 

However, this change implies that in the future we should update both the `compile` script and the `pom` file.

In any case it works.

jjfumero:
I do agree with this. Let's do that in a separate PR

stratika (approved):

Issue #534, opened 2024-08-14
[bug] TestReductionsAutomatic#testIrregularSize03 is failing on all backends


andrii0lomakin:
### Describe the bug 
TestReductionsAutomatic#testIrregularSize03 is failing on all backends

### How To Reproduce

` tornado-test -V uk.ac.manchester.tornado.unittests.reductions.TestReductionsAutomatic`

Result is:
```
Test: class uk.ac.manchester.tornado.unittests.reductions.TestReductionsAutomatic
        Running test: testIrregularSize01        ................  [PASS]
        Running test: testIrregularSize02        ................  [PASS]
        Running test: testIrregularSize03        ................  [FAILED]
```

### Expected behavior 

Test is passed

### Computing system setup (please complete the following information):
 - OS: Windows, Ubuntu 22.04
 - OpenCL and Driver versions -  OpenCL version 3.0 , driver version - 23.43.27642.52
 - If applicable, PTX and CUDA Driver versions - driver version 560.81, cuda version 12.6
 - If applicable, Level Zero & SPIR-V Versions - 1.9.3
 - TornadoVM commit id - `35e7e550c`

jjfumero:
Yes, I think this was introduced in the latest integration with Graal. We will take a look,

jjfumero:
Test now passing. and the fix was merged in #540 

```bash
tornado-test -V uk.ac.manchester.tornado.unittests.reductions.TestReductionsAutomatic

WARNING: Using incubator modules: jdk.incubator.vector

Test: class uk.ac.manchester.tornado.unittests.reductions.TestReductionsAutomatic
        Running test: testIrregularSize01        ................  [PASS] 
        Running test: testIrregularSize02        ................  [PASS] 
        Running test: testIrregularSize03        ................  [PASS] 
        Running test: testIrregularSize04        ................  [FP64 UNSUPPORTED FOR CURRENT DEVICE] 
Test ran: 4, Failed: 0, Unsupported: 1

```

PR #540, opened 2024-08-26
Fix issue with the TestReductionsAutomatic unittest 

mairooni:
#### Description
This PR provides a fix for the `TestReductionsAutomatic#testIrregularSize03` unittest, which has been failing for all backends. The PR is associated with the issue #534.  

#### Problem description
When running the `TestReductionsAutomatic` tests, the test `testIrregularSize03` has been failing consistently for all backends. Upon further investigation, it was discovered that the issue was not with the test itself, since running it individually did not produce any errors. Additionally, removing the test completely from the test file resulted in different tests failing. The actual issue was pinpointed in the `clean()` method that was invoked between tests instead. Specifically, the `clean()` method was throwing a `ConcurrentModificationException`  when accessing the set containing the plan ids. As a workaround, the function now creates a copy of this set and, using this copy, it gets the ids to perform the cleanup. 
 
#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run `tornado-test -V uk.ac.manchester.tornado.unittests.reductions.TestReductionsAutomatic`


jjfumero (commented):
Thanks @mairooni , Shouldn't we use a `ConcurrentHashSet` instead? Is this clean up function thread safe?

mairooni:
> Thanks @mairooni , Shouldn't we use a `ConcurrentHashSet` instead? Is this clean up function thread safe?

Thanks @jjfumero, indeed `HashSet` is not thread safe but AFAIK `ConcurrentHashSet` does not exist. 
Instead, I can use a `ConcurrentHashMap` in the following way: 

        
```
Set<Long> ids = device.getDeviceContext().getRegisteredPlanIds(); 
Set<Long> concurrentIds = Collections.newSetFromMap(new ConcurrentHashMap<>());  
concurrentIds.addAll(ids); 
// use concurrentIds for the modifications
```

        
I ran the unittest with this change and it passes. Does this look good to you or do you have something else in mind?

stratika (approved):
LGTM, I tested it with all backends.

PR #548, opened 2024-09-02
Minor improvements

jjfumero:
#### Description

This PR provides minor improvements:
- It improves MxM example to dump all raw timers to be processed by external tools (e.g., R) 
- It simplifies detection of parallel kernels per backend.

#### Problem description

n/ a.

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
make 
make tests

tornado --jvm="-Ds0.t0.device=0:0" -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixMultiplication2D 512 false
```


stratika (approved):
LGTM, one comment is that we are modifying an example from the compute package to showcase this feature of dumping stats related to execution time for further analysis. We could probably decouple this functionality into a separate package that will be used to show how to do the analysis and potential integration with other tools.

This is a thought.

mairooni (approved):

jjfumero:
> LGTM, one comment is that we are modifying an example from the compute package to showcase this feature of dumping stats related to execution time for further analysis. We could probably decouple this functionality into a separate package that will be used to show how to do the analysis and potential integration with other tools.
> 
> This is a thought.

Yes, I agree. We can extend the benchmark module. I suggest moving this one forward to create the requested demos for EU AERO Project.

PR #557, opened 2024-09-17
[fix] Shared Execution Context fixed when having Multi-Threaded Java applicatons

jjfumero:
#### Description

This PR fixes multi-threaded execution plans. The issue (see below) was related to the unsafe shared of the code cache. The code cache was applied per device context, which is shared across all execution plans that run on the same device. Instead, what we need is to create a code-cache per execution plan ID. 

#### Problem description

The multi-threaded tests stop working due to unsafe sharing of the code-cache within the TornadoVM device context. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

For each backend:

```bash
tornado-test --jvm="-Dtornado.device.memory=2GB" -V --fast uk.ac.manchester.tornado.unittests.multithreaded.TestMultiThreadedExecutionPlans
```

Before we merge, apart from all unit-test, we should test Ray-Tracer and Kfusion.

### DIFF
--- a/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/PTXDeviceContext.java
+++ b/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/PTXDeviceContext.java
@@ -29,12 +29,7 @@ import static uk.ac.manchester.tornado.drivers.ptx.graal.PTXCodeUtil.buildKernel
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
 import uk.ac.manchester.tornado.api.TornadoDeviceContext;
@@ -62,7 +57,7 @@ public class PTXDeviceContext implements TornadoDeviceContext {
 
     private final PTXDevice device;
     private final PTXMemoryManager memoryManager;
-    private final PTXCodeCache codeCache;
+    private final Map<Long, PTXCodeCache> codeCache;
     private final PTXScheduler scheduler;
     private final TornadoBufferProvider bufferProvider;
     private final PowerMetric powerMetric;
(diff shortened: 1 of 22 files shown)


stratika:
An overall comment/question. Would it make sense to add the `executionPlanID` as part of the `SchedulableTask` class? Based on the name, this class should have some info fed from the `TornadoExecutionPlan`?

For the record, I tested on my Linux OS machine with `OpenCL`, `PTX`, `SPIRV` (both levelzero and opencl). I also tested the TornadoVM-Ray-Tracer with `OpenCL` and `SPIRV` (both levelzero and opencl), and the K-FUSION with `OpenCL` and `PTX`. Everything worked.

jjfumero:
I was thinking about this. Maybe yes, but then it will be some inconsistencies with the rest of the runtime (e.g., for copies, in which we pass the `executionPlanId` explicitly).

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/PTXDeviceContext.java
```diff
@@ -29,12 +29,7 @@
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
mairooni, line 32: Remove * import

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoXPUDevice.java
```diff
@@ -65,32 +65,38 @@ public interface TornadoXPUDevice extends TornadoDevice {
     /**
      * It installs the Tornado code for the specified schedulable task.
      *
+     * @param executionPlanId
+     *   ID number for the execution plan that the tasks belongs to.
```
mairooni, line 69: tasks -> task

stratika (approved):
LGTM, thanks @jjfumero

mairooni (approved):
Apart from some minor comments it looks good to me. All tests (+ RayTracer and KFusion) passed on my local Linux machine.

jjfumero:
Thanks. I applied the fixes. I am merging this PR.

PR #561, opened 2024-09-19
[fix] Build for OSx 14.5 and CLANG 16

jjfumero:
#### Description

This patch removes the flag `--export-dynamic` for the OSx build. The problem is that when the CLANG updated on OSx 14.5 along with the XCode, this flag is no longer recognised. 

#### Problem description

See above.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [X] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

The problem was related to the build.
```bash
make BACKEND=opencl
make tests
```


stratika (approved):
LGTM, thanks @jjfumero

mairooni (approved):

PR #562, opened 2024-09-19
[fix] Warmup action from the execution plan with updates IDs

jjfumero:
#### Description

Warmup action from the execution plan with updates IDs. The solution is to pass the `executionPackage` from the execution plan down to the TornadoVM Task Graph, and then update the Ids accordingly.  

#### Problem description

The problem was that the executionPlanId was not updated for the warmup phase (blocking call). 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
make BACKEND=spirv
tornado --threadInfo --jvm "-Xmx6g -Dtornado.recover.bailout=False -Dtornado.unittests.verbose=True -Dtornado.unittests.device=0:1"  -m  tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner  --params "uk.ac.manchester.tornado.unittests.arrays.TestArrays"

```

Then run all tests

```bash
make tests
```


stratika (approved):
LGTM. Thanks @jjfumero, it makes sense that the `warmup()` method should follow the signature that the `execute()` method has, since it pre-compiles the code for execution.

I confirm that I tested the branch on different OSs and across various devices. I created this summary:
Screenshot 2024-09-20 at 10 46 36

mikepapadim (approved):

PR #571, opened 2024-10-01
[OpenCL] Fix issues with brackets for if-conditions in for-loops

mairooni:
#### Description

This PR fixes the closing brackets for OpenCL kernels that contain if-conditions either before or in for-loops. 

#### Problem description

It was observed that in some cases where there was an if-condition in a kernel that contained for-loop(s)  (e.g. before a for-loop, in an inner for-loop etc.) , either some closing brackets were missing or unnecessary additional closing brackets were included. With this PR, the code generation examines these cases and generates the correct brackets.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run `tornado-test -V -pk uk.ac.manchester.tornado.unittests.branching.TestLoopConditions`
All unittests have been tested along with RayTracer and KFusion.

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/branching/TestLoopConditions.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/branching/TestLoopConditions.java
@@ -1,0 +1,544 @@
+/*
+ * Copyright (c) 2024, APT Group, Department of Computer Science,
+ * The University of Manchester.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.manchester.tornado.unittests.branching;
+
+import org.junit.Test;
+import uk.ac.manchester.tornado.api.ImmutableTaskGraph;
+import uk.ac.manchester.tornado.api.TaskGraph;
+import uk.ac.manchester.tornado.api.TornadoExecutionPlan;
+import uk.ac.manchester.tornado.api.annotations.Parallel;
+import uk.ac.manchester.tornado.api.enums.DataTransferMode;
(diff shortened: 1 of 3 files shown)


stratika:
Just a note, I tested the branch in `macOSx` and `Linux Ubuntu 23.10` and the fix is working.

stratika (approved):

mairooni:
> Just a note, I tested the branch in `macOSx` and `Linux Ubuntu 23.10` and the fix is working.

Thanks for testing it Thano!

jjfumero:
That's great. Based on my experience when we fix the OpenCL visitor is that we always need to check with KFusion.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/branching/TestLoopConditions.java
```diff
@@ -0,0 +1,544 @@
...
+ * <code>
+ * tornado-test -V uk.ac.manchester.tornado.unittests.branching.TestLoopConditions
+ * </code>
+ */
+public class TestLoopConditions extends TornadoTestBase {
+
+    public static void conditionBeforeSingleLoopReturn(IntArray a, IntArray b, IntArray c) {
+        if (a.get(0) > b.get(0)) {
+            return;
+        }
```
jjfumero, line 48: What is the expected behaviour of this type of code when running in parallel on a GPU?  We can't stop the GPU kernel.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/branching/TestLoopConditions.java
```diff
@@ -0,0 +1,544 @@
...
+        for (@Parallel int i = 0; i < c.getSize(); i++) {
+            c.set(i, a.get(i) + b.get(i));
+        }
+    }
+
+    public static void conditionInSingleLoopReturn(IntArray a, IntArray b, IntArray c) {
+        for (@Parallel int i = 0; i < c.getSize(); i++) {
+            if (a.get(0) > b.get(0)) {
+                return;
+            }
```
jjfumero, line 67: Same here, what is the expected behaviour when running on a GPU?

stratika:
> That's great. Based on my experience when we fix the OpenCL visitor is that we always need to check with KFusion.

Good note! Just tested `kfusion` with OpenCL, PTX on NVIDIA RTX A2000 GPU.

I also tested `TornadoVM-Ray-Tracer` for OpenCL.

jjfumero (approved):
After the sync, let's clarify in the header of the test and before each kernel why this test is valid and under which conditions will be still valid for TornadoVM.

jjfumero (approved):

PR #601, opened 2024-12-10
[Feature] Select graph from a multi-graph execution plan to launch. 

jjfumero:
#### Description

This PR extends the `TornadoExecutionPlan` API with a new method to select a graph to launch from a multi-graph execution plan configuration. This type of code is useful when we need to share context (e.g., command queues, cuda streams, code cache, etc) for the whole execution plan, but we only need to execute a task-graph at the time. 

The two methods added are as follows:

```java
try (TornadoExecutionPlan executionPlan = new TornadoExecutionPlan(tg1.snapshot(), tg2.snapshot())) {

      // Select graph 1 (tg2) to execute
      // Once selected, every time we call the execute method,
      // TornadoVM will launch the passed task-graph.
      executionPlan.withGraph(1).execute();
  
      // Select the graph 0 (tg1) to execute
      executionPlan.withGraph(0).execute();

      // Select all graphs (tg1 and tg2) to execute.
      // Since we selected individual task-graphs, we should be
      // able to reverse this action and invoke all task-graph
      // again. This is achieved with the `withAllGraphs` from the
      // execution plan.
       executionPlan.withAllGraphs().execute();
 }
```

This allows the following pattern in TornadoVM:

```java
for (numIterations):
   executionPlan.withGraph(1).execute(); 
}

executionPlan.withGraph(2).execute();
...
```

#### Problem description

This path expands the capabilities of TornadoVM towards sharing state across different task-graph within a single execution plan. 

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
tornado-test --printKernel --threadInfo -V uk.ac.manchester.tornado.unittests.executor.TestExecutor#test06
```

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
@@ -17,9 +17,9 @@
  */
 package uk.ac.manchester.tornado.api;
 
-import java.util.Objects;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicLong;
 
 import uk.ac.manchester.tornado.api.common.TornadoDevice;
@@ -32,6 +32,7 @@ import uk.ac.manchester.tornado.api.plan.types.OffMemoryLimit;
 import uk.ac.manchester.tornado.api.plan.types.OffPrintKernel;
 import uk.ac.manchester.tornado.api.plan.types.OffProfiler;
 import uk.ac.manchester.tornado.api.plan.types.OffThreadInfo;
+import uk.ac.manchester.tornado.api.plan.types.WithAllGraphs;
 import uk.ac.manchester.tornado.api.plan.types.WithBatch;
 import uk.ac.manchester.tornado.api.plan.types.WithClearProfiles;
 import uk.ac.manchester.tornado.api.plan.types.WithCompilerFlags;
@@ -40,6 +41,7 @@ import uk.ac.manchester.tornado.api.plan.types.WithDefaultScheduler;
 import uk.ac.manchester.tornado.api.plan.types.WithDevice;
 import uk.ac.manchester.tornado.api.plan.types.WithDynamicReconfiguration;
 import uk.ac.manchester.tornado.api.plan.types.WithFreeDeviceMemory;
+import uk.ac.manchester.tornado.api.plan.types.WithGraph;
 import uk.ac.manchester.tornado.api.plan.types.WithGridScheduler;
 import uk.ac.manchester.tornado.api.plan.types.WithMemoryLimit;
 import uk.ac.manchester.tornado.api.plan.types.WithPrintKernel;
@@ -163,6 +165,34 @@ public sealed class TornadoExecutionPlan implements AutoCloseable permits Execut
         TornadoExecutionResult executionResult = new TornadoExecutionResult(profilerResult);
         planResults.add(executionResult);
         return executionResult;
+    }
+
+    /**
+     * Select a graph from the {@link TornadoExecutionPlan} to execute.
+     * This method allows developers to select a specific graph from the
+     * execution plan to launch. Developers can choose which graph from
+     * the input list to use (passed in the constructor).
+     *
+     * 
+     * @since 1.0.9
+     * @param graphIndex
+     * @return {@link TornadoExecutionPlan}
+     */
+    public TornadoExecutionPlan withGraph(int graphIndex) {
+        tornadoExecutor.selectGraph(graphIndex);
+        return new WithGraph(this, graphIndex);
+    }
+
+    /**
+     * Select all graphs from the {@link TornadoExecutionPlan}. This method
+     * has an effect if the {@link #withGraph(int)} method was invoked.
+     *
+     * @since 1.0.9
+     * @return {@link TornadoExecutionPlan}
+     */
+    public TornadoExecutionPlan withAllGraphs() {
+        tornadoExecutor.selectAll();
+        return new WithAllGraphs(this);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


mikepapadim:
is the original `.execute()` still works or we need to replace all instnaces with `  executionPlan.withAllGraphs().execute();`?

mikepapadim (approved):
LGTM, thanks @jjfumero

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/executor/TestExecutor.java
```diff
@@ -303,5 +303,56 @@ public void test05() throws TornadoExecutionPlanException {
...
+     */
+    @Test
+    public void test06() throws TornadoExecutionPlanException {
+        int numElements = 16;
+        IntArray a = new IntArray(numElements);
+        IntArray b = new IntArray(numElements);
+        IntArray c = new IntArray(numElements);
+
+        a.init(1);
+        b.init(2);
```
stratika, line 321: can we please test the outcome when the elements of the arrays are initialised with different values?

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/executor/TestExecutor.java
```diff
@@ -303,5 +303,56 @@ public void test05() throws TornadoExecutionPlanException {
...
+            executionPlan.withGraph(1).execute();
+            for (int i = 0; i < c.getSize(); i++) {
+                assertEquals(a.get(i) + b.get(i), c.get(i));
+            }
+
+            // Select the graph 0 (tg1) to execute
+            executionPlan.withGraph(0).execute();
+            for (int i = 0; i < c.getSize(); i++) {
+                assertEquals(a.get(i) + b.get(i), c.get(i));
+            }
```
stratika, line 347: in this case, all elements in `c` should be 3, right? I think we should include a test that has different graphs, and potentially the second graph to use the outcome of the first graph.

Does this make sense? In my view this would be a good test to assess the consistency of data used across graphs.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/executor/TestExecutor.java
```diff
@@ -303,5 +303,56 @@ public void test05() throws TornadoExecutionPlanException {
...
+
+        TaskGraph tg1 = new TaskGraph("s0") //
+                .transferToDevice(DataTransferMode.FIRST_EXECUTION, a, b) //
+                .task("t0", TestHello::add, a, b, c) //
+                .transferToHost(DataTransferMode.EVERY_EXECUTION, c);
+
+        TaskGraph tg2 = new TaskGraph("s1") //
+                .transferToDevice(DataTransferMode.FIRST_EXECUTION, a, b) //
+                .task("t1", TestHello::add, a, b, c) //
+                .transferToHost(DataTransferMode.EVERY_EXECUTION, c);
```
stratika, line 331: is it possible to add a test in which the graphs have different compute method and they do not share the same variables?

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
```diff
@@ -165,6 +167,34 @@ public TornadoExecutionResult execute() {
...
+     * Select all graphs from the {@link TornadoExecutionPlan}. This method
+     * has an effect if the {@link #withGraph(int)} method was invoked.
+     *
+     * @since 1.0.9
+     * @return {@link TornadoExecutionPlan}
+     */
+    public TornadoExecutionPlan withAllGraphs() {
+        tornadoExecutor.selectAll();
+        return new WithAllGraphs(this);
+    }
```
stratika, line 196: is this the default way? as @mikepapadim mentioned, would the following still work to run all graphs?

```bash
TornadoExecutionPlan executionPlan = new TornadoExecutionPlan(tg1.snapshot(), tg2.snapshot());
executionPlan.execute();
```

jjfumero:
> is the original `.execute()` still works or we need to replace all instnaces with ` executionPlan.withAllGraphs().execute();`?

No need to use `withAllGraphs` to execute all graphs. This is the action by default. The method `withAllGraphs` is only used when a specific graph was selected a a run, and then the developer wants to select all of them in subsequent runs.

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
jjfumero: My answer below

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/executor/TestExecutor.java
jjfumero: It depends on how accessor type associated with the data. If it is read only for both graphs, then yes. Otherwise,a dependency is carried for the next graph. I added a unit-test to check this.

jjfumero:
Comment addressed.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/executor/TestExecutor.java
```diff
@@ -303,5 +304,104 @@ public void test05() throws TornadoExecutionPlanException {
...
+            // Since we selected individual task-graphs, we should be
+            // able to reverse this action and invoke all task-graph
+            // again. This is achieved with the `withAllGraphs` from the
+            // execution plan.
+            executionPlan.withAllGraphs().execute();
+        }
+    }
+
+    /**
+     * Test Multi-Graphs in an execution plan. Based on the {@link #test06()}, this test checke
```
stratika, line 360:
```suggestion
     * Test Multi-Graphs in an execution plan. Based on the {@link #test06()}, this test checks
```

stratika (approved):
thanks, LGTM

PR #611, opened 2025-01-16
Support Multiplication of Half Float Matrices

mairooni:
#### Description

This PR extends TornadoVM with the necessary functionality to support multiplication of matrices that consist of half float values. In this PR it is assumed that the results of the multiplication do not overflow,  so the output matrix consists also of half float values.

#### Problem description

The changes in this PR take place on the compilation and code generation fronts. Regarding compilation, the key change is regarding the `ValuePhiNode`, which accumulates the dot products of the multiplication to perform the addition. Since Java does not have a half float type by default (this type was introduced by TornadoVM), the half float values are treated as Java Objects by default. To avoid compilation errors due to that fact, a new `HalfFloatStamp` is introduced and used in the `ValuePhiNode` instead. This stamp is also now used in all half float related nodes, which before were represented as shorts. To this extend, new nodes for the operations between half float values are also introduced, since the `HalfFloatStamp` is an illegal stamp for the default add/sub/div/mult nodes of graal. Finally, a node to represent half float constants has been included for the same purposes.  These new operator nodes generate the code that their operators indicate. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
tornado-test -V uk.ac.manchester.tornado.unittests.compute.ComputeTests#testHalfFloatMatrixMultiplication
```

### DIFF
--- a/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/HalfFloatStamp.java
+++ b/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/HalfFloatStamp.java
@@ -1,3 +1,24 @@
+/*
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
+ * School of Engineering, The University of Manchester. All rights reserved.
+ * Copyright (c) 2009, 2017, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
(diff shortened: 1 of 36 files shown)


tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/HalfFloatStamp.java
```diff
@@ -0,0 +1,109 @@
...
+import jdk.vm.ci.meta.JavaKind;
+import jdk.vm.ci.meta.MemoryAccessProvider;
+import jdk.vm.ci.meta.MetaAccessProvider;
+import jdk.vm.ci.meta.ResolvedJavaType;
+import org.graalvm.compiler.core.common.LIRKind;
+import org.graalvm.compiler.core.common.spi.LIRKindTool;
+import org.graalvm.compiler.core.common.type.Stamp;
+import uk.ac.manchester.tornado.drivers.ptx.graal.lir.PTXKind;
+
+public class HalfFloatStamp extends Stamp {
```
jjfumero, line 34: I wonder if the `HalfFloatStamp` should be a common type for all backends. If so, let;s do that in a separate PR

jjfumero:
This is great! Tested with all backends and it works. 

For next PR:
- Enclose an FMA operation for `HalfFloat` too

jjfumero (approved):

stratika (approved):
LGTM, I tested it with all backends on Linux OS.

PR #622, opened 2025-02-10
[Feature] Device Buffer Mapping for different Buffers

jjfumero:
#### Description

This patch introduces a new feature called Mapping On Device Memory Regions. This feature is introduced with an API call from the Execution Plan:

```java
executionPlan.mapOnDeviceMemoryRegion (
    destArray,   // Dest array to map
    sourceArray,    // source tornadovm array  
    offset,      // offset within the source array to map
    fromGraphIndex,  // source is taken from graph marked here
    toGraphIndex     // dest is put into the graph marked here
);
```
This feature is only applicable to Execution Plans with at least one task-graph. 

```java
TornadoExecutionPlan executionPlan = new TornadoExecutionPlan(taskGraph1.snapshot(),  taskGraph2.snapshot());
```

This operation maps the `srcArray` to the `dstArray` from specified offset. The `srcArray` is mapped from `taskGraph` X to `destArray` in graph X. The TornadoVM runtime maps `destArray` to `srcArray` directly on-device, avoiding data transfers between host and device. This is a technique to implement double-buffer across task-graphs. 

[EDIT] This feature only works for `TornadoNativeArrays`, (Panama Segments). We may introduce this feature for Java object arrays in the future. 

#### Problem description

If the patch provides a fix for a bug, please describe what was the issue and how to reproduce the issue.

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
tornado-test  -pbc --threadInfo -V uk.ac.manchester.tornado.unittests.pointers.TestCopyDevicePointers
```

For the Level Zero API, note that custom offsets are not currently supported.  

```bash
make BACKEND=spirv
tornado-test --jvm="-Dtornado.spirv.runtimes=levelzero" -pbc --threadInfo -V uk.ac.manchester.tornado.unittests.pointers.TestCopyDevicePointers
```


jjfumero:
Note that there is an improvement still to be done. `DEALLOC` byte-code can be saved if `transferToDevice` objects are tagged as `UNDER_DEMAND`. Since this PR is already big, we will address this in a separate PR.

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
```diff
@@ -553,4 +552,27 @@ public TornadoExecutionResult getPlanResult(int index) {
         return planResults.get(index);
     }
 
+    /**
+     * This function maps the device memory region that corresponds to a TornadoVM array to another on-device memory region.
```
stratika, line 556: Shall we replace "TornadoVM array" with "TornadoVM object"?

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
```diff
@@ -553,4 +552,27 @@ public TornadoExecutionResult getPlanResult(int index) {
...
+     * This call instructs the TornadoVM runtime to not perform a data transfers between `device` -> `host` -> `device`. Instead,
```
stratika, line 557:
```suggestion
     * This call instructs the TornadoVM runtime to avoid transferring data between `device` -> `host` -> `device`. Instead,
```

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
```diff
@@ -553,4 +552,27 @@ public TornadoExecutionResult getPlanResult(int index) {
...
+     * it can update the corresponding device pointers.
+     *
+     * <p>
+     * The semantics is as follows: there is the source array, and the dest array. This call maps the dest array to the source
```
stratika, line 561:
```suggestion
     * The semantics are as follows: there is the source array, and the destination array. This call maps the destination array to the source
```

tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
```diff
@@ -553,4 +552,27 @@ public TornadoExecutionResult getPlanResult(int index) {
...
+     * array from a given offset. The source array is passed from the task-graph `fromTaskGraphIndex`, and the dest array is located
```
stratika, line 562:
```suggestion
     * array from a given offset. The source array is passed from the task-graph `fromGraphIndex`, and the dest array is located
```

jjfumero:
Comments addressed

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/mm/OCLArrayWrapper.java
```diff
@@ -346,4 +346,9 @@ public void setSizeSubRegion(long batchSize) {
         this.setSubRegionSize = batchSize;
     }
 
+    @Override
+    public void mapOnDeviceMemoryRegion(long executionPlanId, XPUBuffer srcPointer, long offset) {
+        throw new TornadoRuntimeException("[ERROR] not implemented");
```
stratika, line 351: can you elaborate when is this case triggered?

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/compute/ComputeTests.java
```diff
@@ -19,7 +19,7 @@
 
 import static org.junit.Assert.assertEquals;
 
-import java.awt.Color;
+import java.awt.*;
```
stratika, line 22: unfold imported classes

jjfumero: Yes, sorry the code formatter was filtering for awt classes.

jjfumero:
This feature only works for TornadoNativeArrays, (Panama Segments). We may introduce this feature for Java object arrays in the future.

stratika:
> This feature only works for TornadoNativeArrays, (Panama Segments). We may introduce this feature for Java object arrays in the future.

Thanks, yes it makes sense to me. Let's think to add this kind of functionality in the future.

stratika (approved):
Thank you, I have successfully tested the new functionality for all backends.

PR #630, opened 2025-03-05
[feat] Support for atomic addition in OpenCL and PTX

stratika:
#### Description

This PR adds a new feature that has been implemented to enable support for `atomicAdd` in the OpenCL and PTX backends.

At the API level, TornadoVM programmers can use the `KernelContext` object which exposes some new methods that can be replaced during compilation by the TornadoVM compiler with the corresponding OpenCL and PTX instructions.

The replacement is performed by the new plugins that have been added in the `OCLGraphBuilderPlugins` and `PTXGraphBuilderPlugins`.

##### Examples:
Here is an example of the `histogram` method with the `Parallel API`:
```java
public static void histogramKernel(KernelContext context, IntArray input, IntArray output) {
    for (@Parallel int tid = 0; tid < input.getSize(); tid++) {
        int index = input.get(tid);
        context.atomicAdd(output, index, 1);
    }
}
```

Here is an example of the `histogram` method with the `Kernel API`:
```java
public static void histogramKernel(KernelContext context, IntArray input, IntArray output) {
    int tid = context.globalIdx;

    if (tid < input.getSize()) {
        int index = input.get(tid);
        context.atomicAdd(output, index, 1);
    }
}
```

**Note:** In OpenCL double and float types are not supported. There is a work around by using `atomic_xchg` in OpenCL, but I have left it as unsupported for the time being. On the other hand, they are supported in PTX (See unit-tests `testAtomic18_parallel_api_float()`, `testAtomic18_parallel_api_double()`.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [ ] No

```bash
make BACKEND=opencl
tornado-test -V uk.ac.manchester.tornado.unittests.atomics.TestAtomics
tornado --threadInfo -m tornado.examples/uk.ac.manchester.tornado.examples.kernelcontext.reductions.Histogram 1024 256
```

```bash
make BACKEND=ptx
tornado-test -V uk.ac.manchester.tornado.unittests.atomics.TestAtomics
tornado --threadInfo -m tornado.examples/uk.ac.manchester.tornado.examples.kernelcontext.reductions.Histogram 1024 256
```


jjfumero (approved):

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/AtomicAddNodeTemplate.java
```diff
@@ -37,29 +37,26 @@
 import uk.ac.manchester.tornado.drivers.opencl.graal.lir.OCLUnary;
 
 @NodeInfo(shortName = "AtomicAdd")
-public class AtomicAddNodeTemplate extends FloatingNode implements LIRLowerable {
+public class AtomicAddNodeTemplate extends FixedWithNextNode implements LIRLowerable {
```
mairooni, line 40: why is this node changed from floating to fixed?

stratika: I did some experiments with this node being in a fixed position (not reordered). But, I can revert back since the unit-tests are passing.

I will double check also for the `Histogram` computation.

Do you have any strong suggestion of the one node type versus the other?

mairooni: No, I was wondering if it matters whether the node is rearranged in the graph or if you implemented this change for a different reason

mairooni (approved):

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/AtomicAddNodeTemplate.java
stratika: It seems that it does not matter.

stratika: I re-ran the tests, and it seems that it matters. With the Floating Node the kernel is not generated properly.

jjfumero:
From my view atomics need to be fixed nodes. Ordering is important and need need to lock them in specific positions in the control flow. We still can move those nodes around but it will be more difficult.

stratika:
> From my view atomics need to be fixed nodes. Ordering is important and need need to lock them in specific positions in the control flow. We still can move those nodes around but it will be more difficult.

Yes, good point.

PR #645, opened 2025-03-19
[rt] Memory dealloc by default when closing an execution plan resource

jjfumero:
#### Description

It forces to clean up all resources (e.g., all GPU memory used) when the `ExecutionPlan` resource is closed. 
The only exception is when using constant memory (as in CUDA constant) and atomics, which in TornadoVM, these
are separated memory regions handler by the device memory manager. 

In future versions of TornadoVM, we should improve these handler by taking into account the execution plan IDs, we memory management can be further improved. 

#### Problem description

When running with large data sets with multiple execution plans, TornadoVM tends to lock memory to reuse as much as possible. However, when a resource (execution plan) is closed, we can free all resources associated to an execution plan, and make more space for new execution plans. 

This is needed to run GAIA code under the AERO EU project. 

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
make
make tests
tornado-test --fast -V uk.ac.manchester.tornado.unittests.atomics.TestAtomics
```

### DIFF
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java
@@ -224,10 +224,10 @@ public class TornadoOptions {
      *
      * <p>
      * <ul>
-     *   <il>Use <code>-Dtornado.spirv.runtimes=opencl</code> for OpenCL only.
-     *   <il>Use <code>-Dtornado.spirv.runtimes=levelzero</code> for LevelZero only.
-     *   <il>Use <code>-Dtornado.spirv.runtimes=opencl,levelzero</code> for both OpenCL and Level Zero runtimes, being
-     *   OpenCL the first in the list (default).
+     * <il>Use <code>-Dtornado.spirv.runtimes=opencl</code> for OpenCL only.
+     * <il>Use <code>-Dtornado.spirv.runtimes=levelzero</code> for LevelZero only.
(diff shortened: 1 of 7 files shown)


tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java
```diff
@@ -459,4 +468,8 @@ private static boolean isFPGAEmulation() {
         String contextEmulatorXilinxFPGA = System.getenv("XCL_EMULATION_MODE");
         return (contextEmulatorIntelFPGA != null && (contextEmulatorIntelFPGA.equals("1"))) || (contextEmulatorXilinxFPGA != null && (contextEmulatorXilinxFPGA.equals("sw_emu")));
     }
+
+    public static boolean cleanUpAtomicsSpace() {
```
mikepapadim, line 472: can we document also this?

mikepapadim (approved):

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java
jjfumero: Yes, good point. I just pushed the javadoc.

mairooni (approved):
LGTM

PR #646, opened 2025-03-21
[hotfix] Fix to prevent issue with shared buffer allocation when not all buffers are present in args

mikepapadim:
#### Description

There was an issue with the shared buffer functionalty that it was taking into account that shared buffers might not be present in all tasks of taskgraph. This is it was causing negative allocation for some tests. Also, it adds tests to force  buffers to be copied out but adding intermediate dummy tasks. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash 
 tornado-test   -V uk.ac.manchester.tornado.unittests.api.TestSharedBuffers
```

### DIFF
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
@@ -978,23 +978,11 @@ public class TornadoVMInterpreter {
      * Container class that holds information about object allocation counts.
      * Used to track the number of persistent objects and the number of objects
      * that need to be allocated.
+     *
+     * @param persistentObjectCount Number of persistent objects that don't need allocation
+     * @param objectsToAlloc Number of objects that need to be allocated
      */
-    private static class ObjectAllocationInfo {
-        final int persistentObjectCount;
-        final int objectsToAlloc;
-
-        /**
-         * Creates a new object allocation information container.
-         *
-         * @param persistentObjectCount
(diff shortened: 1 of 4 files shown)


tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
```diff
@@ -950,6 +974,29 @@ public void clearInstalledCode() {
         Arrays.fill(installedCodes, null);
     }
 
+    /**
+     * Container class that holds information about object allocation counts.
+     * Used to track the number of persistent objects and the number of objects
+     * that need to be allocated.
+     */
+    private static class ObjectAllocationInfo {
```
jjfumero, line 982: I think you can use a Java `record` since they are both final

jjfumero (changes requested):

jjfumero:
It works for Linux and MacOS. All good

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
mikepapadim: This is fixed now

jjfumero (approved):

mairooni (approved):
LGTM

Issue #665, opened 2025-04-11
Document flags JVM option names


ItayShtainberg:
#### Issue description

Hi, in the docs at [flags.rst](https://github.com/beehive-lab/TornadoVM/blob/fd7cd7f62d41f329bdc3942af5349ce24e86f0ca/docs/source/flags.rst)
Some flags are shown in the format which the `tornado` binary would accept, for example:
`--printKernel`
However, when specifying JVM flags in intellij, this would not be accepted as a valid flag. 
Instead I need to use:
`-Dtornado.print.kernel=True`, And that flag is not documented in the docs.

#### The problem I faced

This caused me some trouble when debugging #663  
I assumed the option name would be `-Dtornado.printkernel=True`, and when the kernel didn't print, I thought it was due to an error, but it was actually the wrong flag.
I then had to look at the source code to find it's actually `-Dtornado.print.kernel=True`

#### My Suggestion
In the documentation, add the `-Dtornado...` flag syntax to each one of the options (many already have it)

Thanks,
Itay

stratika:
Thanks for the suggestion. 

I think it is important to clarify the flags, and update the relevant [documentation page](https://tornadovm.readthedocs.io/en/latest/flags.html). The reason that the JVM flag is different from the tornado flag, is that tornado is a Python program and we internally link the python flag (`--printKernel`) to the Java flag (`-Dtornado.print.kernel=True`). 

All Tornado flags start with `-Dtornado.`, as suggested and they reside in the [TornadoOptions.java](https://github.com/beehive-lab/TornadoVM/blob/master/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java) class.

We can refactor the Tornado Java flag to be more consistent with the Python `tornado` flags, if you believe it can help. But in my opinion, the programmer should be aware of the TornadoOptions class. And this awareness will come from the documentation page.

What do you think?

ItayShtainberg:
I agree, for me the issue is more about visibility of these flags in the documentation, and there's no need to change the names.
I now realize it also might be helpful if an invalid `-Dtornado...` option that doesn't exist would produce an error/warning.

stratika:
We merged the PR with the update of the documentation regarding the flags, and a refactoring of the JVM TornadoVM flag for printing Kernels.

See more [here](https://tornadovm.readthedocs.io/en/latest/flags.html).

I close this issue. Feel free to reopen it.

PR #667, opened 2025-04-14
[docs] Update documentation of the TornadoVM flags

stratika:
#### Description

This PR updates the documentation page regarding the TornadoVM flags. It provides the clarification suggested in #665 and refactors one TornadoVM JVM flag the `-Dtornado.print.kernel` to `-Dtornado.printKernel` in order to follow the same name convension with the Python CLI flag `--printKernel`. Similar approach is already in place for the `-Dtornado.threadInfo` flag.

#### Problem description

If the patch provides a fix for a bug, please describe what was the issue and how to reproduce the issue.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

To read the updated docs run:
```bash
make docs
```

### DIFF
--- a/docs/source/flags.rst
+++ b/docs/source/flags.rst
@@ -49,6 +49,8 @@ Debugging and Logging
    ``-Dtornado.printKernel=true``                    Prints generated OpenCL/PTX/SPIR-V kernels.
    ``-Dtornado.print.kernel.dir=FILENAME``           Saves generated kernels to the specified file.
    ``-Dtornado.threadInfo=true``                     Displays the number of threads used.
+   ``-Dtornado.print.bytecodes=true``                Prints TornadoVM Internal Bytecodes to stdout.
+   ``-Dtornado.dump.bytecodes.dir=FILENAME``         Dumps TornadoVM Internal Bytecodes to the specified file.
    ================================================  ============================================================================
 
 Profiling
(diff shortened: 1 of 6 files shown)


mairooni (approved):

docs/source/flags.rst
```diff
@@ -3,142 +3,161 @@
...
-There is a number of runtime flags and compiler flags to enable experimental features, as well as fine and coarse grain profiling in the context of TornadoVM.
+TornadoVM provides runtime and compiler flags to enable experimental features, tuning, and profiling. These flags fall into two categories:
 
-**Note:** for the following examples ``s0`` represents an arbitrary task-graph, as well as ``t0`` represents a given task's name.
+1. **JVM Flags** (passed with the `-D` prefix via the `--jvm` option)
+2. **TornadoVM CLI Flags** (passed directly to the `tornado` Python wrapper)
 
-All flags needs Java prefix of ``-D``. An example of tornado using a flag is the following:
+.. note::
+   In the examples below, ``s0`` refers to a task graph and ``t0`` to a specific task within that graph.
```
jjfumero, line 12: I think we can remove this

jjfumero (approved):
It would be good also to add what is the default value for each option.

docs/source/flags.rst
stratika: There are some flags that use `s0` and `t0`. So, I thought it would be nice to explain first, what they are.

PR #675, opened 2025-05-05
Apply rescheduling of true branches only for conditions

mairooni:
#### Description
This PR fixes an issue with the scheduling of true branches. The logic to prioritise true branches was applied for all `IfNode` nodes, however, the rescheduling is only necessary if the `IfNode` is associated with a condition (for instance if there is a `break` or a `return` in the code). In this PR, a check is performed to ensure that the `IfNode` is not associated with a `LoopBegin` node (to evaluate the loop bounds) and rather it is generated due to a condition in the code, and only then it applies the logic to reschedule the necessary branches. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run 
`tornado-test -V uk.ac.manchester.tornado.unittests.compute.ComputeTests#testGroupedQueryAttention `
and 
`make tests`


mikepapadim (approved):
Thanks @mairooni , LGTM

PR #678, opened 2025-05-20
[fix] Added guard conditions for UPS metrics when profiling across all backends

stratika:
#### Description

This PR provides a fix for a bug that was spotted on `macOS` systems. When profiling with the TornadoVM profiler was enabled, the system was trying to read metrics from the UPS system. However, this should not be enabled if the flag for UPS reading is not enabled (i.e., if the `-Dtornado.ups.ip` is empty).

This PR adds relevant conditions to perform this check.

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [x] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=opencl && make fast-tests
make BACKEND=ptx && make fast-tests
make BACKEND=spirv && make fast-tests
```


mikepapadim (approved):
LGTM, thanks @stratika

mairooni (approved):
LGTM

Issue #680, opened 2025-05-22
[codegen] End bracket of outer loop not generated for flash attention


mairooni:
### Describe the bug 
Currently, when running the flash attention, the end bracket of the outer loop which performs core the computation is not generated. The issue appears to be in the `closeScope` function of the `OCLBlockVisitor` class.  
```
private void closeScope(HIRBlock block, HIRBlock loopBeginBlock) {
        if (block.getBeginNode() instanceof LoopExitNode) {
            if (!(block.getDominator().getDominator() != null && block.getDominator().getDominator().getBeginNode() instanceof MergeNode)) {
                /*
                 * Only close scope if the loop-exit node does not depend on a merge node. In
                 * such case, the merge will generate the correct close scope.
                 */
                closeBlock(block);
                incrementClosedLoops(loopBeginBlock);
            }
        } else {
            closeBlock(block);
            incrementClosedLoops(loopBeginBlock);
        }
    }
```
Specifically, the issue is located in the condition `if (!(block.getDominator().getDominator() != null && block.getDominator().getDominator().getBeginNode() instanceof MergeNode))`. In the case of flash attention, this condition is not satisfied because there is a merge node two blocks above the node which is supposed to close the scope. However, the merge node does not close the scope when it is scheduled at a later stage, leaving the loop open. 

### How To Reproduce
Below is the code to reproduce the issue, with an arrow pointing to the loop which does not close in the OpenCL generated kernel. 

```
public static void processHeadsFlashAttention(
            KernelContext context,
            FloatArray q,
            FloatArray key_cache,
            FloatArray value_cache,
            FloatArray xb,
            int nHeads,
            int headSize,
            int kvDim,
            int kvMul,
            IntArray positionHolder,
            int layer,
            int contextLength) {

        // Thread and workgroup information
        int tid = context.localIdx;
        int gid = context.globalIdx; // gid is not actively used in the core logic here
        int h = context.groupIdx;  // Each workgroup processes one head
        int localSize = context.localGroupSizeX;

        // Early exit if this workgroup is beyond our head count
        // This relies on the kernel being launched with nHeads workgroups.
        if (h >= nHeads) return;

        int pos = positionHolder.get(0);
        int loff = layer * contextLength * kvDim;
        int kvHeadIdx = h / kvMul;
        int BLOCK_SIZE_C = 4;

        // Allocate shared memory for tiled computation
        float[] q_shared = context.allocateFloatLocalArray(headSize);
[...]

PR #681, opened 2025-05-22
Fix closing bracket for flash attention

mairooni:
#### Description

This PR addressed the code generation error reported in the issue #680.  In case there is a merge block two blocks above the `LoopEnd` block which is being examined by the `closeScope`, the rest of the loop exit points are checked. If there is only one loop exit point, the code in the PR checks if the only `LoopExit` node is followed by and `IfNode`, indicating nesting.  Since, in case of nesting, the `LoopExit` node is not expected to close the loop associated with it, the only other point which can close the loop is the `LoopEnd` block examined by the `closeScope`. Therefore, if this is the case, we proceed by closing the loop. 

Kfusion, TornadoVM-Raytracer and the unittests have all been checked and are not breaking with this fix. 

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run `tornado-test -V uk.ac.manchester.tornado.unittests.codegen.CodeGenTest#testFlashAttention`


mikepapadim (approved):
Thanks @mairooni

stratika (approved):

PR #686, opened 2025-06-04
[proposal] New API Functions for WarmUp

jjfumero:
#### Description

This PR extends the `TornadoExecutionPlan` API with two new calls for representing `warmup`. The current `warmup` functionality does not actually correspond to warmup. What it does it pre-compile a kernel from Java down to the selected backend (e.g., from Java bytecode to OpenCL C, or PTX). However, this call does not run the application, and it does not place data onto the accelerator. 

This PR tries to extend this functionality with two new calls:
- `withWarmUpWithTime(ms)`
- `withWarmUpIterations(numIterations)`

Besides, the old `withWarmUp` is set to `withPreCompilation()`, which only builds and install the generated code in the code-cache, with no data loading. 

#### Problem description

n/ a.

#### Backend/s tested

- [X] OpenCL
- [X] PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
tornado-test -V --live uk.ac.manchester.tornado.unittests.warmup.TestWarmUp
```

### DIFF
--- a/tornado-api/src/main/java/module-info.java
+++ b/tornado-api/src/main/java/module-info.java
@@ -16,7 +16,6 @@
  *
  */
 module tornado.api {
-    requires jmh.core;
     exports uk.ac.manchester.tornado.api;
     exports uk.ac.manchester.tornado.api.annotations;
     exports uk.ac.manchester.tornado.api.common;
--- a/tornado-api/src/main/java/module-info.java
+++ b/tornado-api/src/main/java/module-info.java
@@ -16,6 +16,7 @@
  *
  */
 module tornado.api {
+    requires jmh.core;
     exports uk.ac.manchester.tornado.api;
     exports uk.ac.manchester.tornado.api.annotations;
     exports uk.ac.manchester.tornado.api.common;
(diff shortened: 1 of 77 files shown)


mikepapadim (approved):
Thanks juan, it looks useful. I had to do manually something similar for Llama where i call  warmup:
https://github.com/beehive-lab/GPULlama3.java/blob/9b2dea306f48594a1c2c41e41001cc572c5a49ae/src/main/java/com/example/tornadovm/TornadoVMMasterPlan.java#L69

Followed by an empty warm-up run to place data on device:
https://github.com/beehive-lab/GPULlama3.java/blob/9b2dea306f48594a1c2c41e41001cc572c5a49ae/src/main/java/com/example/tornadovm/TornadoVMMasterPlan.java#L78

So, i had a method to that in API level:
https://github.com/beehive-lab/GPULlama3.java/blob/9b2dea306f48594a1c2c41e41001cc572c5a49ae/src/main/java/com/example/tornadovm/TornadoVMMasterPlan.java#L160


LGTM, one think is that might be useful  to have call to just place data on the GPU.
Something like `withPrecompilation().withPlaceData()`

PR #694, opened 2025-06-27
[fix] API exceptions calling parent to not stop stack-trace

jjfumero:
#### Description

Fix for the API exceptions calling parent to not stop stack-trace

#### Problem description

Stack trace can be lost if we do not call the parent constructor for the custom exceptions. 

#### Backend/s tested

- [X] OpenCL
- [X} PTX
- [X] SPIRV

#### OS tested

- [X] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

```bash
make
make tests
```


stratika (approved):
LGTM, thanks @jjfumero

mikepapadim (approved):

PR #697, opened 2025-07-01
[feat] Added non-interactive mode and restored backend and jdk console arguments

stratika:
#### Description

This PR reverts the `--jdk` and `--backend` arguments of the `tornadovm-installer` which were used in non-interactive mode to build docker images. Those arguments can now be used in both interactive and non-interactive modes.

Additionally, this PR adds a new argument that users can select to define the non-interactive modes. In this case, if user selects also the `--non-interactive` option, then no interactive menu will be displayed and all dependencies will be installed automatically by the script.

```bash
./bin/tornadovm-installer --help                                                                                                
usage: tornadovm-installer [-h] [--jdk JDK] [--backend BACKEND] [--version] [--listJDKs] [--polyglot] [--mvn_single_threaded] [--non-interactive]

TornadoVM Installer Tool. It will install all software dependencies except the GPU/FPGA drivers

options:
  -h, --help            show this help message and exit
  --jdk JDK             Specify a JDK to install by its keyword (e.g., 'jdk21', 'graal-jdk-21'). Run with --listJDKs to view all available JDK keywords.
  --backend BACKEND     Select the backend to install: { opencl, ptx, spirv }
  --version             Print version
  --listJDKs            List supported JDKs
  --polyglot            Enable Truffle Interoperability with GraalVM
  --mvn_single_threaded
                        Run Maven in single-threaded mode
  --non-interactive     Download and install dependencies without user interaction
```

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [ ] No

### DIFF
--- a/bin/tornadovm-installer
+++ b/bin/tornadovm-installer
@@ -631,7 +631,11 @@ class TornadoInstaller:
 
 def parseArguments():
     parser = argparse.ArgumentParser(description="TornadoVM Installer Tool.  It will install all software dependencies except the GPU/FPGA drivers")
-    parser.add_argument("--jdk", action="store", dest="jdk", default=None, help="Select one of the supported JDKs. Use --listJDKs option to see all supported ones")
+    parser.add_argument("--jdk", action="store", dest="jdk", default=None,
+                        help=(
+                            "Specify a JDK to install by its keyword (e.g., 'jdk21', 'graal-jdk-21'). "
+                            "Run with --listJDKs to view all available JDK keywords."
+                        ))
     parser.add_argument("--backend", action="store", dest="backend", default=None, help="Select the backend to install: { opencl, ptx, spirv }")
     parser.add_argument("--version", action="store_true", dest="version", default=False, help="Print version")
     parser.add_argument("--listJDKs", action="store_true", dest="listJDKs", default=False, help="List supported JDKs")
@@ -646,6 +650,7 @@ class UserInterface:
     """Handles user interactions."""
 
     def __init__(self, console: Console, system_info: SystemInfo):
+        self.args = args
         self.console = console
         self.system_info = system_info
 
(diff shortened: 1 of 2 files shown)


stratika:
> what is going to happen if i run:
> 
> `./bin/tornadovm-installer --jdk jdk21 --backend opencl ` ?

It will install TornadoVM with JDK 21 and OpenCL backend. But if you do not have maven or cmake in your path, it will ask you if you want to download and use them. Unless you  use the `--non-insteractive` which will automatically download and use them.

jjfumero:
> > what is going to happen if i run:
> > `./bin/tornadovm-installer --jdk jdk21 --backend opencl ` ?
> 
> It will install TornadoVM with JDK 21 and OpenCL backend. But if you do not have maven or cmake in your path, it will ask you if you want to download and use them. Unless you use the `--non-insteractive` which will automatically download and use them.

To me this is more confusing that before. Wouldn't be simpler if we pass parameters, then everything is automatic. If no parameters are passed, then it is interactive.

stratika:
> > > what is going to happen if i run:
> > > `./bin/tornadovm-installer --jdk jdk21 --backend opencl ` ?
> > 
> > 
> > It will install TornadoVM with JDK 21 and OpenCL backend. But if you do not have maven or cmake in your path, it will ask you if you want to download and use them. Unless you use the `--non-insteractive` which will automatically download and use them.
> 
> To me this is more confusing that before. Wouldn't be simpler if we pass parameters, then everything is automatic. If no parameters are passed, then it is interactive.

By parameters, you mean `--jdk` and `--backend` only? 

The only part that I see is that then if `mvn` or `cmake` are missing, they will be automatically downloaded and used, as it was before, without giving the option to the user to set the PATH. 

I propose to revert the flag of `non-interactive` to be something like `--autodownload`, so that users will select it and instruct the script to download any missing dependencies. Does it make sense?

jjfumero:
My concern is that this:

```bash
./bin/tornadovm-installer --jdk jdk21 --backend opencl 
```

is still interactive.

jjfumero:
So after a sync call with @stratika , the new installer script has new 3 modes of interaction:

1. Fully interactive: no parameters are passed `./bin/tornadovm-installer`
2. Semi-interaction (it depends on the dependencies previously installed in the system). Example: `./bin/tornadovm-installer --backend=opencl` 
3. Non interactive `./bin/tornadovm-installer  --non-interactive `

stratika:
I have also refactored the new argument to be `--auto-deps`:

```bash
./bin/tornadovm-installer --help
usage: tornadovm-installer [-h] [--jdk JDK] [--backend BACKEND] [--version] [--listJDKs] [--polyglot] [--mvn_single_threaded] [--auto-deps]

TornadoVM Installer Tool. It will install all software dependencies except the GPU/FPGA drivers

options:
  -h, --help            show this help message and exit
  --jdk JDK             Specify a JDK to install by its keyword (e.g., 'jdk21', 'graal-jdk-21'). Run with --listJDKs to view all available JDK keywords.
  --backend BACKEND     Select the backend to install: { opencl, ptx, spirv }
  --version             Print version
  --listJDKs            List supported JDKs
  --polyglot            Enable Truffle Interoperability with GraalVM
  --mvn_single_threaded
                        Run Maven in single-threaded mode
  --auto-deps           Automatic download and use any missing dependencies
  ```

PR #721, opened 2025-09-26
[fix]  Integer overflow in TornadoNativeArray implementations that caused `IllegalArgumentException` when allocating large arrays

orionpapadakis:
#### Description

Fix integer overflow in TornadoNativeArray implementations that caused `IllegalArgumentException` when allocating large arrays. The issue occurred when calculating `segmentByteSize` using integer arithmetic, causing overflow for arrays requiring more than ~2GB of memory.

#### Problem description

When creating large arrays (e.g., FloatArray with ~>540M elements), the calculation `numberOfElements * ELEMENT_BYTES + arrayHeaderSize` would overflow the integer range, resulting in negative values being passed to `Arena.allocate()`. This caused the following exception:

```java
Exception in thread "main" java.lang.IllegalArgumentException: Invalid allocation size : -1805647848
at java.base/jdk.internal.foreign.Utils.checkAllocationSizeAndAlign(Utils.java:206)
at java.base/jdk.internal.foreign.MemorySessionImpl.allocate(MemorySessionImpl.java:157)
at java.base/java.lang.foreign.Arena.allocate(Arena.java:273)
at [tornado.api@1.1.2-dev](mailto:tornado.api@1.1.2-dev)/uk.ac.manchester.tornado.api.types.arrays.FloatArray.<init>(FloatArray.java:62)
```

The fix involves casting `numberOfElements` to `long` before multiplication to ensure the calculation is performed using long arithmetic, preventing integer overflow.

In addition, it provides a new test class for large arrays (`uk.ac.manchester.tornado.unittests.arrays.TestLargeArrays`) along with a test for FloatArrays (`testLargeFloatArray()`).

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

To reproduce the original issue and verify the fix:

run the test with and without the fix:
```java
tornado-test -V -J"-Dtornado.device.memory=3GB" uk.ac.manchester.tornado.unittests.arrays.TestLargeArrays
```

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/arrays/TestLargeArrays.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/arrays/TestLargeArrays.java
@@ -1,20 +1,40 @@
+/*
+ * Copyright (c) 2025 APT Group, Department of Computer Science,
+ * The University of Manchester.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
 package uk.ac.manchester.tornado.unittests.arrays;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.stream.IntStream;
 
 import org.junit.Assume;
 import org.junit.Test;
+
 import uk.ac.manchester.tornado.api.ImmutableTaskGraph;
 import uk.ac.manchester.tornado.api.TaskGraph;
 import uk.ac.manchester.tornado.api.TornadoExecutionPlan;
 import uk.ac.manchester.tornado.api.annotations.Parallel;
 import uk.ac.manchester.tornado.api.enums.DataTransferMode;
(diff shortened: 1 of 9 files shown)


mikepapadim:
The proble is that `segmentByteSize` is `long` but `numberOfElements * FLOAT_BYTES + arrayHeaderSize` is `int * int + int`, so the result of the exceeds `Integer.MAX_VALUE` as the type of the rhs dictates the type.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/arrays/TestLargeArrays.java
mairooni, line 1: Add license header

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/arrays/TestLargeArrays.java
```diff
@@ -0,0 +1,89 @@
...
+    public static void addAccumulator(FloatArray a, float value) {
+        for (@Parallel int i = 0; i < a.getSize(); i++) {
+            a.set(i, a.get(i) + value);
+        }
+    }
+
+    @Test
+    public void testLargeFloatArray() throws TornadoExecutionPlanException {
+        //final int numElements = Integer.MAX_VALUE; // last overflow
+        //final int numElements = 600000000; // overflow
```
mairooni: Are these comments left to test different cases? Maybe the test could be split into two tests, one with a safe number of elements and one that overflows?

mikepapadim: this is fixed now, it has two tests. One for failing, one for passing

mikepapadim (approved):

PR #730, opened 2025-10-24
[build] Revamp build infrastructure by adding Maven wrapper `mvnw` 

mikepapadim:
TornadoVM's current build system requires:
  - Maven pre-installed with correct version
  - Remembering build flags (-T1.5C, -Pjdk21,opencl-backend)
  - Repeated full rebuilds even when nothing changed (22-26s every time)

  This creates friction for new contributors and wastes developer time.

  Solution

  Added three standard Maven improvements that make the project work like modern Java projects:

  1. 🔧 Maven Wrapper (./mvnw)

  - Contributors no longer need Maven pre-installed
  - Wrapper automatically downloads correct Maven version (3.9.6)
  - Standard practice in modern Java projects (Spring Boot, Gradle, etc.)

  Files added:
  - mvnw and mvnw.cmd - Unix/Windows wrapper scripts
  - .mvn/wrapper/maven-wrapper.properties - Wrapper configuration

  2. ⚙️ Default Maven Configuration (.mvn/maven.config)

  - Sensible defaults matching current bin/compile behavior
  - No need to remember build flags
  - Easy to override when needed

  Defaults:
  -T1.5C                          # Parallel builds (1.5 threads per CPU core)
  -Dstyle.color=always            # Colored output
  -Pjdk21,opencl-backend          # Default profiles

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -45,7 +45,7 @@ def get_mvn_command():
         str: The Maven wrapper command to use.
     """
     if os.name == 'nt':
-        return "./mvnw.cmd"
+        return ".\\mvnw.cmd"
     else:
         return "./mvnw"
 


Makefile
```diff
@@ -7,6 +7,12 @@ BACKEND ?= opencl
 build jdk21:
 	bin/compile --jdk jdk21 --backend $(BACKEND)
 
+incremental:
+	bin/compile --jdk jdk21 --backend $(BACKEND) --incremental
+
+incremental-full:
+	bin/compile --jdk jdk21 --backend $(BACKEND) --incremental --force-assembly
+
```
stratika: In lines 48-52, shouldn't we update to use the `mvnw`?

Makefile
```diff
@@ -7,6 +7,12 @@ BACKEND ?= opencl
 build jdk21:
 	bin/compile --jdk jdk21 --backend $(BACKEND)
 
+incremental:
+	bin/compile --jdk jdk21 --backend $(BACKEND) --incremental
+
+incremental-full:
```
stratika: Shall we change to something like `incremental-with-sdk` to make it more explicit that it should be used when building the sdk?

tornado-assembly/pom.xml
```diff
@@ -138,7 +138,7 @@
                             <appendAssemblyId>false</appendAssemblyId>
                             <finalName>tornado-sdk</finalName>
                             <outputDirectory>${project.basedir}/../dist</outputDirectory>
-                            <updateOnly>true</updateOnly>
+                            <updateOnly>false</updateOnly>
```
stratika, line 141: why did we change this flag?

tornado-assembly/pom.xml
```diff
@@ -162,6 +162,57 @@
...
+
+    backend_str = ','.join(backends) if backends else 'opencl'
+
+    # Write the backend file
+    with open(backend_file_path, 'w') as f:
+        f.write('tornado.backends=' + backend_str)
+
+    print('Created tornado.backend with backends: ' + backend_str)
+else:
+    print('Warning: SDK directory not found')
```
stratika: can you please explain this change? why is this argument necessary for `tornado-assembly` pom file?

Makefile
mikepapadim: fixed

tornado-assembly/pom.xml
mikepapadim: this is removed now

tornado-assembly/pom.xml
mikepapadim: Assembly plugin RECREATES the entire dist/ directory, this deletes and rebuilds everything to be sure on partial builds executables are the newly build.

stratika (approved):
LGTM, I tested it on macOS, Linux and Windows. It works. The only minor issue is that the `gen-tornado-argfile.sh` cannot be executed on Windows, so let's change it to a Python script in a separate PR.

thanks @mikepapadim, now we have one dependency less to consider.

PR #732, opened 2025-10-30
[types] Int8 support for PTX and OpenCL

mairooni:
#### Description
This PR adds support for int8 arrays in TornadoVM and implements the PTX DP4A instruction, which performs a dot product of four signed 8-bit integer pairs with accumulation into a 32-bit result.

##### DP4A Instruction Variants
Three implementations of DP4A are provided to support different use cases:

1. `dp4a(Int8Array a, long offset_a, Int8Array b, long offset_b, int c)`: Both operands from global memory (Int8Array)
2. `dp4a(Int8Array a_global, long offset_a, byte[] b_local, long offset_b, int c)`: One operand from global memory, one from shared/local memory 
3. `dp4a_packed(int a, int b, int c)`: Developer-packed format - four 8-bit integers packed into a single 32-bit integer

##### Additional Features
To support this functionality, the following features were additionally implemented:

1. Byte array allocation in local/shared memory
2. Predicate emission for constant boolean nodes in PTX code generation

##### Backend Support for DP4A
Since DP4A is a PTX-specific instruction, runtime exceptions are thrown when attempting to use it with OpenCL or SPIR-V backends.

##### Performance Evaluation
The int8 support was evaluated using the matrix-vector multiplication benchmarks (`uk.ac.manchester.tornado.examples.compute.MatrixVectorRowMajor`) with different quantization strategies:

- Quantized activations + weights: DP4A implementations (activations quantized on-the-fly, weights pre-quantized)
- Vectorized dequantization: Quantized weights with on-the-fly fp32 dequantization during computation

**Benchmark Results (vs FP16 baseline)**
Q8 Vectorized:        1.46x speedup  ← Best for Q8
Q8 DP4A 4-Way:        0.80x speedup
Q8 DP4A (standard):   0.74x speedup
Q8 DP4A Packed:       0.65x speedup
Q8 DP4A Local Memory: 0.56x speedup

As can be observed from the above numbers, vectorized dequantization outperforms the DP4A-based implementations. This is likely due to the overhead of quantizing the activations. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [x] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Run the unittest: 
`tornado-test -V uk.ac.manchester.tornado.unittests.quantization.QuantizationTests`
Run the benchmark for performance evaluation of the different implementations of the matrix vector multiplication:
`tornado -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixVectorRowMajor`

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/Int8Array.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/Int8Array.java
@@ -50,7 +50,7 @@ public final class Int8Array extends TornadoNativeArray {
         this.numberOfElements = numberOfElements;
         arrayHeaderSize = (int) TornadoNativeArray.ARRAY_HEADER;
         baseIndex = arrayHeaderSize / INT8_BYTES;
-        segmentByteSize = numberOfElements * INT8_BYTES + arrayHeaderSize;
+        segmentByteSize = (long) numberOfElements * INT8_BYTES + arrayHeaderSize;
 
         segment = Arena.ofAuto().allocate(segmentByteSize, 1);
         segment.setAtIndex(JAVA_INT, 0, numberOfElements);
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/utils/QuantizationUtils.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/utils/QuantizationUtils.java
@@ -19,6 +19,11 @@ package uk.ac.manchester.tornado.api.utils;
 
 import uk.ac.manchester.tornado.api.types.arrays.Int8Array;
 
+/**
+ * Utility class providing operators for quantized data types. Currently, the functions
+ * implemented are compatible with 8-bit quantized models (e.g., Q8_0), but it will be extended
+ * as support is added for additional quantization formats and optimization strategies.
+ */
 public class QuantizationUtils {
 
     public static int dp4a(Int8Array a, long offset_a, Int8Array b, long offset_b, int c) {


tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/compiler/PTXNodeLIRBuilder.java
```diff
@@ -609,6 +609,11 @@ private Variable emitLogicNode(final LogicNode node) {
             final Value x = getProcessedOperand(shortCircuitOrNode.getX(), shortCircuitOrNode.isXNegated());
             final Value y = getProcessedOperand(shortCircuitOrNode.getY(), shortCircuitOrNode.isYNegated());
             append(new AssignStmt(pred, new PTXBinary.Expr(PTXBinaryOp.BITWISE_OR, boolLirKind, x, y)));
+        } else if (node instanceof LogicConstantNode) {
+            LogicConstantNode logicConstantNode = (LogicConstantNode) node;
```
mikepapadim:
```suggestion
        } else if (node instanceof LogicConstantNode logicConstantNode) {
            LogicConstantNode logicConstantNode = logicConstantNode;
```

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
```diff
@@ -81,6 +80,337 @@ public final void emitCode(CompilationResultBuilder crb) {
...
+
+        @Def protected Value result;
+        private final boolean value;
+
+        public SetPredicateConstStmt(Value result, boolean value) {
+            super(TYPE);
+            this.result = result;
+            this.value = value;
+        }
+
```
mikepapadim, line 95: can we add a comment with the instruction it emits?

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
```diff
@@ -81,6 +80,337 @@ public final void emitCode(CompilationResultBuilder crb) {
...
+        @Def
+        protected Value add_header_offset_b;
+        @Def
+        protected Value offseted_address_a;
+        @Def
+        protected Value offseted_address_b;
+        @Def
+        protected Value header_size;
+
+        public Dp4aStmt(Value result, Value int8_a_base, Value load_four_int8_bytes_a, Value int8_b_base, Value load_four_int8_bytes_b, Value accumulator_c, Value offset_a, Value cnv_offset_a, Value add_header_offset_a, Value offset_b, Value cnv_offset_b, Value add_header_offset_b, Value offseted_address_a, Value offseted_address_b, Value header_size) {
```
mikepapadim, line 206: same

mikepapadim (commented):
Only some minor comments. Thanks @mairooni

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
mairooni: Comments added

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
```diff
@@ -81,6 +80,337 @@ public final void emitCode(CompilationResultBuilder crb) {
...
+            asm.emit("dp4a.s32.s32 ");
+            asm.emitValue(result);
+            asm.emitSymbol(COMMA + SPACE);
+            asm.emitValue(load_four_int8_bytes_a);
+            asm.emitSymbol(COMMA + SPACE);
+            asm.emitValue(load_four_int8_bytes_b);
+            asm.emitSymbol(COMMA + SPACE);
+            asm.emitValue(accumulator_c);
+            asm.delimiter();
+            asm.eol();
```
orionpapadakis, line 411: same here (`u32`)

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
mairooni: There are comments regarding the emitted instruction inside the `emitCode`. Would you like me to add also a comment before the constructor?

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
```diff
@@ -81,6 +80,337 @@ public final void emitCode(CompilationResultBuilder crb) {
...
```
orionpapadakis, line 304: from `dp4a` description ([here](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#:~:text=Description-,Four%2Dway%20byte%20dot,c%20has%20type%20.s32.,-Semantics)) it can also produce `u32` result if `a` and `b` are `u32`. Do we need to handle this as well ?

tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/Int8Array.java
```diff
@@ -0,0 +1,289 @@
...
+     * Constructs a new instance of the {@link Int8Array} that will store a user-specified number of elements.
+     *
+     * @param numberOfElements
+     *     The number of elements in the array.
+     */
+    public Int8Array(int numberOfElements) {
+        this.numberOfElements = numberOfElements;
+        arrayHeaderSize = (int) TornadoNativeArray.ARRAY_HEADER;
+        baseIndex = arrayHeaderSize / INT8_BYTES;
+        segmentByteSize = numberOfElements * INT8_BYTES + arrayHeaderSize;
```
orionpapadakis: this line could create overflow in case `INT8_BYTES != 1`. It is the same with `FloatArray` etc. Maybe we could add the long casting for consistency ?

```suggestion
        segmentByteSize = (long) numberOfElements * INT8_BYTES + arrayHeaderSize;
```

tornado-api/src/main/java/uk/ac/manchester/tornado/api/utils/QuantizationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.manchester.tornado.api.utils;
+
+import uk.ac.manchester.tornado.api.types.arrays.Int8Array;
+
+public class QuantizationUtils {
```
orionpapadakis, line 27: A class comment here would be very helpful (i.e. what are those utils intended to be used for? Are those methods transformed to ptx and how? Why do we need the java implementation when we have added compiler support for dp4a?)

stratika: In practise, this class has a Java implementation as if it would be run by the JVM. And in the `OCLGraphBuilderPlugins.java` class (and the corresponding PTX, SPIR-V) we call the Registration constructor for the `Quantization.class` and we replace the nodes associated with the compiler intrinsics we want for the corresponding operation (e.g., `dp4a(Int8Array a, long offset_a, Int8Array b, long offset_b, int c)`).

stratika: I agree with @orionpapadakis, let's add a Java doc that describes the utils in this class.

mairooni: Javadoc added

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
mikepapadim: no its fiine. thanks

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/lir/PTXLIRStmt.java
mairooni: That's a good point, however the reason we don't take this into account is that Q8 models (Q8_0 and Q8_1) use signed integers.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/quantization/QuantizationTests.java
```diff
@@ -0,0 +1,891 @@
...
+
+        for (int i = 0; i < output.getSize(); i++) {
+            assertEquals(outputSeq.get(i), output.get(i), 0.1f);
+        }
+
+    }
+
+    @Test
+    public void testMatrixVector4WayDP4AKernel() throws TornadoExecutionPlanException {
+        assertNotBackend(TornadoVMBackendType.OPENCL);
```
stratika, line 783: why this and previous tests are exempted for OpenCL?

Also, can you please @mairooni add the unit-tests in the test entries within the `tornado-test` to be executed by Jenkins?

mairooni: It is because dp4a is a PTX instruction, there is not OpenCL equivalent. The test is already in tornado-test

stratika: thanks, I missed the `tornado-test` file. sorry

mikepapadim (approved):
Just tested it locally! All works, thanks

stratika (approved):
LGTM

PR #733, opened 2025-10-31
Move argfile generation to python from bash

mikepapadim:
How to test:
```bash

make 

java @tornado-argfile -cp tornado-examples/target/tornado-examples-1.1.2-dev-e3cc6e2.jar  uk.ac.manchester.tornado.examples.compute.MatrixVectorRowMajor

```


stratika (approved):
I tested it on macOS and Windows and works fine. thanks

PR #736, opened 2025-11-13
Implement support for compressed oops (coops)

mairooni:
#### Description
This PR implements support on TornadoVM for compressed oops (coops). It makes coops the default, by removing the `-XX:+UseCompressedOops -XX:+UseCompressedClassPointers` options from the jvm flags, and by setting the default value of the array headers to 16. 
Even though the coops are now the default, running with uncompressed pointers is still supported.  Specifically, all tests run with coops unless specified (e.g. `make tests ` (runs tests with coops) vs `make tests_uncompressed `(runs tests without coops)). This option is provided also for individual tests, by using the `--uncompressed` flag to run without coops. For instance:

- `tornado-test -V uk.ac.manchester.tornado.unittests.fields.TestFields // runs unittest with coops`
- `tornado-test --uncompressed -V uk.ac.manchester.tornado.unittests.fields.TestFields // runs unittest without coops`

#### Problem description
To support this functionality, changes were required both on the runtime and the compiler side for all backends. 
On the runtime side, the main changes were in the way fields are accessed (written & read), as we need to store/access data through the compressed pointers. 
On the compiler side, two new nodes were introduced. The first one is the `[OCL/PTX/SPIRV]DecompressedReadFieldNode`, which replaces the  `ReadNode` during the lowering of a `LoadFieldNode`. This node is only inserted when loading non-primitive fields, and it generates the decompression logic to get the full address from the compressed pointer and access the data correctly. The second node is the `[OCL/PTX/SPIRV]FieldAddressArithmeticNode`. This node essentially replaces the `TornadoArithmeticNode` in the graph, when the latter is associated with an  `[OCL/PTX/SPIRV]DecompressedReadFieldNode`. This is necessary because the `[OCL/PTX/SPIRV]DecompressedReadFieldNode` emits the absolute address, while the `TornadoArithmeticNode` assumes that the address is missing the offset. These two nodes ensure that the data is accessed correctly through the compressed pointers. Additionally, the new read node was included in the exception checks for supported types (fp16, fp64). Furthermore, some hardcoded offsets were removed (e.g. in the code for the atomic plugins)  and instead were parameterised based on whether compression is used. 

Finally, new `_uncompressed` kernels were included for the prebuilt tests. These tests were updated with helper methods to dynamically select the appropriate kernel based on whether coops are used. 

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] SPIRV

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

`make fast-test // runs the "quick" tests with compressed pointers`
`make fast-test_uncompressed // runs the "quick" tests without compressed pointers (header of 24 bytes)`

`make tests // to run the whole set of tests with compressed pointers`
`make tests_uncompressed // runs the whole set of tests without compressed pointers (header of 24 bytes)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,14 +62,14 @@ fast-tests:
 	tornado-test --ea -V -J"-Dtornado.device.memory=1MB" uk.ac.manchester.tornado.unittests.fails.HeapFail#test03
 	test-native.sh
 
-tests_uncompressed:
+tests-uncompressed:
 	rm -f tornado_unittests.log
 	tornado --devices
 	tornado-test --ea --verbose --uncompressed
 	tornado-test --ea -V --uncompressed -J"-Dtornado.device.memory=1MB" uk.ac.manchester.tornado.unittests.fails.HeapFail#test03
 	test-native.sh
 
-fast-tests_uncompressed:
+fast-tests-uncompressed:
 	rm -f tornado_unittests.log
 	tornado --devices
 	tornado-test --ea --verbose --quickPass --uncompressed
--- a/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
+++ b/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
@@ -110,6 +110,7 @@ import uk.ac.manchester.tornado.drivers.opencl.graal.nodes.vector.LoadIndexedVec
 import uk.ac.manchester.tornado.drivers.opencl.graal.snippets.ReduceCPUSnippets;
 import uk.ac.manchester.tornado.drivers.opencl.graal.snippets.ReduceGPUSnippets;
 import uk.ac.manchester.tornado.runtime.TornadoVMConfigAccess;
+import uk.ac.manchester.tornado.runtime.common.TornadoOptions;
 import uk.ac.manchester.tornado.runtime.graal.nodes.GetGroupIdFixedWithNextNode;
 import uk.ac.manchester.tornado.runtime.graal.nodes.GlobalGroupSizeFixedWithNextNode;
 import uk.ac.manchester.tornado.runtime.graal.nodes.LocalGroupSizeFixedWithNextNode;
@@ -408,8 +409,7 @@ public class OCLLoweringProvider extends DefaultJavaLoweringProvider {
         Stamp loadStamp = loadStamp(loadField.stamp(NodeView.DEFAULT), field.getJavaKind());
         AddressNode address = createFieldAddress(graph, object, field);
         assert address != null : "Field that is loaded must not be eliminated: " + field.getDeclaringClass().toJavaName(true) + "." + field.getName();
-        final int headerSize = getVMConfig().getArrayBaseOffset(field.getJavaKind());
-        boolean areCoopsEnabled = (headerSize == 16);
+        boolean areCoopsEnabled = TornadoOptions.coopsUsed();
         // if coops are used and the field is a not a primitive (primitive data is not compressed)
         if (areCoopsEnabled && !field.getJavaKind().isPrimitive()) {
             OCLDecompressedReadFieldNode decompressedNode = graph.add(new OCLDecompressedReadFieldNode(object, address, loadStamp));
--- a/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoHalfFloatVectorOffset.java
+++ b/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoHalfFloatVectorOffset.java
@@ -37,6 +37,7 @@ import org.graalvm.compiler.phases.Phase;
 import uk.ac.manchester.tornado.api.types.arrays.TornadoNativeArray;
 import uk.ac.manchester.tornado.drivers.opencl.graal.lir.OCLAddressNode;
 import uk.ac.manchester.tornado.drivers.opencl.graal.nodes.vector.VectorValueNode;
+import uk.ac.manchester.tornado.runtime.common.TornadoOptions;
 import uk.ac.manchester.tornado.runtime.graal.nodes.VectorHalfRead;
 
 import java.util.Optional;
@@ -54,7 +55,7 @@ public class TornadoHalfFloatVectorOffset extends Phase {
     private static String object_offset;
 
     public TornadoHalfFloatVectorOffset() {
-        object_offset = (HEADER_SIZE == 16) ? "2" : "3";
+        object_offset = TornadoOptions.coopsUsed() ? "2" : "3";
     }
 
     @Override
--- a/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/phases/TornadoHalfFloatVectorOffset.java
+++ b/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/phases/TornadoHalfFloatVectorOffset.java
@@ -37,6 +37,7 @@ import org.graalvm.compiler.nodes.memory.address.OffsetAddressNode;
 import org.graalvm.compiler.phases.Phase;
 import uk.ac.manchester.tornado.api.types.arrays.TornadoNativeArray;
 import uk.ac.manchester.tornado.drivers.ptx.graal.nodes.vector.VectorValueNode;
+import uk.ac.manchester.tornado.runtime.common.TornadoOptions;
 import uk.ac.manchester.tornado.runtime.graal.nodes.VectorHalfRead;
 
 import java.util.Optional;
@@ -54,7 +55,7 @@ public class TornadoHalfFloatVectorOffset extends Phase {
     private static String object_offset;
 
     public TornadoHalfFloatVectorOffset() {
-        object_offset = (HEADER_SIZE == 16) ? "2" : "3";
+        object_offset = TornadoOptions.coopsUsed() ? "2" : "3";
     }
 
     @Override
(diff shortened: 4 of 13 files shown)


tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoHalfFloatVectorOffset.java
```diff
@@ -50,8 +50,16 @@
...
-    private static String OBJECT_OFFSET = "3";
     private static long HEADER_SIZE = TornadoNativeArray.ARRAY_HEADER;
+    private static String object_offset;
+
+    public TornadoHalfFloatVectorOffset() {
+        if (HEADER_SIZE == 16) { // coops
+            object_offset = "2";
+        } else {
+            object_offset = "3";
+        }
```
mikepapadim:
```suggestion
object_offset = (HEADER_SIZE == 16) ? 2 : 3;
```

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
```diff
@@ -406,10 +408,19 @@ protected void lowerLoadFieldNode(LoadFieldNode loadField, LoweringTool tool) {
...
+        // if coops are used and the field is a not a primitive (primitive data is not compressed)
+        if (areCoopsEnabled && !field.getJavaKind().isPrimitive()) {
+            OCLDecompressedReadFieldNode decompressedNode = graph.add(new OCLDecompressedReadFieldNode(object, address, loadStamp));
+            loadField.replaceAtUsages(decompressedNode);
+            graph.replaceFixed(loadField, decompressedNode);
+        } else {
+            FieldLocationIdentity fieldLocationIdentity = new FieldLocationIdentity(field);
+            ReadNode memoryRead = graph.add(new ReadNode(address, fieldLocationIdentity, loadStamp, BarrierType.NONE, GPU_MEMORY_MODE));
+            loadField.replaceAtUsages(memoryRead);
+            graph.replaceFixed(loadField, memoryRead);
```
mikepapadim: I think this can get out of the if

tornado-drivers/spirv/src/main/java/uk/ac/manchester/tornado/drivers/spirv/graal/phases/TornadoHalfFloatVectorOffset.java
```diff
@@ -53,8 +53,16 @@
...
-    private static String OBJECT_OFFSET = "3";
     private static long HEADER_SIZE = TornadoNativeArray.ARRAY_HEADER;
+    private static String object_offset;
+
+    public TornadoHalfFloatVectorOffset() {
+        if (HEADER_SIZE == 16) { // coops
+            object_offset = "2";
+        } else {
+            object_offset = "3";
+        }
```
mikepapadim:
```suggestion
    public TornadoHalfFloatVectorOffset() {
object_offset = (HEADER_SIZE == 16) ? 2 : 3;

        }
```

tornado-assembly/src/bin/tornado-test
```diff
@@ -365,6 +365,7 @@ __PRINT_OPENCL_KERNEL__ = "-Dtornado.printKernel=True "
 __DEBUG_TORNADO__ = "-Dtornado.debug=True "
 __TORNADOVM_FULLDEBUG__ = __DEBUG_TORNADO__ + "-Dtornado.fullDebug=True "
 __THREAD_INFO__ = "-Dtornado.threadInfo=True "
+__UNCOMPRESSED_POINTERS__ = "-XX:-UseCompressedOops -XX:-UseCompressedClassPointers -Dtornado.panama.objectHeader=24 "
```
mikepapadim: do we need to pass the header size if we want to disable the flags?
 ?

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
```diff
@@ -406,10 +408,19 @@ protected void lowerLoadFieldNode(LoadFieldNode loadField, LoweringTool tool) {
         Stamp loadStamp = loadStamp(loadField.stamp(NodeView.DEFAULT), field.getJavaKind());
         AddressNode address = createFieldAddress(graph, object, field);
         assert address != null : "Field that is loaded must not be eliminated: " + field.getDeclaringClass().toJavaName(true) + "." + field.getName();
-        FieldLocationIdentity fieldLocationIdentity = new FieldLocationIdentity(field);
-        ReadNode memoryRead = graph.add(new ReadNode(address, fieldLocationIdentity, loadStamp, BarrierType.NONE, GPU_MEMORY_MODE));
-        loadField.replaceAtUsages(memoryRead);
-        graph.replaceFixed(loadField, memoryRead);
+        final int headerSize = getVMConfig().getArrayBaseOffset(field.getJavaKind());
+        boolean areCoopsEnabled = (headerSize == 16);
```
mikepapadim: The 16 is we have coprossedops enabled/disabled it will be always 16?

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/phases/TornadoHalfFloatVectorOffset.java
```diff
@@ -50,8 +50,16 @@
 public class TornadoHalfFloatVectorOffset extends Phase {
 
     private static int HALF_SIZE = 2;
-    private static String OBJECT_OFFSET = "3";
```
mikepapadim, line 53: is this going to be 3 regardless ?

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/phases/TornadoHalfFloatVectorOffset.java
mairooni: done

tornado-assembly/src/bin/tornado-test
mairooni: these are in case someone was to run with uncompressed. They are included in the jvm options only if the `--uncompressed` flag is passed

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
mairooni: yes, the array header is always 16 with coops

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/phases/TornadoHalfFloatVectorOffset.java
mairooni: this was in develop, now we check whether coops are used and set it accordingly

tornado-drivers/spirv/src/main/java/uk/ac/manchester/tornado/drivers/spirv/graal/phases/TornadoHalfFloatVectorOffset.java
mairooni: done

Makefile
```diff
@@ -62,6 +62,20 @@ fast-tests:
...
 	test-native.sh
 
+tests_uncompressed:
+	rm -f tornado_unittests.log
+	tornado --devices
+	tornado-test --ea --verbose --uncompressed
+	tornado-test --ea -V --uncompressed -J"-Dtornado.device.memory=1MB" uk.ac.manchester.tornado.unittests.fails.HeapFail#test03
+	test-native.sh
+
+fast-tests_uncompressed:
```
stratika:
```suggestion
fast-tests-uncompressed:
```

Makefile
```diff
@@ -62,6 +62,20 @@ fast-tests:
...
+	rm -f tornado_unittests.log
+	tornado --devices
+	tornado-test --ea --verbose --quickPass --uncompressed
+	tornado-test --ea -V --uncompressed -J"-Dtornado.device.memory=1MB" uk.ac.manchester.tornado.unittests.fails.HeapFail#test03
+	test-native.sh	
+
```
stratika: Can you please add these rules also in the `Makefile.mak` which is used for Windows platform?

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/OCLLoweringProvider.java
stratika: ok, so this could be a point that we may need to adjust if/when we adopt object header sizes from Project Lilliput.

tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/graal/phases/TornadoHalfFloatVectorOffset.java
```diff
@@ -50,8 +50,12 @@
...
     private static long HEADER_SIZE = TornadoNativeArray.ARRAY_HEADER;
+    private static String object_offset;
+
+    public TornadoHalfFloatVectorOffset() {
+        object_offset = (HEADER_SIZE == 16) ? "2" : "3";
```
stratika: I see this `HEADER_SIZE == 16` appearing in many classes. Perhaps we could make a method that returns a boolean if it compressed oops are used or not, and invoke that method in this case.

This method could be in the `TornadoOptions` class, and be reused across all backend modules.

stratika (approved):
LGTM

PR #767, opened 2025-12-13
fix: better handling of NPE when key properties not specified, i.e. m…

maxandersen:
#### Description

if you are playing around with build or running and dont use tornado or some of the existing setup you will get NPE's a few places due to assumption system properties can't be null.

#### Problem description

try run without @argfile or tornado and you will see NPE normally, but with this there is a bit more helpful hint.


mikepapadim:
> fixed checkstyle.
> 
> not sure how to actually build cleanly so hoping it works ;)

Thanks! All checks are passing, it just needs the CLA

mikepapadim:
@maxandersen is the CLA possible from your side? otherwise I can implement it if its not possible. thanks for the contribution

PR #781, opened 2025-12-30
Fix OCLFieldBuffer to include inherited instance fields

ToTo3105:
#### Description
This patch improves the handling of instance fields in OCLFieldBuffer.
It extends field discovery to include inherited instance fields and ensures that each field is correctly resolved using its declaring class. As a result, objects with fields defined in superclasses are now properly wrapped and transferred to the device.
This enhancement makes OCLFieldBuffer fully compatible with class hierarchies and prevents incorrect or missing buffer mappings when dealing with inheritance.

#### Problem description

OCLFieldBuffer previously retrieved only the instance fields declared in the concrete class (getInstanceFields(false)) and attempted to resolve them using the runtime class of the object.
When an object contained fields declared in a superclass, reflection failed to resolve those fields correctly, causing them to be ignored or incorrectly wrapped.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [X] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

Tests have been implemented


mikepapadim:
@ToTo3105 it seems this fix has some side-effects and causing multiple other tests to fail -> https://github.com/beehive-lab/TornadoVM/actions/runs/20616171528/job/59209278211?pr=781

mikepapadim:
It seems that now the unit-tests fail in SPIRV, can you please apply the fix to SPIRV Field buffer to avoid CI to fail for spirv. thansk

mikepapadim:
thank you for your contibution @ToTo3105

PR #788, opened 2026-01-26
Introduce OpenCL sub-buffer aliasing for array fields

ToTo3105:
#### Description

This change introduces aliasing between object array fields and OpenCL buffers using sub-buffers.

OCLMemoryManager now tracks field-to-buffer mappings as (parent buffer, offset, size) and manages the creation and release of OpenCL sub-buffers, including reference counting and deferred parent release.

OCLFieldBuffer registers sub-buffers for array fields, aligns offsets according to GPU alignment requirements, avoids remapping array fields that are already registered, and coordinates deferred release of the parent buffer.

OCLArrayWrapper can remap an array to a sub-buffer, tracks the last execution plan ID to ensure proper synchronization, and correctly releases both sub-buffers and parent buffers.

OCLContext adds support for createSubBuffer at the OpenCL context level.

FieldBuffer.getFieldValue is made public to allow reuse in field registration and skip logic.

#### Problem description

Previously, arrays stored in object fields were allocated and copied as independent OpenCL buffers, even when the array data was already part of the object’s buffer.
This resulted in memory duplication, redundant data transfers, and uncoordinated buffer lifecycles between parent object buffers and field buffers.

With this change, array fields can now alias the corresponding region of the object’s OpenCL buffer via sub-buffers, ensuring correct alignment, eliminating unnecessary copies, and providing proper lifetime management between parent and child buffers.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [X] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

Implemented test cases

### DIFF
--- a/tornado-assembly/src/bin/tornado-test
+++ b/tornado-assembly/src/bin/tornado-test
@@ -189,6 +189,7 @@ __TEST_THE_WORLD__ = [
     TestEntry("uk.ac.manchester.tornado.unittests.logic.TestLogic"),
     TestEntry("uk.ac.manchester.tornado.unittests.fields.TestFields"),
     TestEntry("uk.ac.manchester.tornado.unittests.fields.TestInheritedFields"),
+    TestEntry("uk.ac.manchester.tornado.unittests.fields.TestSameVariableDifferentClass"),
     TestEntry("uk.ac.manchester.tornado.unittests.profiler.TestProfiler"),
     TestEntry("uk.ac.manchester.tornado.unittests.bitsets.BitSetTests"),
     TestEntry("uk.ac.manchester.tornado.unittests.quantization.QuantizationTests"),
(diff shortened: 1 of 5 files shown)


tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/fields/TestSameVariableDifferentClass.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * <p>
+ * How to test?
+ * </p>
+ * <code>
+ * tornado-test -V uk.ac.manchester.tornado.unittests.fields.TestSameVariableDifferentClass
+ * </code>
+ */
+public class TestSameVariableDifferentClass extends TornadoTestBase {
```
mikepapadim, line 21: Can you add tests with the Panama-based TornadoNativeArrays, like FloatArray, IntArray etc cause support to pritive array types has be depreacated.

PR #792, opened 2026-01-29
Introduce OpenCL sub-buffer aliasing for array fields

ToTo3105:
#### Description

This change introduces aliasing between object array fields and OpenCL buffers using sub-buffers.

OCLMemoryManager now tracks field-to-buffer mappings as (parent buffer, offset, size) and manages the creation and release of OpenCL sub-buffers, including reference counting and deferred parent release.

OCLFieldBuffer registers sub-buffers for array fields, aligns offsets according to GPU alignment requirements, avoids remapping array fields that are already registered, and coordinates deferred release of the parent buffer.

OCLArrayWrapper can remap an array to a sub-buffer, tracks the last execution plan ID to ensure proper synchronization, and correctly releases both sub-buffers and parent buffers.

OCLContext adds support for createSubBuffer at the OpenCL context level.

FieldBuffer.getFieldValue is made public to allow reuse in field registration and skip logic.

#### Problem description

Previously, arrays stored in object fields were allocated and copied as independent OpenCL buffers, even when the array data was already part of the object’s buffer.
This resulted in memory duplication, redundant data transfers, and uncoordinated buffer lifecycles between parent object buffers and field buffers.

With this change, array fields can now alias the corresponding region of the object’s OpenCL buffer via sub-buffers, ensuring correct alignment, eliminating unnecessary copies, and providing proper lifetime management between parent and child buffers.

#### Backend/s tested

- [X] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [ ] Linux
- [ ] OSx
- [X] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [X] No

Implemented test cases

### DIFF
--- a/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/mm/OCLFieldBuffer.java
+++ b/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/mm/OCLFieldBuffer.java
@@ -26,6 +26,7 @@
 package uk.ac.manchester.tornado.drivers.opencl.mm;
 
 import static uk.ac.manchester.tornado.api.exceptions.TornadoInternalError.shouldNotReachHere;
+import static uk.ac.manchester.tornado.runtime.TornadoCoreRuntime.getVMConfig;
 import static uk.ac.manchester.tornado.runtime.TornadoCoreRuntime.getVMRuntime;
 import static uk.ac.manchester.tornado.runtime.common.TornadoOptions.DEBUG;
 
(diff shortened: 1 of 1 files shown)


mikepapadim:
@ToTo3105 thank you, there are some tests failing

ToTo3105:
The creation of the sub-buffer is now correctly implemented. Before this latest push, the sub-buffer was transferred to the device after each allocation, which led to incorrect results. There is still an important issue that I am unable to resolve. During the creation of the TaskGraph, a DependentNode is created to synchronize execution between two distinct tasks. For example, if task t1 uses variable x and task t2 uses the object that contains x, the TaskGraph does not mark t2 as dependent on t1.

PR #796, opened 2026-02-06
[JDK21] Add `Apple Metal` backend  to run natively on Apple Silicon

mikepapadim:
This PR introduces TornadoVM's 4th hardware backend — a complete Metal driver that compiles Java bytecode into Metal Shading Language (MSL) and dispatches GPU kernels on Apple Silicon via the Metal API. The implementation adapts the architecture of the existing OpenCL backend to Apple's GPU programming model.

Status: Currently, builds and basic tests are passing

how to test:

```bash
make BACKEND=metal
```

image

-----

**To dos before merge:**
- [x] Pass rate to match OpenCL/PTX/SPIRV
- [ ] Check that works with GPULlama3.java
- [ ] Include SDK generation for SDKMAN distribution
- [x] Remove references to CL in the METAL backend and replace them with MSL term intead. For example, boilerplate `CL_BUILD_STATUS` ->`MSL_BUILD_STATUS`
- [x] Add Lisence headers when missing

Early POC perf analysis against OpenCL on M3-PRO:

image

**The Metal backend delivers meaningful GPU acceleration from pure Java on Apple Silicon out of the box, with no OpenCL dependency. Performance is within striking distance of OpenCL on large kernels and already faster on several workloads — with further MSL-specific optimizations (threadgroup memory, SIMD shuffle, async copies) still on the table.**

Metal Shadinng Language Spec:

https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf


tornado-assembly/src/etc/exportLists/metal-exports
```diff
@@ -0,0 +1,91 @@
+#
+# This file is part of Tornado: A heterogeneous programming framework:
+# https://github.com/beehive-lab/tornadovm
+#
+# Copyright (c) 2020, APT Group, Department of Computer Science,
```
stratika:
```suggestion
# Copyright (c) 2026, APT Group, Department of Computer Science,
```

tornado-assembly/src/bin/gen-tornado-argfile.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python3
```
stratika: This file is not necessary. Similar to other backends, we use the `gen-tornado-argfile-template` to generate the argfiles, with the explicit command:

```
tornado --generate-argfile
```

This command is also implicitly invoked when users run the `tornado` command.

stratika: I would remove it.

tornado-assembly/assembly.xml
```diff
@@ -134,6 +134,20 @@
...
+                <include>*.dll</include>
+            </includes>
+        </fileSet>
+        <fileSet>
+            <directory>../tornado-drivers/metal-jni/target/windows-amd64-release/cmake/Release</directory>
+            <outputDirectory>./lib/</outputDirectory>
+            <includes>
+                <include>*.dll</include>
+            </includes>
+        </fileSet>
```
stratika: These filesets are referencing Metal JNI Windows DLLs that will never exist, since Metal is macOS-only. I would remove those changes.

tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/enums/MetalBufferCreateType.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ */
+package uk.ac.manchester.tornado.drivers.metal.enums;
+
+public enum MetalBufferCreateType {
+    CL_BUFFER_CREATE_TYPE_REGION(0x1220);
```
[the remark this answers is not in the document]
stratika: @mikepapadim to fix this.

stratika (changes requested):

stratika (approved):
LGTM, I am ok to merge this. Let's do not delete the metal branch till we have the patch of metal for JDK 25.

Issue #828, opened 2026-04-15
F16 LLM inference produces wrong output (silent miscompilation) on OpenCL backend in Apple M4


stratika:
### Describe the bug 
On `develop` (also since v3.0.0-now), F16 (half-precision) LLM inference via https://github.com/beehive-lab/GPULlama3.java silently produces garbage output on the OpenCL backend - no crash, no bailout, just meaningless tokens. The same workload works on v2.2.0 and on the `jdk25` branch.

### How To Reproduce

Build [GPULlama3.java](https://github.com/beehive-lab/GPULlama3.java) and use any TornadoVM version > 2.2.0:
```bash
./llama-tornado --gpu --opencl --model /path/to/Llama-3.2-1B-Instruct-F16.gguf --prompt "Tell me a joke" 
```

### Expected behavior 

**Expected:** A joke (or at least coherent text). 
**Actual:** Garbage / unrelated tokens. No exception is thrown - the kernel compiles and runs, it just produces wrong results.

### Computing system setup (please complete the following information):
 - OS: macOS Tahoe 26.3.1
Affected TornadoVM versions:
  - TornadoVM `v2.2.0`: works 
  - TornadoVM `v3.0.0`: breaks (contains the merged PR #787)
  - TornadoVM `develop` (`v4.0.0`): broken
  - TornadoVM `jdk25` branch: works (has the fixes)

### Root cause

Two independent bugs in the OpenCL Graal parser plugins, both introduced by PR #787 (`prep-work-jdk25`) and both already fixed on the `jdk25` branch but never backported to `develop`.

---------------------------------------------------------------------

stratika:
I am working on it and will provide a fix soon.

PR #787, opened 2026-01-26
[refactor] Prepare compiler and API infrastructure for Jdk25 migration 

mikepapadim:
This PR contains prep-work for JDK25 and Graal 25+ migration.

* Removed tensor API classes and ONNX dependency from unittests
* Introduced TornadoMemorySegment wrapper class for unified memory segment management
* Refactored all array classes to use TornadoMemorySegment instead of direct MemorySegment access
* Removed deprecated TornadoVMIntrinsics class and related compiler plugins
* Updated compiler plugins to use new memory access registration pattern

PR #829, opened 2026-04-16
[fix] Backport jdk25 OCL plugin fixes: 64-bit address arithmetic, node insertion, and atomics

stratika:
#### Description

Fixes silent F16 (half-precision) miscompilation and a related compilation failure in the OpenCL backend's Graal parser plugins (#828), and adds a null-check guard in the SPIR-V half-float plugin for consistency.
                                                                                                                                                                                     
The OpenCL backend produces incorrect output (no exception, no bailout) when running F16 LLM inference workloads such as https://github.com/beehive-lab/GPULlama3.java on `develop`  (for any version after `v2.2.0`). The same workloads work on `v2.2.0` and on the `jdk25` branch - the fixes already exist on jdk25 but were never backported to develop.

##### Bug 1 - silent miscompilation (OpenCL r`egisterMemoryAccessPlugins`)

The `TornadoMemorySegment.get*AtIndex / setAtIndex` invocation plugins computed the byte offset in 32-bit int:

```java
AddNode absoluteIndexNode = b.append(new AddNode(index, baseIndex));
MulNode mulNode = b.append(new MulNode(absoluteIndexNode, ConstantNode.forInt(kind.getByteCount())));
AddressNode addressNode = b.append(new OffsetAddressNode(receiver.get(true), mulNode));
```
                                                                                                                                                                        
For an F16 weight tensor at the scale of a 1B-parameter Llama model (~2 GB byte offset range), (baseIndex + index) * elementSize exceeds 2³¹ and wraps to a negative byte offset.

The `OffsetAddressNode` then reads from the wrong GPU address. The kernel compiles cleanly and runs to completion — the only symptom is that the model produces unrelated tokens. Unit tests miss this because they use small arrays where the offset fits in 32 bits.                                                                                               

PTX (PTXGraphBuilderPlugins) and SPIR-V (SPIRVGraphBuilderPlugins) already do 64-bit address arithmetic via the same SignExtendNode pattern; only OpenCL lost it during the TornadoMemorySegment interception rewrite in PR #787.
                                                                                                                                                                                     
##### Bug 2 — compilation failure (OpenCL `OCLHalfFloatPlugins`)                                                                                                                           
  
After fixing bug 1, F16 kernels failed earlier in the pipeline with a SchedulePhase NPE inside PartialEscapePhase:                                                                                   
```bash
  java.lang.NullPointerException
    at org.graalvm.compiler.graph.NodeIdAccessor.getNodeId(NodeIdAccessor.java:64)
    at org.graalvm.compiler.graph.NodeMap.get(NodeMap.java:54) 
    at org.graalvm.compiler.phases.schedule.SchedulePhase$Instance.processNodes(SchedulePhase.java:963)
```
                                                                                                                                                                                
  Two causes in OCLHalfFloatPlugins:                                                                                                                                                 

1. The `HalfFloat.<init> NodePlugin` did both `b.append()` and `b.add()` on the same `NewHalfFloatInstance`, then returned `true` to suppress Graal's normal init processing. This left the  frame state inconsistent and inserted the node twice.
2. The `add/sub/mult/div/getHalfFloatValue` invocation plugins called `b.append(new XxxHalfFloatNode(...))` for nodes whose class extends `ValueNode` directly (not `FixedNode` or `FloatingNode`). The corrected pattern is `b.getGraph().addOrUnique(node)` followed by `b.push(kind, node)`, capturing the return value of `addOrUnique` so the unique instance ends up referenced - the local `new` instance may not be the one actually in the graph.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [ ] SPIRV

#### OS tested

- [x] Linux
- [x] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Requires a built https://github.com/beehive-lab/GPULlama3.java and an F16 GGUF model:                                                                                              

Update the pom file to use the current TornadoVM version of this branch (`4.0.1-dev`) and then build it using:
```bash
make
```

and run it:

```bash
  ./llama-tornado --gpu --opencl \                                                                                                                                                   
      --model /path/to/Llama-3.2-1B-Instruct-F16.gguf \                                                                                                                              
      --prompt "Tell me a joke"                                                                                                                                                      
```

- Before this PR: Either a SchedulePhase NPE during compilation, or (depending on which bug is hit first) the model runs and produces unrelated/garbage tokens.
- After this PR: Model produces a coherent joke, identical to `v2.2.0` / `jdk25` behavior.

### DIFF
--- a/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/plugins/OCLGraphBuilderPlugins.java
+++ b/tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/plugins/OCLGraphBuilderPlugins.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024-2025, APT Group, Department of Computer Science,
+ * Copyright (c) 2022, 2024-2026, APT Group, Department of Computer Science,
  * School of Engineering, The University of Manchester. All rights reserved.
  * Copyright (c) 2018, 2020, APT Group, Department of Computer Science,
  * The University of Manchester. All rights reserved.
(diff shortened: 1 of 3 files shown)


mikepapadim (approved):
LGTM, thanks

PR #838, opened 2026-05-26
[fix] Honour compile-time local work size in SPIR-V Level Zero launcher

stratika:
#### Description
There was a bug causing the CI/CD pipeline to break when running SPIR-V backend with the LevelZero runtime. This PR includes the following:

- Fix silent miscompare in SPIR-V Level Zero launcher caused by using the driver-suggested workgroup size instead of the size the kernel was compiled for.
- Drop the `USE_LEVELZERO_THREAD_DISPATCHER_SUGGESTIONS` option (`-Dtornado.spirv.levelzero.thread.dispatcher`). This was asking the driver to suggest local workgroup, but it was causing miscalculated items in case of reductions. This PR sets the compiled-for size and only falls back to `zeKernelSuggestGroupSize` if the driver rejects the GridScheduler-supplied size.

#### Problem description

`SPIRVLevelZeroInstalledCode#suggestThreadSchedulingToLevelZeroDriver` called `zeKernelSuggestGroupSize` and then handed the driver's suggestion to `zeKernelSetGroupSize`. On Intel Iris Xe (Level Zero) the suggestion for a 512-thread global is a single workgroup of 512.

The SPIR-V kernel, however, was compiled with statically-sized workgroup memory matching the local size TornadoVM picked at compile time (e.g. 128 floats for the auto-generated per-WG reduction kernel `testFloat`). Launching the same binary with `local=512` causes threads with `localId >= 128` to write out of bounds in `OpTypePointer Workgroup` arrays, corrupting adjacent kernel state and producing silently-wrong results with random sign.

The OpenCL runtime was unaffected: it has always launched with the compiled-for local size via `clEnqueueNDRangeKernel`.

## Fix

Always pass the local work size computed by `calculateGlobalAndLocalBlockOfThreads` to `zeKernelSetGroupSize`. The
`zeKernelSuggestGroupSize` fallback remains, but only fires when a user-supplied `GridScheduler` size is rejected by the driver with `ZE_RESULT_ERROR_INVALID_GROUP_SIZE_DIMENSION` - the original reason that branch existed.

Removes `USE_LEVELZERO_THREAD_DISPATCHER_SUGGESTIONS` entirely; the if/else it gated is gone and re-enabling driver-suggested sizes would just drive accidental problems.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [x] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

Reduction + matrix unit tests that previously failed only on `device=0:1` (SPIR-V / Level Zero) now pass:
```bash
make BACKEND=spirv
tornado-test --ea --verbose --jvm="-Dtornado.unittests.device=0:1" -qp
```

### DIFF
--- a/tornado-drivers/spirv/src/main/java/uk/ac/manchester/tornado/drivers/spirv/graal/SPIRVLevelZeroInstalledCode.java
+++ b/tornado-drivers/spirv/src/main/java/uk/ac/manchester/tornado/drivers/spirv/graal/SPIRVLevelZeroInstalledCode.java
@@ -2,7 +2,7 @@
  * This file is part of Tornado: A heterogeneous programming framework:
  * https://github.com/beehive-lab/tornadovm
  *
- * Copyright (c) 2021, 2024, APT Group, Department of Computer Science,
+ * Copyright (c) 2021, 2024, 2026, APT Group, Department of Computer Science,
  * School of Engineering, The University of Manchester. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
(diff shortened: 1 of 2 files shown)


mairooni (approved):
LGTM

PR #841, opened 2026-05-26
Add swizzled local memory accessors for FP16 and INT8 on the PTX backend

mairooni:
#### Description
This PR adds swizzled shared-memory accessors to the `KernelContext` for the PTX backend, providing bank-conflict-free layouts for FP16 and INT8 matrix tiles. The new accessors exposed through the `KernelContext` are:

- `swizzleLoadFp16Stride32` / `swizzleStoreFp16Stride32`
- `swizzleLoadFp16Stride16` / `swizzleStoreFp16Stride16`
- `swizzleLoadInt8` / `swizzleStoreInt8`

Each applies an involutive XOR permutation to the logical `(row, col)` coordinate before accessing shared memory, so that the resulting access pattern spreads across distinct memory banks instead of colliding. On NVIDIA GPUs shared
memory has 32 banks of 4 bytes each. A naive row-major tile layout causes many threads in a warp to hit the same bank, serializing the access. The XOR rotates each row's bank assignment to avoid this.

The constants follow the CUTLASS `Swizzle<>`, parameterized per layout:
- **FP16 stride-32**: permutation at the 16-byte boundary over groups of 8 rows (a 32-byte row holds 16 fp16 elements).
- **FP16 stride-16**: the same shape shifted down one bit, for a 16-byte row (8 fp16 elements), used for narrower, transposed tiles.
- **INT8**: identical math to FP16 stride-32. The permutation operates on a 16-byte granularity that is independent of element type, so int8 reuses the stride-32 constants directly. Because int8 elements are one byte, the caller expresses the row stride in bytes (32 for a wide tile, 16 for a narrow one) and the same swizzle serves both.

The layout is primarily intended for staging matrix tiles for future Tensor Core (MMA) work, but is a general bank conflict-avoidance mechanism usable by any kernel with a matching access pattern.

#### Problem description
Efficient Tensor Core (MMA) matrix multiplication requires shared-memory tiles to be laid out so that warp-level matrix loads do not incur bank conflicts. Currently shared arrays are addressed linearly, which produces heavy bank conflicts for the strided access patterns matrix loads use. This PR adds the swizzled-layout support needed to lay those tiles out in shared memory conflict-free, ahead of the MMA work that will consume it.

#### Backend/s tested

- [ ] OpenCL
- [x] PTX
- [ ] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

The load/store functionality can be verified by running the unittest:
 `tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.local.memory.TestSwizzledLocalArrays` 

Each kernel in the test was also profiled with Nsight Compute and produces zero shared-memory bank conflicts (`l1tex__data_bank_conflicts_pipe_lsu_mem_shared_op_ld/st = 0`), with non-zero `smsp__inst_executed_op_shared_ld/st`, confirming the accesses execute.

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/KernelContext.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/KernelContext.java
@@ -304,4 +304,206 @@ public class KernelContext implements ExecutionContext {
     @Override
     public void atomicAdd(DoubleArray array, int index, double val) {
     }
+
+    /**
+     * Loads a half-float from a swizzled shared-memory tile (stride-32 layout).
+     *
+     * <p>Applies an XOR permutation to the logical {@code (row, col)} coordinate so that
+     * the resulting shared-memory access pattern avoids bank conflicts. On NVIDIA GPUs, shared memory has
+     * 32 banks of 4 bytes each; a naive row-major tile layout causes many threads in a warp
+     * to hit the same bank, serializing the access. The XOR rotates each row's bank
+     * assignment so consecutive rows spread across distinct banks.
+     *
+     * <p>The swizzle is computed on the <em>byte</em> offset:
+     * <pre>{@code
+     *   byteOffset = (row * stride + col) * 2          // fp16 = 2 bytes per element
+     *   swizzled   = byteOffset ^ (((byteOffset >> 7) & 0b111) << 4)
+     * }</pre>
+     *
+     * <p>The three constants:
+     * <ul>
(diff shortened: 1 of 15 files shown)


stratika (commented):
LGTM, please consider and try the comment from co-pilot.

mairooni:
The comments have been applied on both this PR and PR #843

stratika (approved):

PR #846, opened 2026-06-11
[OpenCL] [Metal] Fix unbalanced brackets in code generation for nested control flow

stratika:
**Description**
  
 This PR fixes three independent code-generation defects in the OpenCL block structurizer (`OCLBlockVisitor`) that produced syntactically invalid OpenCL C (unbalanced brackets / dangling else) for kernels with non-trivial control flow, and ports the fixes to the Metal backend (`MetalBlockVisitor`, a fork  of the same visitor). It also adds two reproducer unit tests and documents a separate SPIR-V correctness issue.

  **Fix 1** — if-branch closes mis-counted as loop closes (closeIfBlock). Closing any if/else branch bracket incremented closedLoops for the branch's
  enclosing loop, even though no loop scope was closed. Once the counter reached the loop's exit count, wasLoopBlockAlreadyClosed() suppressed the
  closing brackets of all subsequent if-branches in the same loop body. The increment is now restricted to branches that actually terminate the loop
  body (end with a LoopEndNode or begin with a LoopExitNode).

  **Fix 2** — deferred loop-end close that no rule emitted (closeScope / exit). When a loop-end block beginning with a LoopExitNode depends on a merge
  block, closeScope() defers the close, assuming "the merge will generate the correct close scope" — but for some CFG shapes no rule emitted it, losing
  the loop bracket. The deferral is now tracked explicitly (pendingLoopEndClose) and emitted at the merge block's exit, guarded by
  allLoopBlocksEmitted() so that loops whose body blocks are traversed after the merge (e.g. CodeGenTest#test02) keep the previous behaviour.

  **Fix 3** — loop close emitted inside an open if-branch (closeScope). For loops of the shape for { if (c) { ... back-edge } else { break } } (produced by
  compound exit conditions such as i < n && !found), the loop bracket was emitted right after the back-edge block, which lexically sits inside the
  if-branch scope. The bracket closed the if-branch instead of the loop and left the else dangling. The close is now suppressed when the back-edge is
  dominated by a still-open if-branch scope (isNestedInOpenIfBranchScope); the bracket is then emitted by the existing rules at the loop's remaining
  exit block, the lexical end of the loop.

  **_New tests:_**
  - `TestFlashAttentionKernelContext`: standalone reproducer for the GPULlama3 flash-attention kernel (triggers fixes 1 and 2), validated against a
  sequential softmax-attention reference. Excluded on SPIR-V (see below).
[...]

PR #848, opened 2026-06-11
  Replace heuristic brace-placement with structured control-flow recovery (OpenCL + Metal)

mikepapadim:
## Summary

Code generation for the OpenCL and Metal backends reconstructs nested, brace-matched
control flow (`if/else`, `for`, `switch`, `{ … }`) from Graal's **flat** control-flow graph
(CFG) of basic blocks. Today this is done by `OCLBlockVisitor` / `MetalBlockVisitor`, which
decide *where every `{`, `else`, and `}` goes* using ~20 local pattern-matching heuristics plus
a manual "which braces have I already emitted" ledger, plus a traversal-reordering hack in the
result builder.

This is brittle: every new kernel shape that mis-braces gets a **new special case**. The
comments in the visitor literally cite individual unit tests and PRs as the reason a branch
exists. PR #846 is itself six commits of exactly this kind of patch
(*"emit missing closing brackets for if-branches inside loops"*, *"do not close loop scope at a
back-edge nested in an open if-branch"*).

This PR replaces that approach with the standard **two-phase structured control-flow recovery**
used by every structured target (WASM relooper, LLVM stackifier, SPIR-V):

1. **Structure** – build an explicit, properly-nested control-flow **AST** from the dominator
   tree + loop tree.
2. **Emit** – a stateless recursive walk of the AST. **Braces match by construction**, because
   they come from a tree rather than from counting opens/closes across a flat walk.

Net result: the whole class of mismatched-brace bugs becomes structurally impossible, ~1.5k
lines of heuristics are deleted, and PR #846's reproducers pass **without** PR #846's patches.

---

## The problem, visually

The compiler is handed a graph but must print nested text:

```
   Graal CFG (a graph)                        OpenCL C (nested text)
   ───────────────────                        ──────────────────────
        ┌────┐                                for (;i<N;) {
        │ B1 │ loop header                       if (cond) {
        └─┬──┘                                       break;
          │                                      } else {
        ┌─▼──┐  if(cond)                            ...
        │ B2 │────────┐                          }
        └─┬──┘        │                       }
     true │        false│                     ...
        ┌─▼──┐      ┌──▼─┐
        │ B4 │      │ B3 │   ← who emits the matching '}' for each,
        │break      │ i++│     and in what order?  ← the hard part
        └────┘      └────┘
```

`OCLBlockVisitor.enter()` / `exit()` answer "do I open/close a scope here?" with heuristics like
`checkClosingBlockInsideIf`, `updateListEndBracketsForLoopExitNodes`, `isReturnBranchWithMerge`,
`isComplexLoopCondition`, `isBlockInABreak`, … guarded by an 8-field ledger
(`openBlocks`, `closedBlocks`, `rmvEndBracket`, `switchClosed`, `merges`, `pending`, …).

---

## The new approach

Only the **structural scaffolding** is replaced. The straight-line content of each block — the
`if (cond)`, `for(;cond;)`, `switch(x)` headers and all the assignments/loads — is still emitted
by the existing LIR ops via `emitBlock`. The structurer emits only: if-branch `{`, `else`,
`case:` labels, and **every** closing `}`.

### Phase 1 — build a control-flow AST

```
sealed interface Region
  ├─ Seq(List<Region>)                 // straight-line sequence
  ├─ Block(HIRBlock)                   // emit one block's LIR
  ├─ IfElse(head, thenRegion, elseRegion?)
  ├─ LoopRegion(header, body)
  └─ SwitchRegion(head, List<Case>)
```

The builder walks the CFG and nests by two stable facts (**not** post-dominators — see note):

- **Loops are scopes.** A loop body is built within its own scope; the first block reached that
  is *not in the loop* (`!loop.getBlocks().contains(b)`) is a `break`/exit landing pad — emit it
  (its LIR holds the `break;`) and stop. This single rule replaces the entire break/continue/
  early-return special-case family.
- **An `if`/`switch` re-merges at its dominator join** — the block whose *immediate dominator*
  is the split and that has >1 predecessor. If there is none, the branches don't re-merge (both
  break/return) and there is simply no code after the split at that level.

> **Why not `getPostdominator()`?** Graal frequently leaves post-dominators **null** for blocks
> inside loops or near merges. Dumping the break example confirms it — `B1`, `B2`, `B3`, `B6`
> all report `postdom = -`. The builder therefore derives joins from the dominator tree, which
> is always populated.

### Phase 2 — emit (stateless)

```
emit(Seq)      → emit each child
emit(Block b)  → emitBlock(b)
emit(Loop)     → emitLoopBlock(header);  "{"(from LIR)  emit(body)  "}"
emit(IfElse)   → emitBlock(head);  "{"  emit(then)  "}"  [ "else {" emit(else) "}" ]
emit(Switch)   → emitBlock(head);  "{"  for each case: label; emit(body); "break;"   "}"
```

Because each `}` is produced by returning from a recursion that produced its `{`, they cannot
get out of sync.

---

## Worked example: `break` inside a loop

Source (`CodeGenTest#breakStatement`):

```java
for (int i = 0; i < a.getSize(); i++) {
    if (a.get(i) == 5) break;
    a.set(i, a.get(i) + 5);
}
a.set(0, 0);
```

**CFG facts** (from `-Dtornado.opencl.structuredcfg.debug=true`):

```
B0  end=End      loop=-   dom=-   succ=[1]
B1  end=If       loop=H1  dom=0   succ=[2,5]     ← loop header (B2 in loop, B5 exits)
B2  end=If       loop=H1  dom=1   succ=[4,3]     ← user if
B3  end=LoopEnd  loop=H1  dom=2   succ=[1]       ← back-edge → header
B4  end=End      loop=-   dom=2   succ=[6]       ← break landing (outside loop)
B5  end=End      loop=-   dom=1   succ=[6]       ← natural exit landing
B6  end=Return   loop=-   dom=1   succ=[]        ← merge (preds 5,4)
```

**Recovered AST:**

```
Seq
├─ Block(B0)                         ul0 = a
├─ LoopRegion(B1)                    for (;i<N;)
│    └─ Seq
│         └─ IfElse(B2)              if (a[i]==5)
│              ├─ then: Block(B4)        break;          (out-of-loop ⇒ break, stop)
│              └─ else: Block(B3)        i++ etc.        (succ=B1 header ⇒ stop)
├─ Block(B5)                         (empty natural-exit landing)
└─ Block(B6)                         a[0]=0; return
```

**Emitted OpenCL** (post-loop code correctly *after* the loop — no `rmvEndBracket` dance):

```c
ul_0 = (ulong) a;
for (;i_1 < 8192;) {
    if (b_7) {
        break;
    } else {
        i_8 = *((__global int *) ul_5);
        ...
        i_1 = i_10;
    }
}
ul_11 = ul_0 + 16L;
*((__global int *) ul_11) = 0;
return;
```

---

## What changed

| File | Change |
|------|--------|
| `OCLStructuredControlFlow.java` | **new** — AST + builder + emitter (~396 lines) |
| `OCLBlockVisitor.java` | **deleted** (−573) |
| `OCLCompilationResultBuilder.java` | call the structurer; remove `traverseControlFlowGraph`, `rescheduleTrueBranchConditionsIfNeeded`, `rescheduleBasicBlock`, `pending`, `isLoopBlock`, `rescheduledBasicBlocks` (−208) |
| `MetalStructuredControlFlow.java` | **new** — identical algorithm (Metal mirrors OpenCL) |
| `MetalBlockVisitor.java` | **deleted** (−576) |
| `MetalCompilationResultBuilder.java` | same edits as OpenCL (−208) |
| `CodeGenTest.java`, `TestFlashAttentionKernelContext.java` | control-flow reproducers from PR #846, added to the corpus |

**Net: ~1.5k lines of heuristics removed; ~0.8k lines of structured code added.**

The `if(cond)` / `for(;;)` / `switch` header text and the loop/switch opening `{` still come from
the existing LIR ops — unchanged. `elseClause` in the LIR builder was already dead code (never
assigned), so every `else` is structural.

---

### OpenCL — full unit-test suite (Apple M3 Pro)

A/B against the old path on identical graphs: **no deterministic regressions.** The few full-run
deltas vs. baseline are flaky/environment tests that flip on both paths (one even flips to
passing). Broad targeted sweep (~180 tests across arrays, conditionals, loops, loop-transforms,
reductions, math, vectors): **structured == baseline on every suite.**

### The decisive result — PR #846's own reproducers

| Test (PR #846 reproducer) | Old visitor (unpatched) | **Structured** |
|---|:--:|:--:|
| `CodeGenTest#testLoopConditionWithBreakInsideIf` (RayTracer shadowFeeler) | ❌ FAILED | ✅ PASS |
| `TestFlashAttentionKernelContext#testFlashAttention` | ❌ FAILED | ✅ PASS |

The bugs PR #846 hand-patches don't exist in the structured design — they're just what the
dominator-tree + loop-scope algorithm produces.

### Metal — `make fast-tests`

```
PASS = 679   FAIL = 10   UNSUPPORTED = 272
PASS / (PASS+FAIL)        = 98.55%
PASS / (PASS+FAIL+UNSUP)  = 70.66%   (UNSUPPORTED is mostly FP64 — Apple GPUs have no doubles)
```

All 10 failures are pre-existing device/precision/virtual-device tests (identical set to OpenCL,
on both old and new paths). All control-flow stress tests (`CodeGenTest`,
`TestFlashAttentionKernelContext`, loops, conditionals) pass.

---

### OpenCL (default backend)

```bash
make                 # build
source setvars.sh
make tests           # full suite
# or a single suite:
tornado-test --ea -V uk.ac.manchester.tornado.unittests.codegen.CodeGenTest
```

### Metal

```bash
make BACKEND=metal   # builds metal,opencl
source setvars.sh
make fast-tests
```

### Inspect generated kernels / debug the recovered structure

```bash
# Reproducers (fail with clBuildProgram -11 without the OCLBlockVisitor changes):
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestFlashAttentionKernelContext
tornado-test -V uk.ac.manchester.tornado.unittests.codegen.CodeGenTest

# Regression suites exercised during development:
tornado-test -V uk.ac.manchester.tornado.unittests.branching.TestConditionals
tornado-test -V uk.ac.manchester.tornado.unittests.branching.TestLoopConditions
tornado-test -V uk.ac.manchester.tornado.unittests.loops.TestLoops
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.reductions.TestReductionsFloatsKernelContext

# print the emitted kernel source:
tornado-test --ea --printKernel uk.ac.manchester.tornado.unittests.codegen.CodeGenTest#test05

# dump CFG facts the structurer uses (dominators, loop membership, successors):
tornado-test --ea -V \
  -J"-Dtornado.opencl.structuredcfg.debug=true" \
  uk.ac.manchester.tornado.unittests.codegen.CodeGenTest#test05
# (Metal: -Dtornado.metal.structuredcfg.debug=true)
```

> **Note:** `tornado-test` / `make fast-tests` need a Python with `psutil`. If you build with a
> venv that only has the build deps (`requests`), run the tests with the system / test venv
> instead, or `tornado-test` will fail at `import psutil`.

---


mikepapadim:
The job: Graal hands codegen a graph of basic blocks; OpenCL/Metal need nested text with
  matching { }. Something must turn one into the other.

  Old way: walk the blocks flat and, at each block, guess "open a brace? close one?" via ~20
  heuristics + a ledger tracking which braces you've already emitted. Guessing → mis-braces →
  new patch per kernel shape.

  New way — two phases:

  1. Structure → build a tree. Walk the CFG once and produce a nested AST:
  Seq · Block · IfElse(head, then, else?) · Loop(header, body) · Switch
  Nesting is decided by two rules:
  - Loops are scopes. Build the loop body within the loop. The first block reached that's not
  in the loop is a break/exit — emit it, stop. (So break/continue/early-return need zero
  special cases.)
  - An if re-merges at its dominator join — the block whose immediate dominator is the if and
  that has >1 predecessor. No such block ⇒ both branches exit, nothing after the if.

  -  (Joins come from the dominator tree, not getPostdominator(), because Graal leaves
  post-dominators null inside loops.)

  2. Emit → walk the tree. A stateless recursion:
  IfElse → emit head;  "{" then "}"  [ "else {" else "}" ]
  Loop   → emit header "{";  body;  "}"
  Each } is printed by returning from the recursion that printed its { — so they can't
  desync.

  Division of labor: the actual if(cond), for(;;), switch(x) text (and the loop/switch {)
  still come from the existing LIR ops via emitBlock. The new code only places if-branch {,
  else, case: labels, and every closing }.

  Why it's better: braces are correct by construction (it's a tree, not a counter), and new
  control-flow shapes are handled by the general algorithm instead of a new heuristic each
  time.

tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/graal/compiler/MetalStructuredControlFlow.java
```diff
@@ -0,0 +1,396 @@
+/*
+ * Copyright (c) 2024, APT Group, Department of Computer Science,
```
stratika:
```suggestion
 * Copyright (c) 2026, APT Group, Department of Computer Science,
```

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/OCLStructuredControlFlow.java
```diff
@@ -0,0 +1,396 @@
+/*
+ * Copyright (c) 2024, APT Group, Department of Computer Science,
```
stratika:
```suggestion
 * Copyright (c) 2026, APT Group, Department of Computer Science,
```

stratika:
thanks @mikepapadim, LGTM. Some comments:
- There is an issue in building SPIR-V, related to `OCLBlockVisitor`. 
- There are some virtual devices unit-tests failing. let's fix them to pass the CI tests. 
- It need to be backported to JDK25.

stratika (approved):
LGTM. the next step is to port the changes to JDK25.

PR #850, opened 2026-06-15
[Metal] Add hardware matrix units (`simdgroup_float8x8`) for fast gemms to M-silicon

mikepapadim:
## Summary

Adds first-class support for Apple's hardware matrix-multiply units (`simdgroup_float8x8`, the Apple-silicon equivalent of tensor cores) to the Metal backend, exposed through two `KernelContext` intrinsics:

| API | What it is | Reuse |
|-----|-----------|-------|
| `matrixMultiply8x8(a, aBase, lda, b, bBase, ldb, c, cBase, ldc, k)` | One 32-lane SIMD group cooperatively computes a single **8×8** output tile, streaming A/B straight from device memory. | none (per-tile) |
| `matrixMultiplyTiled(a, b, c, m, n, k)` | **Full** self-contained GEMM. Each threadgroup (128 threads / 4 SIMD groups) computes a **32×32** tile, staging A/B blocks in `threadgroup` memory and reusing them across a 4×4 grid of register fragments. | register **+** shared-memory |

The tiled variant is the data-reuse step that the per-tile intrinsic deliberately left as future work. On an **M3 Pro** it reaches **301 GFLOP/s** at 512³ — **1.55× faster than the per-tile MMA** and **1.72× faster than a naive kernel**.

The PR also carries the supporting Metal codegen/perf commits this work was built on (minimal per-kernel MSL preamble, opt-in fast-math and occupancy-hint compile options, and a SIMD-group reduction example).

---

## How the tiled GEMM works

Each **threadgroup** owns a 32×32 output tile of `C`. Its 128 threads form **four SIMD groups** laid out 2×2; each SIMD group owns a 16×16 region = a **2×2 grid of 8×8 register fragments**.

```
            C  (32 x 32 output tile per threadgroup)
        col0..7  col8..15  col16..23 col24..31
       +--------+--------+ +--------+--------+
row0..7 | acc00  | acc01  | |        |        |   SIMD group (0,0)   SIMD group (0,1)
row8..15| acc10  | acc11  | |        |        |   owns these 4      owns these 4
       +--------+--------+ +--------+--------+
       +--------+--------+ +--------+--------+
row16.. |        |        | |        |        |   SIMD group (1,0)   SIMD group (1,1)
row24.. |        |        | |        |        |
       +--------+--------+ +--------+--------+
```

Per step along K (BK = 8), the **whole threadgroup cooperatively** loads a 32×8 block of A and an 8×32 block of B into `threadgroup` memory **once**, then **every** SIMD group reuses those staged blocks:

```
  for kb in 0..K step 8:
      # 128 threads cooperatively stage A[32x8] and B[8x32] into threadgroup memory
      threadgroup_barrier
      # each SIMD group: load its 2 A-frags + 2 B-frags from threadgroup memory,
      #                  issue 4 simdgroup_multiply_accumulate into acc00..acc11
      threadgroup_barrier
  # write acc00..acc11 back with simdgroup_store
```

The accumulator fragments stay in registers for the whole K loop. Staging in `threadgroup` memory amortises device-memory traffic across the four SIMD groups, which is where the speed-up over the per-tile kernel comes from.

```
   device memory                 threadgroup memory            registers (per SIMD group)
   A[M][K], B[K][N]   ──load──▶  As[32][8], Bs[8][32]  ──MMA──▶ acc00 acc01 acc10 acc11
   (loaded once per                (reused by all 4               (held across K, then
    threadgroup K-step)             SIMD groups)                   simdgroup_store → C)
```

---

## Performance (Apple M3 Pro, `C[512×512] = A[512×512] × B[512×512]`)

| Kernel | Avg time | Throughput | vs naive |
|--------|---------:|-----------:|---------:|
| naive (one thread / output element) | 1.537 ms | 174.6 GFLOP/s | 1.00× |
| `simdgroup_float8x8` (per-tile) | 1.384 ms | 194.0 GFLOP/s | 1.11× |
| **`simdgroup tiled` (full MMA)** | **0.892 ms** | **301.0 GFLOP/s** | **1.72×** |

**Tiled vs per-tile MMA: 1.55×.** All three match the CPU reference.

> Context: the earlier opt-in `fast-math` and `[[max_total_threads_per_threadgroup]]` knobs were no-ops on bandwidth-bound kernels — compiler flags can't beat memory bandwidth. The tiled GEMM is the first change that moves the needle, because it actually reduces global-memory traffic through data reuse.

---

## How to run

Build the Metal backend and source the env:

```bash
make BACKEND=metal
source setvars.sh
tornado --devices        # confirm the Apple GPU is detected
```

### Example (naive vs per-tile vs tiled benchmark + correctness)

```bash
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixMultiplySimdgroup
```

Prints the correctness checks and the performance table above.

### Inspect the generated MSL

```bash
tornado-test --printKernel -V \
  uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestSimdgroupTiledMatrix#testSingleTile
```

### Unit tests

```bash
# tiled GEMM
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestSimdgroupTiledMatrix
# per-tile MMA
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestSimdgroupMatrix
```

### Minimal usage

```java
// The whole kernel body is one call.
static void gemm(KernelContext ctx, FloatArray a, FloatArray b, FloatArray c, int m, int n, int k) {
    ctx.matrixMultiplyTiled(a, b, c, m, n, k);
}

// Dispatch: one threadgroup (128 threads) per 32x32 output tile.
int numTiles = (m / 32) * (n / 32);
WorkerGrid1D grid = new WorkerGrid1D(numTiles * 128);
grid.setLocalWork(128, 1, 1);
```

---

## Constraints

- `m`, `n` must be multiples of **32**; `k` a multiple of **8**. Row-major matrices.
- Dispatch must be **1-D**, `(m/32)*(n/32)*128` threads, local size **128**.
- The kernel body must contain **only** the `matrixMultiplyTiled` call — the emitted block declares `threadgroup` arrays at the kernel's top scope, so it must not be nested inside a user `for`/`if`.
- **Metal only.** `simdgroup_matrix` has no OpenCL/PTX/SPIR-V equivalent; the tests skip those backends.

---

## Implementation notes

- `MetalSimdgroupTiledMatmulNode` (`FixedWithNextNode`, convergent) → `MetalLIRStmt.SimdgroupTiledMatmulStmt` emits the whole MSL block. Mirrors the per-tile `MetalSimdgroupMatmulNode`.
- Both MMA nodes implement **`MarkArrayParameterAccess`** (A/B `READ_ONLY`, C `WRITE_ONLY`). This is load-bearing: `TornadoDataflowAnalysis` infers a kernel parameter's access from the node *types* that consume it, so an opaque intrinsic taking arrays directly would otherwise leave C unmarked → the device→host copy is skipped → **results silently come back as zeros**. Any future intrinsic consuming array params directly must do the same.
- The tiled block is emitted **without wrapping braces** so its `threadgroup` arrays are function-scoped (see constraints).
- System indices are derived from the existing kernel signature (`_threadgroup_position_in_grid.x`, `_thread_position_in_threadgroup.x`); no new kernel parameters were added.

## Testing

- `TestSimdgroupTiledMatrix` (4 tests: single tile, deep-K, multi-tile, rectangular) — **pass**.
- `TestSimdgroupMatrix` (per-tile, 4 tests) — **pass**.
- `MatrixMultiplySimdgroup` example — all three kernels match the CPU reference.
- `make fast-tests`: **684 pass / 8 fail / 161 unsupported**, exit 0. The 8 failures are pre-existing/whitelisted (unrelated to this change); the +4 over the prior baseline are the new tiled tests. No regression.


stratika (approved):
LGTM, I tested it macOS.

PR #882, opened 2026-07-02
Fix repeated execution of batched plans

orionpapadakis:
#### Description

Fixes a crash when re-executing a batched (withBatch) plan on the PTX backend.
A batched plan's remainder chunk has a different thread-count than its full
chunks, so re-execution forces the interpreter to invalidate and recompile the
kernel; PTXCodeCache treated the invalidated (module-unloaded) entry as still
cached (installSource returned the dead module, and isCached made compileTask
skip codegen and build a null-target-code result), crashing in
cuModuleLoadDataEx. isCached/installSource are now validity-aware.

#### Problem description

Reproduce: run a withBatch(...) plan whose size is not a multiple of the batch
size (so there is a remainder chunk) more than once. On the 2nd execution it
fails with cuModuleGetFunction -> CUDA_ERROR_INVALID_HANDLE (400) / SIGSEGV.
The batch loop was only ever executed once by the suite, so this was uncovered;
now guarded by TestBatches#testBatchRepeatedExecution.
  
  
#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] CUDA
- [ ] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
$ make BACKEND=<opencl | ptx | cuda >

# Repeated execution of a batched plan (new regression test)
$ tornado-test -V uk.ac.manchester.tornado.unittests.batches.TestBatches#testBatchRepeatedExecution

# Full affected suites
$ tornado-test -V uk.ac.manchester.tornado.unittests.batches.TestBatches
```

### DIFF
--- a/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/PTXCodeCache.java
+++ b/tornado-drivers/ptx/src/main/java/uk/ac/manchester/tornado/drivers/ptx/PTXCodeCache.java
@@ -55,7 +55,13 @@ public class PTXCodeCache {
 
     public PTXInstalledCode installSource(TaskDataContext taskMeta, String name, byte[] targetCode, String resolvedMethodName, boolean debugKernel) {
 
-        if (!cache.containsKey(name)) {
+        // Recompile if the kernel is absent OR its cached module was invalidated (unloaded). The latter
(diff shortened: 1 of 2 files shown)


stratika (approved):
LGTM, thanks @orionpapadakis

PR #886, opened 2026-07-02
Fix batch processing with a remainder chunk under device buffer reuse

orionpapadakis:
#### Description

 This PR (complements #882) and fixes batch processing (`withBatch`) for plans whose data size is not a multiple of the batch size (i.e. plans with a remainder chunk) when device buffer reuse is enabled (`tornado.reuse.device.buffers=true`, the runtime  default). Three coordinated fixes, applied to all five backends (OpenCL,
  PTX, CUDA, SPIR-V, Metal):

  1. **Batch H2D copies honour the sub-region size** (`*MemorySegmentWrapper`
     / `*ArrayWrapper.enqueueWrite`). A batched device buffer can be larger
     than the chunk it currently holds: with buffer reuse the buffers are
     locked, so the deallocation before the smaller remainder chunk is a
     no-op and the remainder runs on the full-chunk buffer with only a
     sub-region in use. The write path copied the full `bufferSize`
     regardless, reading past the end of the host segment. The read path was
     already sub-region-aware; this makes the write path symmetric. The array
     wrappers gate this on `batchSize > 0`, because the lazy partial sync
     sets a read-only sub-region on non-batch buffers that must not shrink
     full-array writes.

  2. **The buffer-reuse path refreshes the sub-region size**
     (`*TornadoDevice.allocateObjects`). When `reuseBatchBuffer` decides the
     object's existing buffer can serve the current batch, `allocate()` — the
     place that normally refreshes the sub-region — is skipped, so a buffer
     kept across executions retained the previous execution's remainder
     sub-region and silently copied too few bytes.

  3. **The per-object chunk counters reset on every batched execution**
     (`TornadoVMInterpreter`). They were initialised once in the constructor
     and only ever grew, so from the second execution of a batched plan on,
     every per-chunk DEALLOC freed for real and the buffers were deallocated
     and reallocated on every chunk.

  #### Problem description

  Reproduce on `develop` (before this patch): run any `withBatch("<size>")`
  plan whose array size is not a multiple of the batch size, with default
  settings (buffer reuse enabled). The first `execute()` crashes the JVM
  (SIGSEGV in native driver code, e.g. `libcuda.so`) during the H2D transfer
  of the remainder chunk, which copies the full-chunk size past the end of
  the host segment.

  The unittest suite never caught this because `TestBatches.before()` sets
  `tornado.reuse.device.buffers=False`. The new
  `TestBatches#testBatchPipelinedRepeatedExecution`-style test
  (`testBatchWithRemainderAndBufferReuse`) enables buffer reuse for itself
  (restoring the property in `finally`), uses a 40MB batch over a 256MB array
  (6 full chunks + a smaller remainder), and repeats the execution with a
  per-index changing input to also cover locked buffers being reused across
  executions (fixes 2 and 3).

#### Backend/s tested

- [x] OpenCL
- [x] PTX
- [x] CUDA
- [x] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
  make BACKEND=ptx,opencl,spirv,cuda
  source setvars.sh

  # New regression test (crashes the JVM without this patch):
  tornado-test --ea uk.ac.manchester.tornado.unittests.batches.TestBatches#testBatchWithRemainderAndBufferReuse

  # Full batch suite per backend :
  tornado-test --ea -V --jvm="-Dtornado.unittests.device=<driver:device>"
  uk.ac.manchester.tornado.unittests.batches.TestBatches
  ```

### CODE REVIEW

**Comment by mikepapadim** [comment]
CI

**Review by stratika — APPROVED** [review]
LGTM, thanks @orionpapadakis

File: tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/batches/TestBatches.java [hunk 1]
```diff
@@ -275,6 +275,59 @@ public void testBatchRepeatedExecution() throws TornadoExecutionPlanException {
...
+     * ({@code cuMemcpyHtoDAsync} returned {@code CUDA_ERROR_INVALID_VALUE}) - so ANY batched plan
+     * with a remainder chunk failed under default settings, even on its first execution. Repeats the
+     * execution to also cover the locked buffers being reused across executions (a reused buffer must
+     * not keep the previous execution's remainder sub-region, and the per-object chunk counters must
+     * reset so per-chunk DEALLOCs stay no-ops).
+     */
+    @Test
+    public void testBatchWithRemainderAndBufferReuse() throws TornadoExecutionPlanException {
+        String reuseProperty = System.getProperty("tornado.reuse.device.buffers");
+        System.setProperty("tornado.reuse.device.buffers", "True");
```
stratika, line 293: Is this necessary? I think the default value of the property is True, as defined [here](https://github.com/beehive-lab/TornadoVM/blob/faaee452adda7acb4afcdcc5a8c64b768cfb5ba2/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/common/TornadoOptions.java#L393-L400).

stratika: After discussion with @orionpapadakis, we decided to keep `before()` in this PR which disables the `tornado.reuse.device.buffers` property as it triggers an issue regarding the lazy×batch. Hence, we need to explicitly enable it for the new unit-test. We will iterate in a future PR to unify the functionality for all batching unit-tests.

PR #889, opened 2026-07-03
[hybrid] CUDA-C backend: delegate @Reduce to CUB warp reductions

mikepapadim:
## Description

Delegates `@Reduce` reductions on the CUDA-C backend to **CUB** (header-only, NVRTC-compilable): ADD/MAX/MIN over int/long/float/double now generate a two-stage `cub::WarpReduce` composition instead of TornadoVM's local-memory reduction tree.

> Codegen-track integration (touches the CUDA-C backend compiler, not the library-task SPI). **Stacked on #887/#888** — this branch contains the merged hybrid-API stack. Review order: #887 → #888 → this.

### Design

- **`CUDACubReduceNode` + `CubWarpReduceStmt`**: an intrinsic node (modeled on `CUDASimdSumNode`) emitting a uniquely-suffixed sequence — per-warp `cub::WarpReduce`, lane-0 partials to shared memory, warp 0 reduces the partials. `cub::BlockReduce` was ruled out: it requires a compile-time block size, while TornadoVM's reduction block size is runtime; warp width (32) is compile-time by nature, and the `valid_items` overloads handle any block size and partial warps.
- **12 marker intrinsics** in `CUDAIntrinsics`, replaced in `TornadoCUDAIntrinsicsReplacements` — which already runs right after `HighTierLoweringPhase` inlines the reduction snippets, so the ordering was designed for exactly this.
- **24 `*Cub` snippet variants** in `ReduceGPUSnippets`, selected in the single `Templates.snippet()` funnel behind `-Dtornado.cuda.reduce.cub` (**default ON**; set `False` to restore the tree). MUL reductions keep the tree (no CUB product functor).
- Kernel preamble adds the **targeted** `#include <cub/warp/warp_reduce.cuh>` — the `<cub/cub.cuh>` umbrella is not NVRTC-compilable (it drags in host std headers); CUB ships with every CUDA toolkit ≥ 11, same include directory as the already-included `cuda_fp16.h`.

### Results (RTX 4090, CUDA 12.6)

- **Fixes a real bug**: `TestReductionsFloats#testComputePi` — a whitelisted known failure of the generated tree — now **passes** (13/13 vs 12/13 before).
- **Zero regressions**: TestReductionsIntegers 15/15, Doubles 17/17, Long 6/6, Automatic 4/4, Instance/Multiple 1/1 each — identical or better with CUB; TestArrays 23/23; hybrid-API suites unaffected (cuBLAS 11, cuBLASLt 5, cuFFT 8, cuDNN 9).
- **Faster**: `ReductionBenchmark` (new, in `tornado-examples`), 64M-float sum: tree 1.713 ms (157 GB/s) → **CUB 1.270 ms (211 GB/s), 1.35×**, both validated vs Java.
- Out of scope (unchanged): `TransformerKernelsTest#testReductionOneBlockWithLayer` — a hand-written KernelContext reduction, not an `@Reduce` snippet.

### Backend/s tested

CUDA backend (`make BACKEND=cuda`), RTX 4090. Other backends unaffected (all changes are inside `tornado-drivers/cuda`).


mikepapadim:
Full `make fast-tests` gate with CUB enabled by default: **completed, exit 0**. Remaining failures are exactly the pre-existing whitelisted set (`mxm2DKernelContext01/02`, `TestDevices#test04/06`, `testJuliaSets`, `testMandelbrot`, `testReductionOneBlockWithLayer`) — and `testComputePi` is no longer among them (fixed by this PR).

PR #894, opened 2026-07-04
 Remove dependency to JVMCI for JDK27+ backwards compatiblity for JDKs 21-26 

mikepapadim:
## Remove the dependency on JVMCI for JDK 27+ compatibility

JVMCI was removed from OpenJDK in JDK 27 (openjdk/jdk#30834). This PR makes TornadoVM run **without JVMCI** by sourcing all compiler metadata through a reflection + ASM + `Unsafe` layer, and makes that the **single, uniform path on every supported JDK (21, 25, 26, 27)** — there is no longer a HotSpot-JVMCI code path to maintain.

### What changed

**JVMCI's four roles, and their JDK-neutral replacements**
| JVMCI role | Replacement |
|---|---|
| Type / method / field / constant metadata for the GPU front-end | `runtime/jvmci/reflection/*` — `ReflectionUniverse`, `Reflection{ResolvedJavaType,Method,Field}`, `ReflectionConstantPool`, `ClassfileParser` … |
| Kernel-entry `Method` resolution from lambda proxies | `TaskUtils.resolveViaSerializedLambda` (SerializedLambda) |
| Host-side compile+install for hybrid `@Reduce` | Reflective tail reduction (`ReduceTaskGraph.runHostTailReduction`) |
| VM layout constants (headers, array offsets) | `TornadoVMConfigAccess` (Unsafe probing, compact-object-header aware) |

The Graal compiler is vendored as the relocated `tornado.graal` module; the `jdk.vm.ci.meta.*` **interfaces** it compiles against are supplied — frozen to the JDK-21 shape — on every JDK.

**Uniform JVMCI vendoring across JDKs.** The reflection providers implement the JDK-21 `jdk.vm.ci` interfaces. JDK 22–26 still ship a *drifted* platform `jdk.internal.vm.ci` (new abstract methods `isDeclared` / `getAllMethods` / `lookupConstant`), so those profiles overlay the frozen JDK-21 classes with `--patch-module jdk.internal.vm.ci=graalJars/jvmci-21.0.2.jar` at both compile and runtime. JDK 27 has no platform JVMCI, so the vendored module goes on the module path directly. JDK 21's platform module already matches, so it needs no patch.

**Architectural cleanup (this update):**
- Deleted the dual-path flags `USE_REFLECTION` / `USE_REFLECTION_FULL` — reflection is unconditional.
- Removed `HotSpotJVMCIRuntime` from the `TornadoBackendProvider.createBackend` SPI and from all backend providers/impls/factories.
- Deleted the dead host-compile machinery (`ReduceCompilationThread`, `CodeAnalysis.getGraalCompiler` / `compileAndInstallMethod`) and the removed-HotSpot-class reflective lookups in `TaskUtils`.
- Slimmed `TornadoMetaAccessProvider` / `TornadoConstantReflectionProvider` (no host backing provider, no A/B hook; unsupported queries throw directly).
- Added a `jdk25` build profile; removed the now-dead `-Dtornado.jvmci.reflection.full` launcher flag.

> Note: all `jdk22`–`jdk26` profiles compile at **`--release 22`** (class-file v66). The frozen Graal 23.1.0 classfile reader used by the `@Reduce` host-analysis path rejects newer class-file versions; v66 runs unchanged on newer JVMs.

### How to run

**Build + run the OpenCL backend on each JDK.** Set `JAVA_HOME` to the matching JDK, build, source the generated env, then run the test suite:

```bash
# JDK 21 (platform JVMCI == JDK-21 shape; no patch needed)
bin/compile --jdk jdk21 --backend opencl
source setvars.sh
tornado-test --verbose --quickPass

# JDK 25  (platform JVMCI overlaid with the frozen JDK-21 classes via --patch-module)
export JAVA_HOME=/path/to/jdk-25
bin/compile --jdk jdk25 --backend opencl
source setvars.sh
tornado-test --verbose --quickPass

# JDK 26  (same as JDK 25)
export JAVA_HOME=/path/to/jdk-26
bin/compile --jdk jdk26 --backend opencl
source setvars.sh
tornado-test --verbose --quickPass

# JDK 27  (no platform JVMCI; vendored jdk.internal.vm.ci module on the module path)
export JAVA_HOME=/path/to/jdk-27
bin/compile --jdk jdk27 --backend opencl
source setvars.sh
tornado-test --verbose --quickPass    # == `make fast-tests-jdk27`
```

The launcher (`tornado`) selects the right module wiring automatically from the running JDK version (vendored module on 27+, `--patch-module` on 22–26).

> **Use `--quickPass`, not `make fast-tests`, on JDK ≥ 18.** `make fast-tests` runs with `-ea`, and the vendored Graal 23.1.0 registers an invocation plugin for `Unsafe.getObject` (renamed `getReference` in JDK 18), whose `checkResolvable` assertion aborts backend init. `make fast-tests-jdk27` already drops `-ea`; the `--quickPass` invocation above is the assertion-free equivalent for every JDK.

### Validation

OpenCL backend, `tornado-test --verbose --quickPass` (assertions off):

| JDK | PASS / FAIL | Notes |
|-----|-------------|-------|
| 21 | 760 / 20 | all failures pre-existing / out-of-scope (device FP, `.ftz` half-float, matrices, atomics-18, mandelbrot, julia) |
| 26 | 760 / 20 | same set |
| 27 | 760 / 20 | same set |

JDK 25: build profile added and wired; not yet validated in CI (no JDK 25 runner locally) but uses the identical JDK 22–26 `--patch-module` mechanism that JDK 26 exercises.

### Trade-offs
- The in-tree A/B comparison against HotSpot providers is gone (recoverable from history).
- JDK 21/26 move from the JVMCI path onto the reflection stack, surfacing the same out-of-scope device/FP failures JDK 27 already had — no new in-scope regressions.
- PTX/CUDA/Metal/SPIRV are updated for **compile compatibility only**; their runtime is out of scope for this PR (OpenCL is the validated backend).


mikepapadim:
## ✅ Validation matrix — post-JVMCI reflection path

Pass rate = `passed / (ran − unsupported)` from `tornado-test --quickPass` (unsupported = feature-gated skips: FP16/FP64/MMA/multi-device, not failures). CUDA + OpenCL numbers below were run locally on an NVIDIA RTX 4090.

| Backend | JDK 21 | JDK 25 | JDK 27 |
|---------|--------|--------|--------|
| **CUDA**   | ✅ 99.0% (8 fail) | ✅ 99.2% (7 fail) | ✅ 99.2% (7 fail) |
| **OpenCL** | ✅ 98.7% (10 fail)³ | ✅ 98.7% | ✅ 98.7% |
| **Metal**  | 🍎 macOS runner² | 🍎 macOS runner² | 🍎 macOS runner² |

¹ "run" excludes unsupported (CUDA 913 total − 85 unsupported ≈ 828 effective; OpenCL 916 − 136 ≈ 780).
² Metal is macOS-only; validated on the self-hosted mac runner (`build-test-jdk21.yml` → `build-macOs-metal`).

### Extras
- **GPULlama3** — tested end-to-end ✅ (maintainer run)
- **Hybrid API** (cuBLAS/cuFFT/cuDNN) — tested ✅ (maintainer run)
- **CI** — added a no-GPU unit-test stage (GitHub-hosted, `unit-tests-reflection.yml`); reflection-layer unit tests **23/23** on JDK 21 + 27.

³ OpenCL reflection-path fixes landed this round took it from 21 → 10 real fails (~3 of the 10 are multi-device/virtual tests that need >1 accelerator): `ctx.atomicAdd` recovery (`AtomAddNodeTemplate`, OpenCL had none), HalfFloat device-function params typed `half` not `short` (getFloat32 truncated 1.5→1), and local HALF array reads/writes unproxified past the reflection-path PiNode. Mirrors existing CUDA-backend fixes. Greens atomics, quantized matrix-vector, and the HalfFloat local-array/reduction tests.

### Fixes landed while validating
- **jdk25/jdk26 CUDA build** was broken: the cpp-only `cublas/cufft/cudnn-jni` modules NPE'd in `maven-compiler-plugin` (`getPathElements() null`) under those profiles' `patch-module` args. Fixed by skipping the empty Java compile on those modules (`maven.main.skip`).
- **quickPass reporting** counted unsupported tests as failures; now split into a separate `Unsupported` column (matches the verbose runner), so feature-gated skips no longer read as regressions.

_Full quickPass baseline is behavior-neutral vs the pre-cleanup branch tip (verified identical fail/timing)._

mikepapadim:
## ⏱️ Perf: JVMCI (master) vs reflection path (this branch)

**Setup:** JDK 21, CUDA backend, NVIDIA RTX 4090. Kernel = row-major matrix-vector multiply (KernelContext local-memory reduction), `8192 × 2048`. Metric = **end-to-end** wall time of `plan.execute()` (Java dispatch + `transferToHost` copy + kernel launch + sync), not kernel-only. 3 runs each, `System.nanoTime`.

| | master (JVMCI) | this branch (reflection, no JVMCI) | verdict |
|---|---|---|---|
| **COLD** — first exec (TornadoVM sketch + graph-build + device compile) | ~148 ms | **~114 ms** | branch **~23 % faster** |
| **WARM** — fully-warmed steady exec (5k warmup) | ~0.091 ms | **~0.093 ms** | **parity** (±3 %, noise) |

### Notes
- **No steady-state regression.** Once warm, the reflection path runs at master's speed. An earlier under-warmed measurement (150 iters) showed a false ~2× slowdown — the reflection-metadata classes carry more bytecode, so HotSpot needs more JIT iterations to reach steady state (~2–5k vs ~150). Warm-up ramp on the branch: `0.120 → 0.112 → 0.105 → 0.093 ms` at warmup `50 → 200 → 1k → 5k`.
- **COLD is faster on the branch**, thanks to the reflection-layer caching in this PR (classfile parsed once, memoized signature/toString, cached member/CP/field lookups). So first-compile is *not* a reflection-path penalty here — it beats HotSpot JVMCI on this kernel.
- **GC ruled out** — 0 GC pauses across 650+ iterations.

### Takeaway
Removing JVMCI (reflection metadata layer) costs **nothing at steady state** and **wins on cold compile** for this workload; the only trade-off is a slightly longer JIT warm-up ramp, inherent to the larger metadata codebase.

<sub>Measured with a small harness driving `MatrixVectorRowMajor`'s core kernel with an explicit COLD/WARM split (the example's built-in timers only report steady-state).</sub>

mikepapadim:
## 📌 Branch state — post-JVMCI reflection path

Snapshot at `43d195056`. TornadoVM runs without JVMCI on the reflection metadata layer; validated at parity with the JVMCI path.

### Backend status
| Backend | State |
|---|---|
| **CUDA** | ✅ validated — quickPass 913 run / 7 real fail / 85 unsupported |
| **OpenCL** | ✅ validated — quickPass 916 run / 10 real fail / 138 unsupported |
| **Metal** | ⏳ **remaining** — needs a quickPass on a macOS runner |
| PTX / SPIRV | — being removed next release; not validated on purpose |

(Real fails are pre-existing port-tail items; the quickPass reporting fix below stopped counting *unsupported* feature-gated tests as failures.)

### Fixes landed on this branch (all backend-verified)
- **atomics** — OpenCL `ctx.atomicAdd` reflection-path recovery (`AtomAddNodeTemplate`); OCL had none.
- **HalfFloat** — device-function params typed `half` (not `short`, was truncating `getFloat32`); local HALF array read/write unproxified past the reflection-path PiNode. Greens quantization + the HalfFloat cluster.
- **Q8 ByteArray** — intrinsify `ByteArray.getHalfFloat/setHalfFloat` (OCL + CUDA) so the sketcher never descends into the abstract `MemorySegment.getShortAtIndex/getAtIndex`; unblocks the Q8_0 quantized matmul.
- **jdk25/jdk26 CUDA build** — skip the empty java compile on the cpp-only cublas/cufft/cudnn JNI modules (mvn-compiler `getPathElements()` NPE under `--patch-module`).
- **quickPass reporting** — split `Unsupported` from `Failed` in the terse runner (feature-gated skips were read as regressions).
- **Reflection-layer caching, correctness hardening, 23 pure-JVM unit tests, no-GPU CI stage** (WP-A).
- **Launch** — `--add-modules ALL-SYSTEM` → explicit list (unblocks Leyden AOTCache; `LEYDEN-next-steps.md`).

### Perf (jdk21, CUDA, matvec row-major, master JVMCI vs this branch)
- **COLD** (first exec incl. compile): ~148 ms → **~114 ms** (−23%, reflection-layer caching).
- **WARM** (fully warmed steady): **parity** (~0.09 ms both). No steady-state regression; branch just needs more JIT-warmup iterations (larger metadata codebase). GC ruled out.

### Leyden / startup
- AOTCache works via `tornado --jvm="-XX:AOTCacheOutput=x.aot"` → ~18% faster COLD (partial). Full-module-graph archive blocked by `jdk.incubator.vector` (incubator → hard-disables full-graph CDS); core TornadoVM doesn't require it, so non-vector apps get the full cache. See `LEYDEN-next-steps.md`.
- **`java @argfile` launches TornadoVM** — confirmed, given a *flat* argfile (exportLists inlined; Java doesn't expand nested `@`-files) with the jvmci dir on the module-path. Note: the shipped `gen-tornado-argfile-template.py` is incomplete for the reflection path (omits the jvmci module-path + inlining) — a small follow-up.

### Known-remaining (not blockers)
- OpenCL `TestHalfFloats#testMatrixVectorHalfFloatOptimized` — NVIDIA-OpenCL ptxas `.ftz`-on-half-convert quirk (platform-specific).
- ~3 multi-device/virtual tests need >1 accelerator.

mikepapadim:
## 🍎 Metal — validated (Apple M3 Pro, macOS aarch64)

Filling in the Metal row of the validation matrix. All numbers below run locally on an **Apple M3 Pro**.

### `tornado-test --quickPass` (JDK 21, Metal + OpenCL dist)

**913 run · 13 real fail · 259 unsupported → 98.0% pass** (`passed / (ran − unsupported)` = 641 / 654).

This is up from an earlier **78 real fails** — the jump came from fixing the HalfFloat store readback (mark `WriteHalfFloatNode` as an array write so the output gets a `TRANSFER_DEVICE_TO_HOST`, mirroring the OpenCL `bd981f4ca`). The whole HalfFloat / quantization / transformer family (~24 tests) is now green.

**The residual 13 contain zero Metal-specific / JVMCI-reflection regressions:**

| Bucket | Count | Detail |
|---|---|---|
| Test-data bug (out of scope) | 1 | `foundation.TestHalfFloats#testMatrixVectorHalfFloatOptimized` — `fillRandomDataFp16` uses `random.nextInt() * range` (~2e9) → fp16 `Inf` → the **sequential reference** is `NaN` (`expected:<NaN> but was:<0.0>`); fails on any backend. |
| Env / enum flakes | 7 | `api.TestDevices#test04/05/06`, `virtual.TestVirtualDeviceKernel` (device-enum / prebuilt-desc); `TestWarmUp` / `TestConcurrentBackends` / `atomics.TestAtomics` **pass in isolation** (the TestAtomics quickPass fail is an OpenCL `cl2Metal failed` / `clBuildProgram err:-2` on the *OpenCL* device, not Metal). |
| Reproducible correctness — **confirmed OpenCL-parity** | 5 | `kernelcontext.matrices.TestMatrixMultiplicationKernelContext#mxm2DKernelContext01/02`, `compute.ComputeTests#testMandelbrot` + `#testJuliaSets`, `compute.TransformerKernelsTest#testReductionOneBlockWithLayer`. **Re-ran each on the OpenCL device (`-Dtornado.unittests.device=0:0`) in the same dist: all fail identically.** Shared reflection-path / pre-existing reduction+compute bugs, not a Metal regression. |

So **Metal now matches OpenCL's residual failure set** — JVMCI removal on Metal is functionally complete. Every reflection-path gap is landed: accessor intrinsics, HotSpot-cast removal, KernelContext recovery (`atomicAdd`, `allocate*LocalArray`, `local/globalBarrier`), HalfFloat device-fn param typing, local-half read (PiNode unproxify), `ByteArray.getHalfFloat/setHalfFloat`, and the HalfFloat writeback.

### Multi-JDK build + run — **JDK 21 / 25 / 26 / 27 all ✅**

`bin/compile --jdk jdk<NN> --backend metal,opencl` then `arrays.TestArrays` on the Metal device:

| JDK | Build | `arrays.TestArrays` (Metal) | JVMCI |
|---|---|---|---|
| **21** | ✅ | 23 pass / 0 fail / 5 unsup | native (patch-module) |
| **25** | ✅ | 23 pass / 0 fail / 5 unsup | native (patch-module) |
| **26** | ✅ | 23 pass / 0 fail / 5 unsup | native (patch-module) |
| **27** | ✅ | 23 pass / 0 fail / 5 unsup | **none — reflection path** (vendored module) |

**Build fix for jdk25/jdk26:** `tornado-drivers/metal-jni` is cpp-only (zero Java sources); its empty `maven-compiler-plugin:default-compile` NPEs with `getPathElements() null` under those profiles' patch-module compilerArgs (mvn-compiler 3.11.0). Fixed with `<maven.main.skip>true</maven.main.skip>`, mirroring the cublas/cufft/cudnn fix (`0ceef1243`). jdk21/jdk27 were unaffected.

### Updated matrix

| Backend | JDK 21 | JDK 25 | JDK 26 | JDK 27 |
|---------|--------|--------|--------|--------|
| **CUDA**   | ✅ 99.0% | ✅ 99.2% | — | ✅ 99.2% |
| **OpenCL** | ✅ 98.7% | ✅ 98.7% | — | ✅ 98.7% |
| **Metal**  | ✅ 98.0% (13 fail) | ✅ build+run | ✅ build+run | ✅ build+run |

_Metal quickPass measured on JDK 21; JDK 25/26/27 confirmed via build + `arrays.TestArrays` on the Metal device. Pushed to `mikepapadim/jdk27-jvmci-removal` (`cf476a8e1`)._

mikepapadim:
## 🏁 Concluding assessment — JVMCI removal, complete work

> Context: the automated Copilot review bailed (PR exceeds 300 files), so this is a structured closing assessment of the full branch. Scope: **947 files, +17,419 / −8,809, 84 commits** vs `master`.

### What this PR delivers

**1. A JDK-neutral metadata layer replacing HotSpot JVMCI** (`tornado-runtime/.../jvmci/reflection/`, ~14 new classes + unit tests): `ReflectionResolvedJavaMethod/Type/Field`, `ReflectionConstantPool`, `ReflectionUniverse`, `ClassfileParser`, `ReflectionMemoryAccessProvider`, plus `TornadoConstantReflectionProvider` / `TornadoSnippetReflectionProvider`. Graal compiles kernels against plain `java.lang.reflect` + class-file parsing instead of `HotSpotJVMCIRuntime` / `CompilerToVM`.

**2. All three shipping backends ported and validated at parity.** The systematic failure mode on the reflection path (Graal's `InvocationPlugins.lookupInvocation` missing at sketch time → intrinsics surviving as plain invokes → crashes / truncations / missing readbacks) was fixed with the same recipe per backend: recover surviving invokes in `*IntrinsicsReplacements`, register native-array / HalfFloat / ByteArray accessor intrinsics in `*GraphBuilderPlugins`, remove `HotSpotResolvedJavaType` casts, rewrite `*FieldBuffer` marshalling to `Unsafe.objectFieldOffset`, and fix write-detection so outputs emit `TRANSFER_DEVICE_TO_HOST`.

**3. Multi-JDK support with one codebase.** JDK 21/25/26 run with the vendored jvmci module patched in; JDK 27 (no JVMCI in the platform at all) runs fully on the reflection path. Build profiles, launcher (`tornado.py`), argfile generation, and the cpp-only JNI module fixes (`cublas/cufft/cudnn/metal-jni` `maven.main.skip`) are all in place.

**4. CI for the reflection layer** — `unit-tests-reflection.yml` runs the metadata-layer unit tests on a GitHub-hosted runner with no accelerator (first such job in TornadoVM).

### Validation matrix (final)

| Backend | JDK 21 | JDK 25 | JDK 26 | JDK 27 (no JVMCI) |
|---------|--------|--------|--------|-------------------|
| **CUDA** (RTX 4090)  | ✅ 99.0% | ✅ 99.2% | — | ✅ 99.2% |
| **OpenCL** (RTX 4090) | ✅ 98.7% | ✅ 98.7% | — | ✅ 98.7% |
| **Metal** (M3 Pro) | ✅ 98.0% (13 fail / 913) | ✅ build+run | ✅ build+run | ✅ build+run |

Pass rate = `passed / (ran − unsupported)`. Every backend's residual failures were individually triaged; **none are reflection-path regressions** — Metal's 13 are 1 test-data bug (`NaN` reference), 7 env/ordering flakes that pass in isolation, and 5 correctness bugs that reproduce **identically on OpenCL** (shared, pre-existing).

**Performance:** end-to-end parity when warm (±3%, RTX 4090, matvec 8192×2048); **~23% faster cold start** (no JVMCI runtime init). GPULlama3 and the Hybrid API (cuBLAS/cuFFT/cuDNN) validated end-to-end by maintainer runs.

### Honest limitations (known, documented in `JDK27-CHECKLIST.md`)

- **`@Reduce` on JDK 27** — the host-side reduction detector still uses Graal-as-JIT via HotSpot JVMCI; guarded off on the pure reflection path (§4 of the checklist). Works on JDK 21/25/26. This is the one deferred feature, an architectural follow-up, not a porting gap.
- **`-ea` on JDK 27** — vendored Graal 23.1.0 registers an `Unsafe.getObject` plugin (renamed `getReference` in modern JDKs); assertion-enabled runs trip it. Suites run without `-ea`.
- **PTX / SPIRV** — intentionally not ported (scheduled for removal next release).
- Cross-backend correctness tail (~5 tests: KernelContext matmul reduction, Mandelbrot/Julia, transformer reduction) — fails on all backends alike; separate workstream.

### Bottom line

The branch does what its title says: **TornadoVM no longer needs JVMCI**, from JDK 21 through JDK 27, across CUDA / OpenCL / Metal, at test-suite and performance parity with master — with the deltas triaged to named, pre-existing causes rather than regressions. The reflection layer is unit-tested and exercised in CI without a GPU. Remaining items are tracked and scoped (`@Reduce` host-graph rewrite, packaging polish, `-ea` Unsafe rename).

From my side this work is **complete and ready for maintainer review/merge decision**. Given the 947-file footprint, the most reviewable units are: (1) the `runtime/jvmci/reflection/` package, (2) one backend's factory + `IntrinsicsReplacements` + `GraphBuilderPlugins` trio (OpenCL is the reference; CUDA/Metal mirror it), and (3) the build tooling (`build_jvmci_module.py`, profiles, launcher).

mikepapadim:
## 📦 State update — packaging verified, `@Reduce` confirmed complete

Two follow-ups from the closing assessment, both now resolved and pushed (`bf3c91f15`…`8950c3cb1`):

### 1. Packaging / reproducibility (checklist §5) — ✅ complete

A clean `bin/compile --jdk jdk27` now provably produces a runnable SDK with **zero manual fixups**. Verified by a clean-room test on macOS / JDK 27.ea.24:

1. Deleted every regenerable jvmci artifact: `graalJars/jvmci-21.0.2.jar`, `~/.m2/repository/tornado/jvmci`, and the whole `dist/`.
2. Ran one `bin/compile --jdk jdk27 --backend metal,opencl`.
3. The build regenerated the vendored `jdk.internal.vm.ci` module from the local JDK 21 image, staged it to `graalJars/`, installed `tornado.jvmci:jvmci` to the local Maven repo, and the assembly shipped it to `share/java/jvmci/`.
4. Fresh dist works immediately: `tornado --devices` shows Metal + OpenCL, `arrays.TestArrays` = **23 pass / 0 fail** on the Metal device.

The full chain was already wired (`bin/compile` → `pull_graal_jars.main(jdk)` → `build_jvmci_module.build()` → `assembly.xml` → `tornado.py` module-path/patch-module logic) — it had just never been clean-room verified. One small gap fixed: `bin/pull_graal_jars.py` now takes a `--jdk` CLI arg so standalone invocations also stage the jvmci module (previously only the `bin/compile` code path did).

### 2. `@Reduce` on JDK 27 (checklist §4) — ✅ already complete, now re-verified

The checklist still listed `@Reduce` as deferred ("host-side Graal-as-JIT via `HotSpotJVMCIRuntime`"), but that was stale: commits `1e2f8e691` (build the analysis high-level graph via the backend reflection providers; accept `InvokeWithExceptionNode` in reduce detection) and `dd208420e` (delete the HotSpot JVMCI dual-path entirely; the host combiner is always the reflective tail reduction) already landed the JDK-neutral path.

Re-verified on the clean-room **JDK 27** Metal dist — the whole reductions family is green:

| Suite | Result |
|---|---|
| `TestReductionsIntegers` | 15 / 0 |
| `TestReductionsFloats` | 13 / 0 |
| `TestReductionsLong` | 6 / 0 |
| `TestReductionsAutomatic` | 4 / 0 |
| `InstanceReduction`, `MultipleReductions` | 1 / 0 each |
| `TestReductionsDoubles` | 16 unsupported (FP64 device gating, expected) |

So the one "deferred feature" in the closing assessment is not deferred: **`@Reduce` works on JDK 27 without JVMCI.** There are no remaining code paths that reach `HotSpotJVMCIRuntime`/`CompilerToVM` at runtime.

### In flight

A full `tornado-test --quickPass` on the clean-room **JDK 27** Metal dist is running to upgrade the Metal × JDK 27 matrix cell from "build + smoke test" to a full-suite number (JDK 21 Metal baseline: 13 real fails / 913). Will follow up with the result.

mikepapadim:
## ✅ Follow-up: JDK 27 Metal full quickPass — parity confirmed

The in-flight run finished. Full `tornado-test --quickPass` on the **clean-room JDK 27.ea.24 dist** (no JVMCI anywhere — pure reflection path), Apple M3 Pro:

**913 ran / 12 real fails / 259 unsupported → 98.2% pass** — one fewer failure than the JDK 21 baseline (13), with an **identical failing-class profile**: same test-data bug (`testMatrixVectorHalfFloatOptimized` NaN reference), same device-enum items (`TestDevices`, `TestVirtualDeviceKernel`), same 5 shared cross-backend correctness tests (fail identically on OpenCL), minus one JDK 21 ordering flake (`TestWarmUp`).

### Final validation matrix

| Backend | JDK 21 | JDK 25 | JDK 26 | JDK 27 (no JVMCI) |
|---------|--------|--------|--------|-------------------|
| **CUDA** (RTX 4090)  | ✅ 99.0% | ✅ 99.2% | — | ✅ 99.2% |
| **OpenCL** (RTX 4090) | ✅ 98.7% | ✅ 98.7% | — | ✅ 98.7% |
| **Metal** (M3 Pro) | ✅ 98.0% (13 fail / 913) | ✅ build+run | ✅ build+run | ✅ **98.2% (12 fail / 913)** |

Every backend × JDK cell that runs a full suite lands within noise of its JVMCI-era baseline, and the JDK 27 no-JVMCI cell on Metal is now a full-suite number, not just a smoke test. Combined with `@Reduce` re-verified green on this same dist (previous comment), there is no remaining functional gap on the reflection path.

Pushed as `6c4525e47`.

mikepapadim:
## 🧪 Reflection-layer unit tests extended — 23 → 44, all green

The standalone reflection metadata API (`tornado-runtime/.../jvmci/reflection/`) now has three new pure-JVM (no-GPU) test suites covering the classes that previously had none (`ca9fdfbff`). All run in the existing `unit-tests-reflection.yml` CI job on a GitHub-hosted runner.

### What's newly covered and why it matters

**`ReflectionResolvedJavaFieldTest` (7 tests)** — the field metadata every marshaller and field lowering depends on:
- the reported offset **is** the `Unsafe` field offset (instance and static) — the exact property that keeps the host-side object marshaller (`OCLFieldBuffer`/`MetalFieldBuffer`) and the kernel's field-access codegen writing/reading the same displacement;
- RUNTIME annotations are visible through the SPI (the `@SegmentElementSize` guard in `TornadoNativeTypeElimination` reads these), and a missing annotation returns `null`, not a throw;
- universe canonicalization: the same `Field` always resolves to the **same** `ResolvedJavaField` instance (Graal keys maps on it).

**`ReflectionMembersTest` (8 tests)** — the constant-pool-descriptor → reflective-member resolver, which must follow JVM resolution rules because `ReflectionConstantPool` feeds it raw `Methodref`/`Fieldref` triples during graph building (a wrong walk order surfaces as a sketch-time unresolved invoke):
- declared method by exact descriptor; inherited method through the superclass chain; interface **default** method after the superclass chain; `<init>` → constructor;
- signature-polymorphic lookup by name only (`MethodHandle.invokeExact`);
- a miss returns `null` (wrong name **or** wrong descriptor), never throws;
- the resolution cache hands back the identical `Executable` so canonicalization holds upstream.

**`ReflectionUniverseTest` (6 tests)** — the shared helpers and identity guarantees:
- JVM type descriptors in both directions (`int → I`, `double[][] → [[D`, `String[] → [Ljava/lang/String;`) with a full round-trip over all primitives/arrays/objects, and unknown descriptors throwing `ClassNotFoundException`;
- classfile bytes are real (`0xCAFEBABE`) and cached — one read per declaring class;
- type lookups are canonical per-universe and null-safe;
- **signature-polymorphic canonicalization is per call-site descriptor**: same `(method, "(I)I")` → same instance, `"(J)J"` → different instance, and the signature reflects the call-site descriptor rather than the declared `(Object...)` shape — the exact `InvocationPlugins`-keying guarantee the sketcher relies on.

Plus a new fixture hierarchy (`ReflectionFieldSample` + base class + interface) exercising annotated fields, inherited members, and default methods against real compiled bytecode.

### Verification
- `./mvnw -Pjdk27 -pl tornado-runtime test -DskipTests=false`: **Tests run: 44, Failures: 0, Errors: 0** (JDK 27.ea.24; suite previously 23).
- `checkstyle:check` clean.

Together with the existing parser/method/type/signature suites, every public class of the reflection SPI now has direct unit coverage except the two thin delegates (`ReflectionProfilingInfo` — static no-profile stubs; `ClassfileResolvedJavaField` — exercised transitively via `ClassfileParserTest`).

tornado-annotation/src/main/java/uk/ac/manchester/tornado/annotation/ASMClassVisitor.java
```diff
@@ -55,17 +55,29 @@ public MethodVisitor visitMethod(int access, String name, String descriptor, Str
         return null;
     }
 
+    // ASM's ClassReader only knows class-file versions up to the ASM release it ships with (asm 9.7 = Java 23,
+    // major 67); a newer JDK class on the class path (e.g. java.lang.Float is v71 on JDK 27) makes it throw.
+    // @Parallel-annotation extraction is class-file-version independent, so cap the major version for the read.
+    private static final int MAX_SUPPORTED_MAJOR = 67;
```
stratika: I think, this should be higher since we will support up to JDK27.

tornado-assembly/src/bin/tornado-test
```diff
@@ -322,7 +322,6 @@ __TORNADO_TESTS_WHITE_LIST__ = [
 
     ## Precision errors
     "uk.ac.manchester.tornado.unittests.compute.ComputeTests#testNBodyBigNoWorker",
-    "uk.ac.manchester.tornado.unittests.compute.MMwithBytes#testMatrixMultiplicationWithBytes",
```
stratika, line 355: why is this test removed?

tornado-assembly/src/bin/tornado-test
```diff
@@ -351,7 +351,6 @@ __TORNADO_TESTS_WHITE_LIST__ = [
...
```
stratika, line 355: why is this test removed?

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/foundation/TestHalfFloats.java
```diff
@@ -107,7 +107,7 @@ public static float matrixVectorRowMajorOptimized(KernelContext context, int loc
 
 
         // Store partial sum in local memory
-        localSum[localId] = partialSum.getHalfFloatValue();
+        localSum[localId] = partialSum.getFloat32();
```
stratika, line 110: I think we should revert this.

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/WriteHalfFloatNode.java
```diff
@@ -21,25 +21,26 @@
...
+import tornado.graal.compiler.nodes.spi.LIRLowerable;
+import tornado.graal.compiler.nodes.spi.NodeLIRBuilderTool;
 
 import jdk.vm.ci.meta.Value;
 import uk.ac.manchester.tornado.drivers.cuda.graal.HalfFloatStamp;
 import uk.ac.manchester.tornado.drivers.cuda.graal.CUDAArchitecture;
 import uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDAKind;
 import uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDALIRStmt;
 import uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDAUnary;
+import uk.ac.manchester.tornado.runtime.graal.nodes.interfaces.MarkOCLWriteNode;
```
stratika: the `MarkOCLWriteNode` is used across all backends. We should rename it to be generic.

tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/graal/nodes/WriteHalfFloatNode.java
```diff
@@ -23,25 +23,29 @@
...
+import tornado.graal.compiler.nodes.spi.LIRLowerable;
+import tornado.graal.compiler.nodes.spi.NodeLIRBuilderTool;
 
 import jdk.vm.ci.meta.Value;
 import uk.ac.manchester.tornado.drivers.metal.graal.HalfFloatStamp;
 import uk.ac.manchester.tornado.drivers.metal.graal.MetalArchitecture;
 import uk.ac.manchester.tornado.drivers.metal.graal.lir.MetalKind;
 import uk.ac.manchester.tornado.drivers.metal.graal.lir.MetalLIRStmt;
 import uk.ac.manchester.tornado.drivers.metal.graal.lir.MetalUnary;
+import uk.ac.manchester.tornado.runtime.graal.nodes.interfaces.MarkOCLWriteNode;
```
stratika: same here regarding renaming to be generic to the backend type.

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/plugins/OCLHalfFloatPlugins.java
```diff
@@ -111,13 +112,37 @@ public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Rec
...
+            @Override
+            public boolean apply(GraphBuilderContext b, ResolvedJavaMethod targetMethod, Receiver receiver) {
+                HalfFloatPlaceholder placeholder = new HalfFloatPlaceholder(receiver.get(false));
+                b.getGraph().addOrUnique(placeholder);
+                OCLConvertHalfToFloat convertHalfToFloat = new OCLConvertHalfToFloat(placeholder);
+                b.getGraph().addOrUnique(convertHalfToFloat);
+                b.push(JavaKind.Float, convertHalfToFloat);
+                return true;
+            }
+        });
```
stratika, line 192: why this is not necessary for JDK21 and `develop`?

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/WriteHalfFloatNode.java
```diff
@@ -21,25 +21,26 @@
...
 import uk.ac.manchester.tornado.drivers.opencl.graal.HalfFloatStamp;
 import uk.ac.manchester.tornado.drivers.opencl.graal.OCLArchitecture;
 import uk.ac.manchester.tornado.drivers.opencl.graal.lir.OCLKind;
 import uk.ac.manchester.tornado.drivers.opencl.graal.lir.OCLLIRStmt;
 import uk.ac.manchester.tornado.drivers.opencl.graal.lir.OCLUnary;
+import uk.ac.manchester.tornado.runtime.graal.nodes.interfaces.MarkOCLWriteNode;
 
 @NodeInfo
-public class WriteHalfFloatNode extends FixedWithNextNode implements LIRLowerable {
+public class WriteHalfFloatNode extends FixedWithNextNode implements LIRLowerable, MarkOCLWriteNode {
```
stratika: this interface to be renamed in a more generic name to cover all backends

docs/source/installation.rst
```diff
@@ -19,10 +19,12 @@ Via SDKMAN!:
    # e.g.:
    sdk install tornadovm 5.0.0-jdk21-cuda
    sdk install tornadovm 5.0.0-jdk25-metal
+   sdk install tornadovm 5.0.0-jdk26-opencl
+   sdk install tornadovm 5.0.0-jdk27-opencl
```
stratika: remove these

.gitignore
```diff
@@ -1,5 +1,6 @@
 .idea/
 .unittestingStatus
+.flattened-pom.xml
```
stratika, line 3: what is this pom file?

docs/source/faq.rst
```diff
@@ -39,7 +39,7 @@ TornadoVM currently supports:
 
 -  JDK 21 (e.g. OpenJDK 21, Red Hat Mandrel 21, Amazon Corretto 21, Azul Zulu JDK 21).
 -  JDK 25.
--  GraalVM (JDK 21), for polyglot interoperability (see :ref:`truffle-languages`).
+-  GraalVM (JDK 21), as an alternative JDK 21 distribution (the ``graal-jdk-21`` build profile).
```
stratika, line 42: I think this is not necessary any more.

stratika (approved):

tornado-annotation/src/main/java/uk/ac/manchester/tornado/annotation/ASMClassVisitor.java
mairooni: That's the ceiling of what ASM 9.7 can parse, not the JDKs we support. The code downgrades newer class files to it, so raising it to 71 would make ASM reject JDK 27 classes. Made it `Opcodes.V23` so it stays tied to the asm version.

tornado-assembly/src/bin/tornado-test
mairooni: This test was not removed, it is just not whitelisted anymore.

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/foundation/TestHalfFloats.java
mairooni: `getHalfFloatValue()` returned a short, but the `localSum` array is float, that's why this change is made.

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/WriteHalfFloatNode.java
mairooni: Done

tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/graal/nodes/WriteHalfFloatNode.java
mairooni: Done

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/compiler/plugins/OCLHalfFloatPlugins.java
mairooni: This plugin is necessary now because of the change in the unittest `TestHalfFloats` (see above).

tornado-drivers/opencl/src/main/java/uk/ac/manchester/tornado/drivers/opencl/graal/nodes/WriteHalfFloatNode.java
mairooni: Done

docs/source/installation.rst
mairooni: Done

.gitignore
mairooni: This is the `flatten-maven-plugin` output. It lets each profile stamp its own artifact version. Since this is regenerated by every build and is removed by clean, it is correctly ignored.

docs/source/faq.rst
mairooni: Since we still have the `graal-jdk-21` profile, I think it should stay.

PR #910, opened 2026-07-10
[cuda] Fix misleading NVRTC compilation errors: memoized FP16 header resolution + explanatory driver/toolkit diagnostics

orionpapadakis:
#### Description

This PR removes the misleading NVRTC_ERROR_COMPILATION noise emitted by the CUDA backend at kernel-compilation time and replaces opaque compile/load failures with actionable diagnostics that name the stale component (toolkit vs. driver) and the remediation.

The first NVRTC attempt for an FP16 kernel (`#include <cuda_fp16.h>`) can legitimately fail with a retryable "could not open source file" and then succeed once the toolkit include path is supplied — but that intermediate failure was printed unconditionally as an ERROR, alarming users on a healthy setup and obscuring real errors. Retryable attempts are now quiet; terminal failures are still reported with the full build log.

Whether NVRTC resolves `<cuda_fp16.h>` via built-ins, needs an explicit --include-path, or cannot resolve it at all is a fixed property of the (NVRTC, toolkit) pair, so it is now discovered once (by the first FP16 kernel) and memoized, collapsing the built-in-fail→include-path retry from once-per-kernel to once-per-process. The canonical `#include <cuda_fp16.h>` is retained (rather than a hand-rolled __half preamble) so the backend stays immune to the code generator emitting new fp16 intrinsics.

Genuine capability gaps now produce explanatory, categorized diagnostics embedded in the build log (which also propagates into the thrown `TornadoBailoutRuntimeException`, not just `System.err`):

- Unresolvable `<cuda_fp16.h>` → root cause + remediation (install a toolkit / set `CUDA_PATH`).
- PTX fallback rejected at load (toolkit older than GPU, driver too old for the PTX ISA) → update the GPU driver, or use a toolkit that natively supports the GPU arch to skip PTX JIT.
- Native cubin rejected at load (toolkit newer than driver) → update the driver / use a matching toolkit.
- Working-but-suboptimal PTX-JIT fallback (toolkit older than GPU, driver JITs successfully) → a one-time WARNING that the path is functional but not optimal (extra load-time JIT; codegen capped at the fallback virtual ISA), advising a toolkit upgrade for native codegen.

#### Problem description

 On a system whose CUDA toolkit predates the GPU (e.g. NVRTC 12.0, max sm_90, on a Blackwell sm_120 GPU), starting any FP16 workload printed a wall of:
```
  [TornadoVM-CUDA-NVRTC-JNI] ERROR : nvrtcCompileProgram -> NVRTC_ERROR_COMPILATION
```
These were benign — first-attempt header-resolution misses that the backend already retried successfully via --include-path — but they looked like hard failures and obscured any real compilation error. Genuine failures also surfaced only as opaque `NVRTC_ERROR_COMPILATION` / `CUDA_ERROR_INVALID_PTX` codes with no indication of cause or fix.

Reproduce (pre-fix): run any FP16 kernel / GPULlama3 model on a host where NVRTC has no built-in CUDA headers (e.g. Ubuntu's distro NVRTC 12.0) → benign ERROR spam at every kernel compile.

  #### Backend/s tested

  - [ ] OpenCL
  - [ ] PTX
  - [x] CUDA
  - [ ] SPIRV
  - [ ] Metal

  #### OS tested

  - [x] Linux
  - [ ] OSx
  - [ ] Windows

  #### Did you check on FPGAs?

  - [ ] Yes
  - [x] No (not applicable — CUDA backend only)

  ```bash
  make BACKEND=cuda
  make tests

  # FP16 unit suites (counts unchanged vs. pre-patch; zero NVRTC_ERROR lines):

  tornado-test -V uk.ac.manchester.tornado.unittests.foundation.TestHalfFloats                          # 4/4
  tornado-test -V uk.ac.manchester.tornado.unittests.vectortypes.TestHalfFloats                         # 25 ran, 0 failed, 23
  unsupported
  tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.reductions.TestLocalMemoryReductionsHalfFloats  # 4 ran, 0 failed, 2
  unsupported
  tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestHalfFloatLocalArray          # 1/1
```

Verified additionally:
  - FP16 workloads compile and run correctly with zero NVRTC_ERROR output; --printKernel confirms #include <cuda_fp16.h> is present in
  emitted FP16 kernels.
  - The built-in-header miss + include-path retry happens exactly once per process (instrumented count = 1 across dozens of FP16
  kernels), not per kernel.
  - On a toolkit-older-than-GPU host, the one-time suboptimal-fallback WARNING fires exactly once and the workload produces correct
  output.
  - Exception enrichment: forcing a synthetic NVRTC compile error makes the thrown TornadoBailoutRuntimeException message carry the
  full NVRTC build log (previously only on System.err).

### DIFF
--- a/tornado-drivers/cuda-jni/src/main/cpp/source/CUDAProgram.cpp
+++ b/tornado-drivers/cuda-jni/src/main/cpp/source/CUDAProgram.cpp
@@ -38,6 +38,23 @@
 static std::mutex g_cubin_cache_mtx;
 static std::unordered_map<std::string, std::string> g_cubin_cache;
 
+// One-time NVRTC header-resolution mode. Whether NVRTC can resolve
+// #include <cuda_fp16.h> on its own, needs an explicit toolkit include path, or
+// cannot resolve it at all is a fixed property of the (NVRTC, toolkit) pair on
+// this machine — not of any individual kernel. It is therefore discovered once
+// (by the first FP16 kernel) and reused, so later FP16 kernels compile in a
+// single attempt instead of repeating the failed built-in probe. Kernels with
+// no #include never consult it. See compile_with_nvrtc.
+enum class Fp16IncludeMode { UNKNOWN, NEEDS_INCLUDE_PATHS, UNRESOLVABLE };
+static std::mutex g_fp16_mode_mtx;
+static Fp16IncludeMode g_fp16_mode = Fp16IncludeMode::UNKNOWN;
+static std::vector<std::string> g_fp16_include_opts; // "--include-path=..." dirs, once resolved
+
+// One-time latch for the working-but-suboptimal PTX-JIT fallback warning (GPU
+// arch newer than the toolkit). The fallback runs correctly, so this is a
+// warning, not an error; emit it once per process rather than once per kernel.
+static std::once_flag g_arch_fallback_warn_once;
+
 extern "C" {
 
 /* OpenCL cl_program_build_info / cl_program_info values used by the Java clone. */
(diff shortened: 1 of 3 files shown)


mikepapadim (approved):
LGTM

stratika (approved):

PR #919, opened 2026-07-11
[cuda] NVTX instrumentation for library tasks + cuSPARSE hybrid provider (SpMV/SpMM)

mikepapadim:
#### Description

Two additions toward first-class NVIDIA-tooling and CUDA-X coverage from the JVM, plus a roadmap doc.

**1. NVTX instrumentation of library tasks.** TornadoVM already NVTX-labels its JIT kernels and transfers; this extends the same treatment to **hybrid-API library tasks** so cuBLAS / cuDNN / CUTLASS / cuTENSOR / cuSPARSE calls show up as named spans in **Nsight Systems**. A host-side NVTX range (`libraryName/functionName`) wraps `provider.dispatch(...)` in the interpreter, routed through a new no-op-default `nvtxRangePush/Pop` on `TornadoNativeStreamSupport`, implemented by the CUDA backend via two `cuda-jni` natives (header-only NVTX3, near-zero cost with no profiler attached).

Verified with `nsys profile --trace=cuda,nvtx`: library tasks appear as ranges, e.g. `:nvidia/cublas/cublasGemmExFP16` and `:nvidia/cublas/cublasSgemm`. `TestNvtx` (4) guards that the instrumentation is transparent (correct results, balanced push/pop across single / repeated / mixed-JIT / CUDA-graph runs). No regression: TestCuBlas 11/11, TestCuDnn 9/9.

**2. cuSPARSE hybrid provider (`nvidia/cusparse`).** FP32 sparse BLAS as library tasks, mirroring the cuDNN/cuTENSOR module pattern:

| Factory | Operation |
|---|---|
| `Cusparse.cusparseSpMV(rows, cols, nnz, csrRowOffsets, csrColInd, csrValues, x, y)` | `y = A·x` |
| `Cusparse.cusparseSpMM(rows, k, n, nnz, csrRowOffsets, csrColInd, csrValues, B, C)` | `C = A·B` |

CSR (32-bit indices, zero-based), row-major dense operands; handle bound to the plan stream; a capture-safe default workspace (the external buffer size depends on the sparse structure, queried at dispatch). cuSPARSE ships with the CUDA toolkit; `cusparse-jni` is skipped if not found (build stays green, tasks report UNSUPPORTED).

Exhaustive `TestCusparse` (11), all green on an RTX 4090: SpMV over general / identity / diagonal / rectangular / empty-row / dense-ish patterns, SpMM, JIT→cuSPARSE→JIT, CUDA-graph capture, multiple shapes in one context, and 50-iteration stability.


stratika (approved):
LGTM

PR #928, opened 2026-07-14
Add FP8 (E4M3/E5M2) support for the CUDA backend + fix NaN/Infinity float constants

mikepapadim:
## FP8 (E4M3 / E5M2) support for the CUDA backend

Adds 8-bit floating-point storage — the low-precision weight format modern GPUs and vLLM use — to the CUDA backend.

### What's added
- **`types.FP8`** — OCP FP8 codecs in both formats: **E4M3** (bias 7, no infinity, max ±448) and **E5M2** (bias 15, IEEE-like). Encode is round-to-nearest-even + saturation; **the decoders are kernel-safe** (integer div/mod + float arithmetic only, no `Float.intBitsToFloat`), so FP8 weights held in a `ByteArray` dequantize to `float` **inside a TornadoVM kernel on the CUDA backend** with no special code generation — the same pattern the Q8 dequant path uses:
  ```java
  out.set(i, FP8.e4m3ToFloat(byteArray.get(i)));   // runs on the GPU
  ```
- **`types.arrays.FP8Array`** — a 1-byte-per-element native array (host-side typed storage; registered in the CUDA device allocator).
- **CUDA backend fix (general):** `Float.NaN` / `Float.POSITIVE_INFINITY` / negative infinity float and double constants were emitted as the bare identifier `NaN`/`Infinity` (→ undefined `NaNF`) and failed to compile. They now materialise via `__int_as_float(0x7fc00000)` / `__longlong_as_double(...)`. Any kernel using these constants benefits.
- **`TestFP8`** — host round-trip over **all 256 codes in both formats**, host `FP8Array`, and **on-device dequantization** kernels checked bit-for-bit against the host decode.

### Verification (RTX 4090, `make BACKEND=cuda`)
```
tornado-test -V ...arrays.TestFP8   →  Test ran: 8, Failed: 0
make fast-tests                     →  0 failures (no regressions)
```

### Notes for reviewers
The kernel-safe decoders are written deliberately to avoid three CUDA-codegen pitfalls found while bringing this up on the GPU (each worked around in `FP8`, and worth flagging):
1. **Multiple `return`s** in a helper inlined into a `@Parallel` loop mis-lower (the `if` guard captured only its first statement) → single-return `if/else` chains.
2. **Two loops in separate `if/else` branches** merge their results incorrectly in the CUDA phi-lowering (the merge block overwrote the taken-branch value with the other branch's uninitialised loop variable) → `pow2` uses two *sequential* loops on one accumulator.
3. `>>>` / `& 0xFF` on a byte are treated as unsigned and mis-extract fields → division/modulo on an explicitly non-negative byte.

Items 1–2 look like latent CUDA-backend bugs beyond FP8; happy to file them separately with the reduced repros.

Native `__nv_fp8_e4m3` (cuda_fp8.h) hardware-conversion codegen — mapping the type through `CUDAKind` + a read node like `__half` — is the follow-up for tensor-core FP8; this PR lands the storage + software-decode path first.

@orionpapadakis — CUDA-only, off `develop`. Review welcome.


mikepapadim:
@orionpapadakis FP8 (E4M3/E5M2) for the CUDA backend — OCP codecs with kernel-safe decoders (FP8 weights in a ByteArray dequantize on the GPU like Q8), + a general fix for Float.NaN/Infinity constants that didn't compile on CUDA. TestFP8 8/8 on RTX 4090, make fast-tests green. Found 2 latent CUDA phi/inline codegen bugs en route (noted in the PR, can file separately). Native __nv_fp8_e4m3 hardware path is the documented follow-up.

mikepapadim:
## Update: device-side FP8 encode, cuBLASLt FP8 GEMM, and a FP8-vs-FP16 benchmark

Following review-readiness, three additions (commit cef2afb):

### 1. Device-side FP8 encode (in-kernel quantization)
The encoders are now **kernel-safe** (single return, bounded normalisation loop instead of `Math.getExponent`, `(int)(x+0.5)` instead of `Math.round`), so float activations can be **quantized to FP8 inside a CUDA kernel**, not just dequantized. Verified by `testE4M3EncodeOnDevice` (float -> FP8 -> float round-trip on the GPU). Also: E5M2 overflow now rounds to infinity at the correct boundary (61440), and `-0.0` collapses to `+0.0`.

### 2. Hardware FP8 GEMM via the hybrid library API (`CuBlasLt.ltMatmulFP8`)
FP8 (E4M3) matmul on the GPU's **FP8 tensor cores** through cuBLASLt: FP8 A/B (one byte/element, `ByteArray`) -> FP16 C, FP32 compute. The cuBLASLt plan path was already data-type-generic, so this is Java-only (no JNI change). `testLtMatmulFP8` passes on an RTX 4090 (sm_89).

### 3. Benchmark + profiling: FP8 vs `HalfFloatArray` (RTX 4090)
`tornado-examples/.../FP8Benchmark` — dequant-and-sum of N weights, input-read-bound (tiny output).

| dequant | ms/iter | eff. read BW | bytes read |
|---|---:|---:|---:|
| FP8 (1 B/elem, software decode) | 0.522 | 128.5 GB/s | 64 MB |
| FP16 (2 B/elem, hardware `__half2float`) | 0.237 | 566.8 GB/s | 128 MB |

nsys per-kernel (N=33M): **`sumFP8` 208 us vs `sumFP16` 21 us**.

**Finding (honest):** software FP8 decode is **compute-bound** (the `pow2` loop + div/mod extraction), so despite moving half the bytes it is ~2-10x slower than the hardware FP16 expansion. FP8's win today is **memory capacity** (fits 2x the weights) and **compute via the FP8 tensor-core GEMM above** - not in-kernel software dequant. This is the concrete case for the **native `__nv_fp8_e4m3` conversion intrinsic** (cuda_fp8.h) as the follow-up: it would make in-kernel FP8 decode ~free and turn the 50% bandwidth saving into a real speedup.

### Codegen bugs filed
The two latent CUDA-backend codegen issues found while bringing FP8 up on the GPU are now filed with reduced repros: #930 (two loops in if/else branches -> wrong phi merge) and #931 (early returns in an inlined helper mis-lower). Both are worked around in FP8.

`make checkstyle` and `make fast-tests` green (the 8 fast-test failures are pre-existing flaky reduction races - `testComputePi` etc. pass in isolation and shift between runs; unrelated to these changes).

@orionpapadakis

mikepapadim:
Gated the five on-device FP8 tests to the CUDA backend (commit 6346bdf): `assumeCudaBackend()` -> they report Unsupported (not Failed) on OpenCL/PTX/SPIRV/Metal, matching the cuBLAS/cuTENSOR guard; host-codec tests stay backend-agnostic. Verified on RTX 4090: TestFP8 13/13 on CUDA; TestHello/TestCuBlas/TestCuBlasLt all green alongside it.

mairooni (approved):
LGTM

PR #934, opened 2026-07-15
Auto-disable intra-plan concurrency on serial task graphs

orionpapadakis:
#### Description

Removes the performance overhead `withIntraPlanConcurrency()` (multi-stream/CUDA-graph-compatible intra-plan concurrency, "IPC") previously added to serial (chain-shaped) task graphs, and trims general per-operation and per-join bookkeeping cost for all IPC plans. Stacked on the `vm.deps`/`WorkerGrid` correctness fix #932 .

Three changes:

1. **Width-1 auto-disable.** `IntermediateTornadoGraph.isTaskChain()` computes, at graph-compile time, whether a (non-batched) graph's tasks form a pure dependency chain (max antichain width 1 — no two tasks are mutually independent). Such a graph has no kernel-level concurrency for multi-stream issue to exploit; running it single-stream avoids cross-stream event overhead entirely. The result is threaded through `TornadoVMBytecodeResult` and consulted in `TornadoVMInterpreter.isIntraPlanConcurrencyActive()`, so a plan that requested IPC still runs its serial graphs on the default stream while any graph with real parallel width (e.g. batched prefill) keeps using the multi-stream path. Batched graphs are always excluded from the chain check (their pipeline overlaps chunks of one task and needs the role streams regardless of intra-task width).

2. **Dirty-queue joins (CUDA).** `CUDADeviceContext.joinRoleQueues` previously enqueued a marker + cross-stream wait on every secondary queue (up to 6: H2D/D2H role queues + the COMPUTE pool) on every `execute()`, even when only one was touched. Each `CUDACommandQueue` now carries a `dirty` flag set when it receives work; `joinRoleQueues` joins and clears only the queues actually touched since the last join.

3. **Removed a redundant full-buffer clear.** `CUDAEventPool`/`OCLEventPool`'s `serialiseEvents` cleared the entire `waitEventsBuffer` (sized by `-Dtornado.eventpool.maxwaitevents`, e.g. 256KB at a common `32000` setting) on every call, even though only `[0..index]` is ever read by the native side and those entries are freshly written in the same call.

#### Problem description

Not a bug fix — a performance regression introduced by IPC itself. Profiling (`nsys`) an IPC-enabled decode-chain workload (GPULlama3.java) against the same workload with IPC off showed +67k `cuStreamWaitEvent`, +34k `cuEventRecord` calls, and `cuStreamSynchronize` going from a 323ns median (in-order queue, effectively free) to 175µs (real cross-stream sync) for identical work — roughly 400ms of pure host-side blocking over a 2-second run. Net effect: IPC-on decode was 10-15% slower than IPC-off on small models, shrinking to ~1-3% on large models (the fixed per-op overhead amortizes over bigger kernels), even though IPC never produces wrong output once the correctness fix lands — it was strictly a self-inflicted cost on workloads with no parallel structure to exploit.

#### Backend/s tested

- [x] OpenCL
- [ ] PTX
- [x] CUDA
- [ ] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=cuda
source setvars.sh
tornado-test --ea uk.ac.manchester.tornado.unittests.kernelcontext.api.TestCombinedTaskGraph
tornado-test --ea uk.ac.manchester.tornado.unittests.streams.TestCUDAStreams
tornado-test --ea uk.ac.manchester.tornado.unittests.batches.TestBatches
```
All green (5/5, 11/11, 31/31).

End-to-end (GPULlama3.java, RTX 5090 Laptop, `-Dgpullama.intraPlanConcurrency=true` vs `=false`, `-Dtornado.cuda.host.pinning=false`), eval tok/s delta IPC-on vs IPC-off, before -> after this PR:

| model | mode | before | after |
|---|---|---|---|
| 1B | standard | -13.5% | 0.0% |
| 1B | prefill-decode | -11.6% | +0.2% |
| 1B | batched-prefill-decode | -10.2% | -0.7% |
| 3B | standard | -4.8% | +0.4% |
| 3B | batched-prefill-decode | -14.2% | +4.9% |
| 8B | standard | ~0% | ~parity |

Output verified coherent in every cell, including the all-features-on combination (IPC + cuda-graphs + staged transfers).

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDACommandQueue.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDACommandQueue.java
@@ -54,6 +54,13 @@ public class CUDACommandQueue extends CommandQueue {
     private final long properties;
     private final int openclVersion;
 
+    /**
+     * Set when work is issued to this (secondary) queue under intra-plan concurrency; drained
+     * by the per-plan join (see {@code CUDADeviceContext#joinRoleQueues}), which only needs to
+     * synchronise queues that were actually touched since the last join.
+     */
+    private volatile boolean dirty;
+
     public CUDACommandQueue(long commandQueuePtr, long properties, int version) {
         this.commandQueuePtr = commandQueuePtr;
         this.properties = properties;
@@ -64,6 +71,22 @@ public class CUDACommandQueue extends CommandQueue {
 
     public long getCommandQueuePtr() {
         return commandQueuePtr;
+    }
+
+    /** Marks this queue as having received work since the last join. */
+    public void markDirty() {
+        if (!dirty) {
+            dirty = true;
+        }
+    }
+
+    /** Returns whether this queue was touched since the last join, clearing the flag. */
+    public boolean pollDirty() {
+        if (dirty) {
+            dirty = false;
+            return true;
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


stratika (approved):
LGTM

PR #936, opened 2026-07-15
Stage large H2D transfers through a pinned host buffer ring

orionpapadakis:
#### Description

Adds an opt-in staged host-to-device transfer path for the PTX and CUDA backends, aimed at large one-shot uploads (e.g. `FIRST_EXECUTION` model weights). Large transfers are split into chunks; each chunk is memcpy'd (multi-threaded) into a pinned host staging slot from a small ring, then uploaded asynchronously, so staging chunk `i+1` overlaps the DMA of chunk `i`. Unlike the direct path, the source buffer does not need to be page-locked up front — on both backends, read-only segments on this path skip the whole-segment host-pin registration (`cuMemHostRegister` on CUDA), whose synchronous page-in is the dominant cost for large mmap'd sources (e.g. GGUF model files). See the background on host pinning below.

Opt-in per execution plan via `TornadoExecutionPlan.withStagedTransfers()` (with `withoutStagedTransfers()` to force the direct path):

```java
try (TornadoExecutionPlan plan = new TornadoExecutionPlan(itg)) {
    plan.withStagedTransfers();
    plan.execute();
}
```

`-Dtornado.staged.transfers=true` still enables it process-wide, and acts as the default that the plan-level call overrides — the same "`-D` seeds a read-once default, the API overrides a per-plan field" pattern the runtime already uses for other plan options. Tuning stays on system properties: chunk size, ring depth, minimum transfer size to engage, and fill-thread count (`-Dtornado.staged.chunk.size`, `.ring.depth`, `.min.size`, `.fill.threads`). Default minimum size to engage is 16MB (lowered from an initial 64MB during evaluation — see below).

The pinned staging ring is allocated lazily on first use and released at device teardown (`clean()`), draining any in-flight slot DMA before freeing the pinned block; a later staged transfer rebuilds it. This mirrors the PTX backend, which frees its ring in `PTXStream.cuDestroyStream()`.

#### Problem description

Not a bug fix — a load-time throughput improvement for large one-shot host-to-device uploads. Some background on why the existing pinning path is the wrong tool here:

**Background: CUDA host pinning.** `-Dtornado.cuda.host.pinning` (default: on) controls whether TornadoVM page-locks (`cuMemHostRegister`) a host buffer before an async transfer, so the GPU DMA engine can copy directly to/from those physical pages instead of going through the driver's own pageable staging copy. For small/medium buffers that are allocated once and reused across many executions (activations, KV cache, weights that fit comfortably), this is a clear win and the reason it defaults on — the registration cost is paid once and amortized forever, and it's what makes true async transfer/compute overlap possible.

The cost of `cuMemHostRegister` is a **synchronous** walk of every page in the range. For a large, cold, one-shot buffer — the classic case being an mmap'd multi-gigabyte GGUF model file, uploaded exactly once at startup — that walk has to fault in and page-lock the *entire* segment (paying the mmap page-in cost too) before the first byte can even start transferring. Measured on this box, that upfront cost is large enough to roughly **halve** effective weight-load throughput versus leaving pinning off for this specific access pattern (3B: ~1810ms -> ~900ms; 8B: ~3865ms -> ~1808ms, pinning on vs off, staged transfers not involved).

This PR's staged path applies **only to large H2D transfers** and does not touch the pinning behavior of any other buffer. When staged transfers is enabled, a transfer is routed through the staging ring only if it is non-batched, read-only, and at or above the size threshold (`-Dtornado.staged.min.size`, default 16MB). For exactly those buffers — and only those — the whole-segment pin is skipped, because the staging ring's own small pre-pinned slots already make the chunked DMA async, so pinning the (often cold, mmap'd) source would just re-add the synchronous page-in cost the ring exists to avoid. Every other buffer — small transfers, read-write buffers, batched transfers — keeps the exact default behavior, including host pinning. This scoping is deliberate: `access == READ_ONLY` is only a safe proxy for "one-shot upload" in combination with the large-size gate (a large buffer re-uploaded every execution is already pathological); a small read-only buffer may well be a hot, repeatedly-transferred one where pinning genuinely pays off, so it is left pinned.

Evaluated end-to-end on GPULlama3.java (RTX 5090 Laptop, FP16 weights, warm page cache), weight copy-in time, host pinning left at its default (on), median of 3 reps per cell:

| model | default (staged OFF) | `-Dtornado.staged.transfers=true -Dtornado.staged.min.size=16777216` | speedup | decode tok/s (OFF → ON) |
|---|---|---|---|---|
| 1B | 860ms | 476ms | **1.8x** | 72.6 → 71.1 |
| 3B | 1846ms | 785ms | **2.4x** | 31.7 → 31.2 |
| 8B | 3891ms | 1340ms | **2.9x** | 17.2 → 17.4 |

Benefit scales with model size, as expected (bigger one-shot uploads amortize staging setup better and are exactly the transfers that were previously eating the full pinning page-in cost). Decode tok/s is unaffected either way (weight copy-in is a one-time cost at model load, not on the token-generation hot path). Output verified coherent in every cell. Combines cleanly with `withIntraPlanConcurrency()` (see companion PR) and CUDA graphs — no interaction penalty in the full evaluation matrix.

Tuning note: the initial 64MB `staged.min.size` default was too conservative — most weight tensors in common quantizations fall under that threshold and never engaged the staging ring. Lowered to 16MB based on a sweep (8MB was slightly worse — per-chunk overhead starts to dominate).

#### Backend/s tested

- [ ] OpenCL
- [x] PTX
- [x] CUDA
- [ ] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=cuda
source setvars.sh
tornado-test --ea uk.ac.manchester.tornado.unittests.streams.TestCUDAStreams
tornado-test --ea uk.ac.manchester.tornado.unittests.streams.TestStreamsPerformance
```

Enable per plan via `plan.withStagedTransfers()`, or process-wide (single flag; leave host pinning at its default):
```bash
-Dtornado.staged.transfers=true -Dtornado.staged.min.size=16777216
```

End-to-end with GPULlama3.java: apply the flags above and compare `Read-only weights Copy-in` from `--verbose-init` against a run without them (the current default). Host pinning is left at its default; only the large read-only weight uploads served by the staging ring skip pinning.


stratika (approved):
LGTM

Issue #940, opened 2026-07-15
NVIDIA feature parity: native FP8 MMA + async-copy (cp.async)/TMA


mikepapadim:
## NVIDIA feature parity: native FP8 MMA + async-copy (`cp.async`) / TMA

**Type:** enhancement / roadmap proposal (no code in this issue)

Motivated by a comparison with NVIDIA Labs' **cuda-oxide** (Rust -> PTX,
[nvlabs.github.io/cuda-oxide](https://nvlabs.github.io/cuda-oxide/)): as a NVIDIA-Labs compiler it
will track new NVIDIA hardware features (tensor cores, low precision, async memory) closely. This is
the area where a single-target NVIDIA compiler moves fastest, and where TornadoVM should have a
codegen roadmap.

### Current state
FP8 *storage* + a cuBLASLt FP8 tensor-core GEMM landed (PR #928), but **in-kernel FP8 decode is
software/arithmetic and compute-bound** (a benchmark showed FP8 software dequant ~2-10x slower than
hardware `__half2float`, despite half the bytes). The bandwidth win only materializes with hardware
FP8 conversion.

### Proposal (codegen roadmap)
- **Native `__nv_fp8_e4m3` / `__nv_fp8_e5m2`** conversion + FP8 tensor-core MMA in the CUDA backend
  (`cuda_fp8.h` include injection + `CUDAKind` mapping + a read/convert node, mirroring the existing
  `__half` machinery) so in-kernel FP8 is free rather than software-decoded.
- **Async global->shared copy (`cp.async`)** and, on Hopper+, **TMA** + warp-specialization, to
  overlap memory with compute in the MMA GEMMs.
- BF16 as a first-class MMA input type.

Reference: `CUDA_OXIDE_COMPARISON.md` (section 5.2) and the earlier hybrid-libs/FP8 findings.

Issue #931, opened 2026-07-14
CUDA backend: early returns in a helper inlined into a @Parallel loop mis-lower


mikepapadim:
## CUDA backend: early `return`s in a helper inlined into a `@Parallel` loop mis-lower (guard misses block)

**Backend:** CUDA · **TornadoVM:** develop (5.0.1-jdk21-dev) · **GPU:** RTX 4090 (sm_89)

A `static` helper with **multiple `return` statements** (early returns), inlined into a `@Parallel` kernel loop, generates CUDA where an `if` guard captures **only its first statement** (no block braces), so the following statements execute unconditionally and the control flow is wrong.

### Reduced repro
```java
static float classify(int b) {
    int sign = b / 128;
    float sig = (sign == 0) ? 1.0f : -1.0f;   // ternary
    if (sign == 0) {
        return sig * 2.0f;                     // early return A
    }
    return sig * 3.0f;                         // return B
}

public static void kernel(IntArray in, FloatArray out) {
    for (@Parallel int i = 0; i < out.getSize(); i++) {
        out.set(i, classify(in.get(i)));
    }
}
```
For `sign == 0` the sign factor is emitted as always `-1.0` (the ternary's else branch runs unconditionally), so the result has the wrong sign / value.

### Generated CUDA (abridged)
```c
if (b_17)          // sign == 0
  f_18 = 1.0F;
f_18 = -1.0F;      // <-- runs unconditionally: no braces, so f_18 is always -1.0F
```

### Workaround
Use a **single return** through a result variable and an `if`/`else if`/`else` chain (and prefer arithmetic, e.g. `sig = 1.0f - 2.0f * sign`, over a ternary). With a single return + explicit blocks the generated CUDA has correct `{ }` guards.

Found while bringing up FP8 codecs on the CUDA backend. Reproduces reliably; dump the kernel with `--printKernel`.

Issue #930, opened 2026-07-14
CUDA backend: two loops in separate if/else branches merge incorrectly (phi lowering)


mikepapadim:
## CUDA backend: two loops in separate `if`/`else` branches merge incorrectly (phi lowering)

**Backend:** CUDA · **TornadoVM:** develop (5.0.1-jdk21-dev) · **GPU:** RTX 4090 (sm_89)

When a helper contains an `if`/`else` where **each branch runs its own loop that produces a value**, the CUDA code generator's merge (phi) block overwrites the taken branch's result with the *other* branch's (uninitialised) loop variable. The value used after the merge is garbage.

### Reduced repro
```java
// 2^n via two loops, one per direction, in if/else branches:
static float pow2(int n) {
    float r = 1.0f;
    if (n >= 0) {
        for (int i = 0; i < n; i++) { r = r * 2.0f; }     // branch A -> r
    } else {
        for (int i = 0; i < -n; i++) { r = r * 0.5f; }    // branch B -> r
    }
    return r;
}

public static void kernel(IntArray exps, FloatArray out) {
    for (@Parallel int i = 0; i < out.getSize(); i++) {
        out.set(i, pow2(exps.get(i)));   // e.g. exps[i] = 0 -> expect 1.0
    }
}
```
For `n = 0` the expected result is `1.0`; the device produces a large garbage value.

### Generated CUDA (abridged)
```c
if (b_37) {            // n < 0
    ... for (...) f_39 = f_39 * 0.5F; ...   // BLOCK 13: f_43 = f_39
} else {
    ... for (...) f_44 = f_44 * 2.0F; ...   // BLOCK 17: f_43 = f_44
}
// BLOCK 18 MERGES [17 13]
f_43 = f_39;           // <-- BUG: unconditionally overwrites the merged f_43 with the
                       //     n<0 branch's loop var, which is uninitialised on the n>=0 path
f_50 = ... * f_43;
```
The merge should keep each predecessor's `f_43`; instead it re-assigns `f_43 = f_39`.

### Workaround
Rewrite so both loops are on the straight-line path (sequential, one accumulator) instead of in `if`/`else` branches — then there is no branch-merge of loop results and codegen is correct.

Found while bringing up FP8 dequant on the CUDA backend (a `pow2` helper). Reproduces reliably; the whole generated kernel can be dumped with `--printKernel`.

PR #952, opened 2026-07-18
[feat] CUDA backend: native FP8 conversion + FP8/BF16 tensor-core MMA and cp.async copies

mairooni:
#### Description

Implements #940 (NVIDIA feature parity for the CUDA backend): native FP8 conversion, FP8
tensor-core MMA, cp.async global→shared copies, and BF16 as a first-class MMA input type.
Four commits (all benchmarks: RTX 4070 Laptop, sm_89):

**1. Native FP8 conversion + FP8 tensor-core MMA (58b86b9b8)**

- `FP8.e4m3ToFloat`/`e5m2ToFloat` now compile to `__nv_cvt_fp8_to_halfraw` + `__half2float`
  inside CUDA kernels, with `cuda_fp8.h` injected under the same source-scan gating as
  `cuda_fp16.h`. Decode only: the hardware encoder rounds ties to even while the software
  codec rounds half away from zero, so encode stays software on every backend, and all other
  backends keep inlining the Java decoder.
- FP8 (OCP E4M3/E5M2) is wired as a third MMA input format: `mma.sync.aligned.m16n8k32` with
  an f32 accumulator (`KernelContext.mmaFP8E4M3`/`mmaFP8E5M2`). The A/B tile layout is
  byte-identical to the int8 m16n8k32 path, so the FP8 loads reuse the int8 load nodes; only
  the compute op carries the element type (`MMAComputeStmt` no longer infers it from the tile
  shape, which m16n8k32 no longer identifies uniquely).
- Gating: sm_89 (Ada/Hopper — FP8 tensor cores don't exist at the sm_80 floor), enforced
  through the same compiler phase (`TornadoDeviceMMANotSupported`) that gates fp16/int8 MMA
  at sm_80 — not the dispatch-time probe used by the cuBLASLt library path in #928.
- Benchmark: the FP8 dequant benchmark drops from 0.522 ms/iter (software decode, compute-bound) 
  to **0.222 ms/iter** — now **2.29x faster than the fp16 hardware-convert path** in the same run, 
  as a memory-bound kernel should be. The in-kernel decode gap flagged in #928 is closed.

**2. cp.async global→shared copies (aadac999f)**

- `KernelContext.asyncCopyToLocal` + `asyncCopyCommit`/`asyncCopyWaitGroup` expose Ampere's
  LDGSTS: one 4-byte slot (two fp16/bf16 halves or four fp8/int8 bytes) is copied straight
  from a global Tornado array into a shared int tile, bypassing the register round-trip of
  the previous load-pack-store loops. `cp.async.commit_group`/`cp.async.wait_group` order the
  copies before the barrier that publishes the tile.
- Route taken: inline PTX asm, consistent with the existing dp4a/ldmatrix/mma statements.
  `cuda::pipeline` was rejected because it needs libcudacxx headers under NVRTC, which the
  toolkit include probing does not guarantee on every install.
- Gating: no new gate. cp.async shares mma.sync's sm_80 floor, so its nodes are gated by the
  existing tensor-core support phase — stated explicitly here rather than silently assumed.
- Benchmark: 2048³ fp16 GEMM: 6082 → 7931 GFLOP/s (**1.30x**) over the synchronous-load MMA
  kernel; Llama-3.2-1B FFN gate-up shape (M=128 N=8192 K=2048): 0.944 → 0.696 ms (**1.36x**).

**3. BF16 as a first-class MMA input type (9c3dc5233)**

- New `types.BFloat16` codec (kernel-safe software encode/decode, same style as `FP8.java`);
  on CUDA the decoder lowers to `__bfloat162float` via `cuda_bf16.h` (same include gating
  pattern), plus a `CUDAKind.BFLOAT` (`__nv_bfloat16`) entry. Encode stays software
  everywhere (ties-to-even vs half-away-from-zero, as with FP8).
- `KernelContext.mmaBF16` emits `mma.sync.aligned.m16n8k16.row.col.f32.bf16.bf16.f32`.
  Confirmed against the PTX ISA: bf16 shares fp16's m16n8k16 tile shape and per-lane fragment
  layout, so the existing `mmaLoadA`/`mmaLoadB` fragment loads work unchanged on raw bf16 bit
  pairs; only the mma.sync element type differs. Same sm_80 gate as fp16/int8 MMA.
- HalfFloat-replacement audit: every backend (including CUDA) runs a
  `TornadoHalfFloatReplacement` phase that keys off `HalfFloat`/`HalfFloatArray` node types.
  BF16 values deliberately travel as raw bits in `ShortArray`, so the phase never sees them
  on any backend; and the `HalfFloat[]` MMA fragment handles are the same opaque register
  tuples the existing fp16 MMA path already runs through that phase safely — they are never
  element-accessed, so no fp16 decode is ever applied to bf16 bits. No fp16-only assumption
  is reachable from this path.
- Numerical delta: M=N=K=64, U[−1,1] inputs, same data quantized to both formats and compared
  against the exact f32 GEMM: max |err| = **0.0207 (bf16) vs 0.0029 (fp16), 7.2x** — tracking
  the 3-bit mantissa gap (2³ = 8x expected). Asserted in
  `TestMatrixMultiplicationMMABF16#testBF16VersusFP16Delta`, printed in the test log.
- Benchmark: 2048³ GEMM: **6161 GFLOP/s (bf16) vs 5995 GFLOP/s (fp16)** — throughput parity
  (1.03x, within run-to-run noise), as expected for identical tile shapes and bandwidth.
  BF16 is a range/robustness tradeoff, not a speed win.

**4. Non-CUDA backends reject the new intrinsics (14d243049)**

- The OpenCL, PTX, SPIRV and Metal graph-builder plugins register `unimplemented()` stubs for
  all ten new `KernelContext` methods (FP8 loads/computes, `mmaBF16`, the three
  `asyncCopyToLocal` overloads, `asyncCopyCommit`/`asyncCopyWaitGroup`),  so a kernel using them 
  fails the sketch with a clear "only supported for the CUDA backend" diagnostic instead of inlining the Java
  fallback bodies and silently computing wrong results. PTX gets the stubs too: it implements
  the fp16/int8 MMA ops but not these extensions. (The `BFloat16`/`FP8` codecs themselves
  stay usable on every backend — they're plain kernel-safe Java; only the tensor-core and
  cp.async intrinsics are gated.)

**Skipped: TMA + warp-specialization (Hopper, sm_90+)** — no Hopper-class hardware was
available to validate against, and an unvalidated Hopper path is worse than an absent one.

**Notes on #930/#931 (latent CUDA codegen bugs):** neither bug recurred in the new codegen
(the FP8 convert node, cp.async statements and BF16 convert node are all straight-line
single-statement emissions). The one place the patterns could have appeared is the software
`BFloat16` codec, which must compile on all backends: it is written in the FP8.java
kernel-safe style (single return through a result variable to avoid #931; sequential
power-of-two loops on one accumulator instead of branched loops to avoid #930), with comments
citing both issues at each spot. This is the documented workaround, not a root fix — the
phi-lowering (#930) and inliner (#931) fixes remain open follow-ups, deliberately out of this
PR's scope.

#### Problem description

Not a bug fix — this closes the NVIDIA hardware-feature gaps tracked in #940. The concrete
problem it removes: in-kernel FP8 decode was the software arithmetic path from `FP8.java`
(2–10x slower than a hardware convert per the #928 benchmark), MMA GEMM tile loads
round-tripped through registers instead of using Ampere's async copies, and BF16 had no MMA
path at all.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [x] CUDA
- [ ] SPIRV
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestFP8
tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestBFloat16
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMAFP8
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMACpAsync
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMABF16
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMA
```
Benchmarks (reproduce the numbers above):

```bash
# FP8 dequant: hardware vs software decode vs fp16
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.arrays.FP8Benchmark

# GEMM variants: baseline / MMA / MMA+cp.async / MMA bf16 / swizzled
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.compute.MatrixMultiplicationMMA
```


mikepapadim (approved):

mikepapadim:
While integrating this branch on top of `develop` + the cuTENSOR provider for the LLM showcase (#953), I hit two pre-existing build/packaging bugs. Both are fixed in #953; documenting here since anyone stacking on this branch will hit them.

**1. `cusparse-jni` native library never packaged → all cuSPARSE tests silently `UNSUPPORTED`.**
The CMake build produces `libtornado-cusparse.so` under `target/cmake/`, but the `copy-binaries` execution only copied from `target/cmake/build/`, so the `.so` never reached the driver jar and `System.loadLibrary("tornado-cusparse")` failed at runtime — every cuSPARSE test reported `CUDA CONFIGURATION UNSUPPORTED` instead of running.
Fix: add the `${project.build.directory}/cmake` resource dir in `tornado-drivers/cusparse-jni/pom.xml` (+7 lines at pom.xml:82).
Commit `9ca84b1b649894e959df737c7e0a8359bdfa48d5`.

**2. `tornado-cutensor` / `cutensor-jni` poms pinned at `5.0.1-jdk21-dev` → reactor build fails against the 5.1.1 tree.**
`Could not find artifact io.github.beehive-lab:tornado-cutensor:jar:5.1.1-jdk21-dev` — the two cuTENSOR poms carried a stale parent/module version.
Fix: bump `5.0.1-jdk21-dev` → `5.1.1-jdk21-dev` in `tornado-cutensor/pom.xml:9` and `tornado-drivers/cutensor-jni/pom.xml:9`.
Commit `922ef58ef613f03794dac24a37c21a44a5e2fe58`.

After both, the full stack builds clean with CUDA 12.6 and all provider suites pass on an RTX 4090 (cuSPARSE 11/11, cuTENSOR 4/4).

PR #962, opened 2026-07-22
Packed half2 support for the CUDA backend

orionpapadakis:
#### Description

This PR lets `Half2` values stay packed in a single 32-bit `__half2` register through
loads, stores, arithmetic and local-memory staging on the CUDA backend, instead of being
decomposed into scalar half lanes. This halves the memory instructions and local-memory
footprint of FP16 pair work (e.g. attention K/V dot-products and RoPE-style cache writes).

The change comes in three self-contained commits:

1. **Enable FP16 on the CUDA backend.** The CUDA JNI reported only `cl_khr_fp64`, so
   `CUDAFP16SupportPhase` rejected every FP16 vector kernel. It now reports `cl_khr_fp16`
   for compute capability >= 5.3, and `CUDAKind.getCUDATypeName` maps the width-2 half
   vector to the native `__half2`. Wider half vectors (half3/4/8/16) have no CUDA built-in
   and are surfaced as `TornadoDeviceFP16NotSupported` so they are treated as an
   unsupported device rather than a compiler failure.

2. **Packed loads, stores and intrinsics.**
   - `HalfFloatArray.getHalf2/setHalf2` lower to a single 32-bit `*((__half2 *)addr)` access
     (safe because `__half2` is 4-byte aligned and the element index is even); non-CUDA
     backends use the plain-Java pairwise bodies.
   - Vector load/store of `HALF2` and `Half2` add/sub/mult/div emit the packed access and
     `__hadd2/__hsub2/__hmul2/__h2div` (previously the width-2 half path emitted a broken
     `make___half2`).
   - New `Half2.fma/lowFloat/highFloat/fromFloats` lower to
     `__hfma2/__low2float/__high2float/__floats2half2_rn`, bridging FP16 pairs and FP32
     accumulation (attention-style dot products / RoPE-style packed stores).

3. **Packed half2 local memory arrays.** `KernelContext.allocateHalf2LocalArray(int)`
   declares a `__shared__ __half2` array so kernels can stage FP16 pairs in local memory
   without expanding them to FP32. Indexed accesses on the array become packed HALF2 vector
   load/stores. This commit also fixes a latent limitation where local/private vector
   accesses only supported constant indices (the index was string-parsed out of the LIR
   value), so a runtime index such as `threadIdx.x` was silently mis-emitted.

#### Problem description

Not a bug fix for a reported issue. Before this PR the CUDA backend could not keep FP16
pairs packed: `cl_khr_fp16` was never reported (all FP16 vector kernels were rejected), the
width-2 half vector had no native type mapping, and vector arithmetic on halves was emitted
componentwise. The packed path removes those limitations and the extra scalar traffic they
caused.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [x] CUDA
- [ ] SPIRV
- [ ] Metal

The new codegen is CUDA-only. The API additions (`HalfFloatArray.getHalf2/setHalf2`,
`Half2` helpers, `KernelContext.allocateHalf2LocalArray`) have plain-Java fallback bodies,
so they build on other backends but are not accelerated there.

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

Ubuntu, JDK 21, RTX 5090 (sm_120), CUDA backend.

#### Did you check on FPGAs?

- [ ] Yes
- [x] No

```bash
make BACKEND=cuda
source setvars.sh

# Functional: packed copy/add/mult/fma, FP32-accumulated row dot, packed RoPE store,
# and a __shared__ __half2 local tile behind a barrier.
tornado-test --verbose uk.ac.manchester.tornado.unittests.vectortypes.TestHalf2Packed

# Codegen: compiles kernels through the CUDA JIT and asserts the generated CUDA C keeps
# the packed accesses (((__half2 *)...), __shared__ __half2) and the cuda_fp16.h intrinsics
# (__hfma2, __hadd2, __hmul2, __floats2half2_rn, __low2float, __high2float), so the codegen
# cannot silently regress to scalar half code. Also wired into test-native.sh.
tornado uk.ac.manchester.tornado.drivers.cuda.tests.TestHalf2PackedCodegen
```

Both suites pass on the CUDA backend (`TestHalf2Packed`: 7/7; `TestHalf2PackedCodegen`:
Test success).

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/HalfFloatArray.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/HalfFloatArray.java
@@ -20,6 +20,7 @@ package uk.ac.manchester.tornado.api.types.arrays;
 import uk.ac.manchester.tornado.api.annotations.Parallel;
 import uk.ac.manchester.tornado.api.internal.annotations.SegmentElementSize;
 import uk.ac.manchester.tornado.api.types.HalfFloat;
+import uk.ac.manchester.tornado.api.types.vectors.Half2;
 
 import java.lang.foreign.MemorySegment;
 import java.util.Arrays;
@@ -234,6 +235,34 @@ public final class HalfFloatArray extends TornadoNativeArray {
     public HalfFloat get(int index) {
         short halfFloatValue = segment.getShortAtIndex(index, baseIndex);
         return new HalfFloat(halfFloatValue);
+    }
+
+    /**
+     * Gets two consecutive {@link HalfFloat} values starting at the specified element index as a packed {@link Half2}.
+     * On backends with packed half2 support this maps to a single 32-bit load; {@code index} must be even so the
+     * access is 4-byte aligned.
+     *
+     * @param index
+     *         The element index of the first lane; must be even.
+     * @return A {@link Half2} holding elements {@code index} and {@code index + 1}.
+     */
+    public Half2 getHalf2(int index) {
+        return new Half2(get(index), get(index + 1));
+    }
+
+    /**
+     * Stores a {@link Half2} into two consecutive elements starting at the specified element index.
+     * On backends with packed half2 support this maps to a single 32-bit store; {@code index} must be even so the
+     * access is 4-byte aligned.
+     *
+     * @param index
+     *         The element index of the first lane; must be even.
+     * @param value
+     *         The {@link Half2} whose lanes are stored at {@code index} and {@code index + 1}.
+     */
+    public void setHalf2(int index, Half2 value) {
+        set(index, value.getX());
+        set(index + 1, value.getY());
     }
 
     /**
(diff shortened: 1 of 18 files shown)


tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/HalfFloatArray.java
```diff
@@ -236,6 +237,34 @@ public HalfFloat get(int index) {
...
+    /**
+     * Gets two consecutive {@link HalfFloat} values starting at the specified element index as a packed {@link Half2}.
+     * On backends with packed half2 support this maps to a single 32-bit load; {@code index} must be even so the
+     * access is 4-byte aligned.
+     *
+     * @param index
+     *         The element index of the first lane; must be even.
+     * @return A {@link Half2} holding elements {@code index} and {@code index + 1}.
+     */
+    public Half2 getHalf2(int index) {
```
mikepapadim, line 249: we should guard that index < size

tornado-api/src/main/java/uk/ac/manchester/tornado/api/types/arrays/HalfFloatArray.java
```diff
@@ -236,6 +237,34 @@ public HalfFloat get(int index) {
...
+     * On backends with packed half2 support this maps to a single 32-bit store; {@code index} must be even so the
+     * access is 4-byte aligned.
+     *
+     * @param index
+     *         The element index of the first lane; must be even.
+     * @param value
+     *         The {@link Half2} whose lanes are stored at {@code index} and {@code index + 1}.
+     */
+    public void setHalf2(int index, Half2 value) {
+        set(index, value.getX());
```
mikepapadim, line 270: same here - guard

mikepapadim (approved):

PR #982, opened 2026-07-25
CUDA stream usage: visible routing, DAG-based stream selection, plan-level tuning, examples and measurements

mikepapadim:
#### Description

Follow-up work on CUDA stream usage after #800 (intra-plan concurrency) and #936 (staged H2D
transfers), both now on `develop`. This PR makes stream routing observable, fixes two ordering defects,
makes kernel-to-stream assignment dependency-aware and reproducible, exposes the tuning knobs on the
execution plan, and adds the examples and docs the feature was missing. It also reports two
optimizations that were implemented, measured, and then **dropped** because the measurements did not
support them — the data is in the branch so the decision can be re-litigated with numbers rather than
intuition.

One commit per change, each with tests:

| commit | what |
|---|---|
| `[cuda] Show the stream each bytecode was issued to in --printBytecodes` | `[stream=COMPUTE_2]` on every transfer/launch line |
| `[cuda] Honour the wait list in clEnqueueMarkerWithWaitList` | JNI marker ignored its wait list |
| `[cuda] Pick COMPUTE streams from the task DAG instead of blind round-robin` | chains stay on one stream; assignment is deterministic per execution |
| `[api] Add withIntraPlanConcurrency(int) and withStagedTransfers(minSize, chunkSize, ringDepth)` | tuning is no longer `-D`-only |
| `[examples] Add stream and staged-transfer examples under examples/streams` | 4 runnable examples (there were none) |
| `[docs] Document intra-plan concurrency, staged transfers and their JVM flags` | `programming.rst`, `flags.rst`, `withCUDAGraph()` javadoc |
| `[docs] Add the CUDA stream measurements behind these changes` | `docs/perf/cuda-streams/` |
| `[skill] Add a streams and concurrency section to the tornadovm-nvidia skill` | + `references/streams-concurrency.md` |

#### Problem description

Not a bug fix in the main; a mix of one real defect, one diagnostic gap, and one scheduling improvement.

**1. Stream routing was invisible.** `--printBytecodes` printed the opcode, object, size, batch, offset
and `[event list=N]`, but never which queue the operation went to (`DebugInterpreter`), so the only way
to see how a plan was routed was to read an Nsight timeline. Every operation now reports its stream:

```
bc:  TRANSFER_HOST_TO_DEVICE_ALWAYS  [0x2c5529ab] FloatArray@2c5529ab on [NVIDIA CUDA] -- RTX 4090,
     size=25165840, batchSize=0, offset=0 [event list=-1] [stream=DATA_TRANSFER_H2D]
bc:  LAUNCH  task overlap.u3 - compute on ..., numThreadBatch=0, offset=0 [event list=2] [stream=COMPUTE_3]
```

Backed by `TornadoDevice#getLastQueueLabel(long)` (default `""`, so other backends' output is
unchanged), which also lets tests assert routing without scraping stdout. Two log calls moved from
before to after their device call, because the stream is only known once the operation has been routed
(`LAUNCH`, `TRANSFER_DEVICE_TO_HOST_ALWAYS*`); the launch log is in a `finally` so a bailout still logs.
The missing `CUDA` case in `InterpreterUtilities#debugDeviceBC` is filled in.

**2. `clEnqueueMarkerWithWaitList` dropped its wait list.** The JNI entry point took the array and
called `record_event` without ever calling `wait_events`, so a marker with dependencies completed
immediately (`cuda-jni/.../CUDACommandQueue.cpp`). In practice nothing broke, because
`CUDADeviceContext#enqueueMarker` also joins the plan's dirty role queues into the default queue first,
and that join provided the ordering — so this is defence in depth, not a live bug, and I could not write
a test that fails without it. The new
`TestCUDAStreams#testLibraryTaskOrderingUnderConcurrency` covers the path (a slow producer kernel feeding
a cuBLAS `sgemv` under concurrency) and would catch a regression that removed both guards.

**3. COMPUTE stream assignment was dependency-blind and non-reproducible.**
`nextComputeIndex` was `cursor.getAndIncrement() % poolSize` with no per-execution reset. Two
consequences: a dependency chain was split across streams (paying a `cuStreamWaitEvent` that buys
nothing, since an in-order stream already orders it), and the mapping rotated with the number of kernels
issued so far, so two profiles of the same plan disagreed. Now the producer streams in a kernel's
dependency list decide: if they are all the same COMPUTE stream, the kernel is issued there; otherwise
it takes the next stream round-robin. The cursor is reset per execution and the pool is created in full
on first use, so stream identities are stable.

The dependency list is used for *selection only* — the launch's wait list is unchanged (the
argument-write event). That distinction is the whole point; see the rejected variant below.

**4. Tuning was `-D`-only.** `withIntraPlanConcurrency(int computeStreams)` and
`withStagedTransfers(long minTransferSize, long chunkSize, int ringDepth)` thread through the existing
plan-option path (`TornadoExecutor` → `ImmutableTaskGraph` → `TaskGraph` → `TornadoTaskGraph` →
`TornadoExecutionContext`), are validated (`>= 1`, ring depth `>= 2`), and are honoured per plan by the
CUDA backend — the COMPUTE pool size stops being `static final`, and changing the ring geometry rebuilds
the pinned block. Other backends ignore the values through defaulted interface methods.

**5. No examples, no docs.** Nothing in `tornado-examples` used `withIntraPlanConcurrency`,
`withStagedTransfers`, `withCUDAGraph` or `withProfiler`, and nothing in `docs/source` mentioned them
(#800 shipped an example that was dropped when the API moved off `TaskGraph.withCUDAStreams()`). Added
`examples/streams/`: `MultiStreamOverlap`, `ConcurrentKernelsPoolSweep`,
`BulkUploadWithIndependentKernels`, `StagedWeightUpload`, plus a `programming.rst` section, the
`tornado.staged.*` / `tornado.cuda.compute.streams` flags, and a javadoc for `withCUDAGraph()` (it had a
`//TODO`) documenting that capture forces single-stream and suppresses device-side profiler timings.

#### Measurements

Full data, CSVs and reproduction commands: **`docs/perf/cuda-streams/README.md`**.
RTX 4090, driver 565.57.01, CUDA 11.5, nsys 2024.5.1, JDK 21.0.2, baseline `develop @ 06616ddd4`.

**Throughput: parity, not a win.** `TestStreamsPerformance`, median of 10, three repetitions:

| scenario | develop, single | develop, concurrent | branch, single | branch, concurrent |
|---|---|---|---|---|
| transfer/compute overlap | 32.33 ms | 28.89 ms | 32.45 ms | 28.96 ms |
| compute concurrency, large kernels | 43.55 ms | 37.45 ms | 43.73 ms | 37.07 ms |
| compute concurrency, small kernels | 4.28–4.30 ms | 1.27 / 1.30 / 1.29 ms | 4.27–4.29 ms | 1.32 / 1.33 / 1.34 ms |

The ~3% on the small-kernel row is inside this box's spread (develop itself produced 1.16 ms once and
1.27–1.30 ms on three later repetitions). I am not claiming a speedup from the scheduling change: what
it buys is correct placement and reproducibility.

**Placement: this is what changed.** `TestCUDAStreams#testIndependentChainsUnderConcurrency`, two
independent two-kernel chains, from `nsys stats --report cuda_gpu_trace`:

<details><summary>GPU trace (stream ids) and API counters</summary>

```
  start      0.0 us  dur 1.2 us  stream 16  scale     <- chain a, kernel 1
  start    199.6 us  dur 1.2 us  stream 16  scale     <- chain a, kernel 2  (same stream)
  start    395.0 us  dur 1.2 us  stream 17  scale     <- chain b, kernel 1
  start    590.5 us  dur 1.2 us  stream 17  scale     <- chain b, kernel 2  (same stream)

  cuLaunchKernel       calls=   12 total=    0.08 ms
  cuEventRecord        calls=   54 total=    0.02 ms
  cuStreamWaitEvent    calls=   36 total=    0.02 ms
  cuCtxSynchronize     calls=   12 total=    0.01 ms
```

and the same from `--printBytecodes`, identical across all three executions:

```
task chains.a1 - scale [stream=COMPUTE]
task chains.a2 - scale [stream=COMPUTE]
task chains.b1 - scale [stream=COMPUTE_1]
task chains.b2 - scale [stream=COMPUTE_1]
```
</details>

**Pool sizing works per plan** (`--printBytecodes` over `testComputeStreamPoolSizeIsHonoured`, which
runs the same graph at pool sizes 1, 2 and 8): 8 launches on `COMPUTE`, then alternating
`COMPUTE`/`COMPUTE_1`, then `COMPUTE`..`COMPUTE_7`. And the example sweep on 8 small independent kernels:

```
single stream                  2.16 ms      -
concurrency, pool=1            1.89 ms   1.14x
concurrency, pool=2            1.15 ms   1.88x
concurrency, pool=4            1.03 ms   2.11x
concurrency, pool=8            0.89 ms   2.42x
```

#### Two optimizations measured and dropped

Both were on the plan for this PR. I implemented enough of each to measure it, and the numbers said no.

**(a) Kernel-argument writes on their own stream + full dependency list on the launch.** The premise:
argument writes share the bulk H2D stream, so an independent kernel cannot launch until a multi-megabyte
copy ahead of it drains. Implemented (a `KERNEL_ARGS` role stream, launches carrying their real producer
events) and **it regressed `testSmallKernelConcurrency` from 1.30 ms to 1.68–1.79 ms**, ~35%. Cause: once
the argument write no longer shares a stream with the transfers it depends on,
`CUDAEventPool#serialiseEvents` can no longer drop those waits as same-stream, so each launch pays one
`cuStreamWaitEvent` per producer instead of exactly one. Separately, `nsys` showed the premise mostly
does not bite: the bytecode order that the graph compiler emits already places argument writes ahead of
the bulk copy for the shapes I could build, and `BulkUploadWithIndependentKernels` measured 1.06x on
develop vs 1.08x with the change — noise. Reverted; the DAG-based selection was kept by passing the
dependency list as a selection hint that adds no waits. The one piece of that work worth keeping is in:
`CUDAInstalledCode`'s shared `int[1] internalEvents` field is now a local array (it was mutated per
launch and handed to a queue, which is not safe with more than one queue in play).

**(b) Stream-ordered device allocation (`cuMemAllocAsync` + `cuMemPool*`).** The stated motivation was
removing the device-wide `cuCtxSynchronize` that guards every `cuMemFree` and every
`cuMemHostUnregister` (`cuda-jni/.../CUDAContext.cpp`). Measured first, and those guards cost **0.20 ms
across 404 calls (0.5 µs each)** over the whole `TestStreamsPerformance` run — the device is already idle
when buffers are freed:

<details><summary>cuda_api_sum, whole TestStreamsPerformance run</summary>

| API | calls | total | note |
|---|---|---|---|
| `cuStreamSynchronize` | 654 | 1760.84 ms | waiting for real GPU work at plan end |
| `cuMemHostRegister_v2` | 202 | 128.51 ms | page-locking transfer sources |
| `cuMemHostUnregister` | 202 | 63.83 ms | driver page-unlock work |
| `cuMemFree_v2` | 202 | 61.47 ms | driver unmap work |
| `cuCtxSynchronize` | 404 | **0.20 ms** | the guards a memory pool would remove |
| `cuStreamWaitEvent` | 2652 | 0.59 ms | cross-stream ordering |
| `cuEventCreate`+`Record`+`Destroy` | 15460 | 2.86 ms | event bookkeeping |

</details>

The 125 ms in free plus unregister is the driver's own unmap and page-unlock work at **plan teardown**,
not the guard, so a pool would be optimizing plan churn, not steady-state execution — while taking on
stream-ordered allocation lifetimes on the critical path (allocation has no stream at
`CUDABufferProvider`, and a stream-ordered free is only safe once every consumer stream is ordered
after it). Not worth that trade on this evidence. If wanted as a follow-up, the design I would use is
cuda-oxide's: store the owning stream on the buffer so an async-pool pointer can never be freed with the
sync free (`crates/cuda-core/src/device_buffer.rs:136-168`), plus its event-gated reclamation limbo
(`crates/cuda-async/src/reclaim.rs`) for frees that cannot yet be proven safe.

Related, same reasoning: **asynchronous staging-slot retirement** (replacing the per-slot
`cuEventSynchronize` with `cuLaunchHostFunc`) is not here either, because the staging ring's cost on this
box is the host memcpy, not the slot wait — 1 GB of warm weights uploads at 16.6–17.0 GB/s direct versus
10.4 GB/s staged. Staged still wins for the case #936 targeted (large cold mmap'd sources, where the
whole-segment page-lock dominates); `StagedWeightUpload` documents both sides so nobody reads the ring as
a universal win.

#### Other findings, reported not fixed

- `withCUDAGraph()` and `withIntraPlanConcurrency()` are mutually exclusive in practice: capture records
  one stream and `isMultiStreamEnabled` ANDs `!isStreamCapturing`, so a captured plan silently runs
  single-stream. Now documented in the javadoc and the skill; making it work needs multi-stream capture
  (`CU_STREAM_CAPTURE_MODE_RELAXED` fork/join, or `cuStreamBeginCaptureToGraph`).
- With the profiler on, the interpreter waits on nearly every operation's event, so overlap disappears;
  any concurrency measurement has to run without it.
- Plan options are held on the execution context shared through the `ImmutableTaskGraph`, so a second
  `TornadoExecutionPlan` built from the same snapshot inherits the first plan's concurrency and staging
  settings. `testStreamRoutingIsReported` documents this by calling `withoutIntraPlanConcurrency()`
  explicitly for its single-stream case.
- `CUDACommandQueueTable` keys the default queue by thread id, so one plan driven from two Java threads
  gets two default streams and `reset()` cleans only the calling thread's entry.
- `CUDAEventPool#serialiseEvents` writes into `waitEventsBuffer` without bounds-checking against
  `MAX_WAIT_EVENTS` (64).

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [x] CUDA
- [ ] Metal

CUDA-only: the PTX backend has its own copy of this machinery and is untouched. The runtime-side changes
(`TornadoDevice#getLastQueueLabel` default, `DebugInterpreter` log position, execution-context fields)
are no-ops for other backends, and OpenCL/Metal bytecode output is byte-identical because the stream tag
is only appended when a backend reports a label.

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
source setvars.sh

# correctness: 19 tests, 6 of them new (routing, chains, bulk-upload dependency, library ordering,
# pool sizing, staging tuning)
tornado-test --ea -V uk.ac.manchester.tornado.unittests.streams.TestCUDAStreams

# printed, not asserted
tornado-test --ea uk.ac.manchester.tornado.unittests.streams.TestStreamsPerformance

# routing, per operation
tornado --printBytecodes -m tornado.examples/uk.ac.manchester.tornado.examples.streams.MultiStreamOverlap

# examples
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.streams.ConcurrentKernelsPoolSweep
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.streams.StagedWeightUpload
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.streams.BulkUploadWithIndependentKernels

# timeline + API counters
nsys profile --trace=cuda,nvtx -o run tornado -m tornado.examples/uk.ac.manchester.tornado.examples.streams.MultiStreamOverlap
nsys stats --report cuda_gpu_trace --report cuda_api_sum --format table,csv --output . run.nsys-rep
```

`tornado-test --quickPass`: 87 failing tests on this branch versus 86 on `develop @ 06616ddd4`, same 22
classes. The single difference is `compute.MMwithBytes`, which is flaky on both — it failed 3 of 10 runs
on develop and 3 of 12 on the branch with `Output contains NaN`.
`compute-sanitizer --tool memcheck` reports 0 errors on the four new ordering tests
(`testStreamRoutingIsReported`, `testIndependentChainsUnderConcurrency`,
`testKernelWaitsForBulkUploadUnderConcurrency`, `testLibraryTaskOrderingUnderConcurrency`).


mikepapadim:
@orionpapadakis mostly cleanups and api usage simplifications for stream management, plus mores examples

PR #984, opened 2026-07-26
Add warp-vote intrinsics (simdAny, simdAll, simdBallot) to KernelContext on the CUDA backend

mikepapadim:
#### Description

Adds warp-vote intrinsics to `KernelContext` on the CUDA backend: `simdAny`, `simdAll` and
`simdBallot`, lowered to `__any_sync`, `__all_sync` and `__ballot_sync` over the full member mask.

These complete the warp-level primitive set the backend already had. `simdSum`, `simdShuffleDown` and
`simdBroadcastFirst` move *data* between lanes; there was no way to ask a *question* of the warp. Vote
is what makes lane-cooperative patterns expressible without a shared-memory round trip:

- **warp-aggregated atomics** — one atomic per warp instead of one per lane, which is the standard fix
  for atomics-bound kernels (histograms, appends, compaction). `Integer.bitCount(simdBallot(p))` is the
  group's count and `Integer.bitCount(simdBallot(p) & ((1 << lane) - 1))` is a lane's rank within it, so
  one lane reserves space for the whole warp.
- **early exit** — `if (!ctx.simdAny(needsWork))` skips a block that no lane in the warp needs, with
  every lane agreeing on the answer.
- **uniformity checks** — `simdAll(...)` to take a fast path only when it holds for the whole warp.

`Integer.bitCount` already lowers to `__popc`, so the aggregation idiom composes with no further work.

#### API

```java
public boolean simdAny(boolean predicate)    // true if any active lane passes
public boolean simdAll(boolean predicate)    // true if every active lane passes
public int     simdBallot(boolean predicate) // bit i set when lane i passes
```

Same shape and conventions as the existing `simd*` family: the `KernelContext` bodies answer for a
single lane (`simdAny`/`simdAll` return the predicate, `simdBallot` returns `p ? 1 : 0`) so sequential
Java execution of the same kernel still works, and the CUDA graph-builder plugin replaces the call with
the real intrinsic. Backends without a lowering keep the single-lane default, which is why the tests skip
them explicitly rather than asserting wrong answers.

Implementation follows `CUDASimdBroadcastFirstNode` / `CUDALIRStmt.ShuffleSyncStmt` exactly:

- `CUDAWarpVoteNode` — one node, mode enum (`ANY`/`ALL`/`BALLOT`), stamp `Boolean` or `Int` per mode.
  `FixedWithNextNode` because votes are convergent and must not be reordered or duplicated across
  control flow.
- `CUDALIRStmt.WarpVoteStmt` — emits the intrinsic with the `0xffffffff` member mask; `ANY`/`ALL` wrap
  the result in `(... != 0)` since the intrinsics return `int` and the Java result is `boolean`.
- `CUDAGraphBuilderPlugins#registerSIMDPlugins` — three `InvocationPlugin` registrations.

Generated CUDA-C, from `--printKernel`:

```c
// simdBallot(localIdx % 2 == 0)
i_3  =  __ballot_sync(0xffffffff, i_2);

// simdAny(localIdx == 7) / simdAll(localIdx == 7)
i_3  =  (__any_sync(0xffffffff, i_2) != 0);
i_4  =  (__all_sync(0xffffffff, i_2) != 0);

// Integer.bitCount(mask & ((1 << lane) - 1)) — warp-aggregated rank
i_14 =  __popc(i_13);
```

#### Problem description

Feature, not a bug fix. Before this, a kernel that needed any warp-wide decision had to fake it through
a local array plus `localBarrier()`, which costs shared memory and a block-wide barrier for something the
hardware answers in one instruction — and it is not even equivalent, since a barrier synchronises the
whole block rather than the warp.

#### Backend/s tested

- [ ] OpenCL
- [ ] PTX
- [x] CUDA
- [ ] Metal

CUDA-only by design. The API and the single-lane defaults are backend-neutral, so nothing changes for
other backends, and `TestWarpVote` skips OpenCL, SPIR-V, Metal and PTX rather than asserting against a
lowering that does not exist yet. Natural follow-ups if wanted: PTX (`vote.sync.any/all/ballot.b32`),
Metal (`simd_any` / `simd_all` / `simd_ballot`), OpenCL (`sub_group_any` / `sub_group_all`, no ballot in
core).

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
source setvars.sh

# all four tests
tornado-test --ea -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote
```

Each test asserts a value a single lane cannot produce, so a missing or wrong lowering fails rather than
returning the caller's own predicate.

**`simdBallot`** — every lane votes "my lane index is even"; all 32 lanes of every warp must read back
the same even-lane mask `0x55555555`:

```bash
tornado-test --ea -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testBallotEvenLanes
tornado --printKernel -ea -m tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner \
        --params "uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testBallotEvenLanes" \
  | grep __ballot_sync
# i_3  =  __ballot_sync(0xffffffff, i_2);
```

**`simdAny` and `simdAll`** — only lane 7 holds a true predicate, so `any` must be true on all 32 lanes
(including the 31 whose own predicate is false) and `all` false on all of them; encoded as `any*2 + all`,
asserted as `2`:

```bash
tornado-test --ea -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testAnyAndAll
tornado --printKernel -ea -m tornado.unittests/uk.ac.manchester.tornado.unittests.tools.TornadoTestRunner \
        --params "uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testAnyAndAll" \
  | grep -E "_any_sync|_all_sync"
```

**`simdAll`, true everywhere** — the complementary case, so a lowering that always returned false would
not pass:

```bash
tornado-test --ea -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testAllTrue
```

**Warp-aggregated rank** — the pattern the feature exists for: lanes where `localIdx % 3 == 0` pass, and
each lane derives its rank among them from the ballot mask with no atomic and no shared memory. Checked
against a CPU reference per lane:

```bash
tornado-test --ea -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote#testWarpAggregatedLaneRank
```

Regression check (RTX 4090, driver 565.57.01, CUDA 11.5, JDK 21.0.2): `tornado-test --quickPass` reports the same
86 pre-existing failures on this branch as on `develop @ 06616ddd4` (same 22 classes, unrelated CUDA
backend gaps: vector types, prebuilt kernels, virtual layer, Metal-only simdgroup tests) — the new tests are additive and the plugin registrations only fire
for the three new `KernelContext` methods. `mvn checkstyle:check` clean.

Usage shape, for reviewers who want to see what the intrinsics are for — the ballot/rank half is exactly
what `testWarpAggregatedLaneRank` exercises:

```java
int lane  = ctx.localIdx % 32;
boolean keep = predicate(input.get(ctx.globalIdx));
int mask  = ctx.simdBallot(keep);            // which lanes of my warp keep their element
int count = Integer.bitCount(mask);          // how many, no atomic involved
int rank  = Integer.bitCount(mask & ((1 << lane) - 1));  // my slot within the group

if (ctx.simdAny(keep)) {                     // whole warp agrees; skip the block otherwise
    // reserve `count` slots once per warp (one atomic instead of 32), then each keeping
    // lane writes at base + rank
}
```

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/KernelContext.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/KernelContext.java
@@ -383,6 +383,54 @@ public class KernelContext implements ExecutionContext {
     }
 
     /**
+     * Returns whether {@code predicate} holds for at least one active lane of the SIMD group.
+     * <p>
+     * Warp/SIMD-group vote. Useful to skip work that no lane in the group needs, without a
+     * shared-memory round trip: every lane gets the same answer.
+     * <p>
+     * CUDA equivalent: {@code __any_sync(0xffffffff, predicate)}
+     *
+     * @param predicate
+     *     per-lane condition
+     * @return true when any active lane passes {@code predicate}
+     */
+    public boolean simdAny(boolean predicate) {
+        return predicate;
+    }
+
+    /**
+     * Returns whether {@code predicate} holds for every active lane of the SIMD group.
+     * <p>
+     * CUDA equivalent: {@code __all_sync(0xffffffff, predicate)}
+     *
+     * @param predicate
+     *     per-lane condition
+     * @return true when all active lanes pass {@code predicate}
+     */
+    public boolean simdAll(boolean predicate) {
+        return predicate;
+    }
+
+    /**
+     * Returns a bit mask with one bit per lane of the SIMD group, set where {@code predicate} holds.
+     * Bit {@code i} corresponds to lane {@code i}, so on a 32-lane warp the result covers bits 0-31.
+     * <p>
+     * This is the building block for warp-aggregated atomics and stream compaction:
+     * {@code Integer.bitCount(simdBallot(p))} counts the lanes that pass without any atomic, and
+     * {@code Integer.bitCount(simdBallot(p) & ((1 << laneId) - 1))} gives each lane its rank among
+     * them, so one lane can reserve space for the whole group with a single atomic.
+     * <p>
+     * CUDA equivalent: {@code __ballot_sync(0xffffffff, predicate)}
+     *
+     * @param predicate
+     *     per-lane condition
+     * @return lane mask of the lanes that pass {@code predicate}
+     */
+    public int simdBallot(boolean predicate) {
+        return predicate ? 1 : 0;
+    }
+
+    /**
      * Cooperative 8x8 single-precision matrix multiply for one SIMD group, using
      * Apple's {@code simdgroup_float8x8} hardware matrix units (MMA).
      * <p>
--- a/tornado-assembly/src/bin/tornado-test
+++ b/tornado-assembly/src/bin/tornado-test
@@ -207,6 +207,7 @@ __TEST_THE_WORLD__ = [
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestLocalArrayDynamicIndexAtomics"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestAsyncCopyToLocal"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestFlashAttentionKernelContext"),
+    TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestWarpVote"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestHalfFloatLocalArray"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.local.memory.TestSwizzledLocalArrays"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationKernelContext"),
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/plugins/CUDAGraphBuilderPlugins.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/plugins/CUDAGraphBuilderPlugins.java
@@ -96,6 +96,7 @@ import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAShuffleDownNode;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDASimdBroadcastFirstNode;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAAtomicRmwNode;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDASimdSumNode;
+import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAWarpVoteNode;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.DecAtomicNode;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.GetAtomicNode;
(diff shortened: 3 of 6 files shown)


mikepapadim:
@stratika

mikepapadim:
Rebased onto `develop` @`ba28f1526` (the branch was 95 commits behind). Three conflicts, all resolved as additions:

- `KernelContext.java` — `develop` added the local-array atomics (`atomicCAS`/`atomicExchange`/`atomicMin`/`atomicMax`) in the same place this PR adds the warp-vote methods. Both kept.
- `CUDALIRStmt.java` — `develop`'s new `AtomicRmwStmt` class overlapped structurally with this PR's `WarpVoteStmt`, so git interleaved them. Rebuilt by taking `develop`'s file and re-applying `WarpVoteStmt` intact.
- `bin/tornado-test` — both sides append to the test list; both entries kept.

One follow-up fix the rebase made necessary: `skipUnsupportedBackends()` referenced `TornadoVMBackendType.SPIRV` and `.PTX`, and both enum constants are gone from `develop` now. Dropped those two guards, leaving `OPENCL` and `METAL`.

Verified on RTX 4090 / JDK 21.0.2:
- `make BACKEND=cuda` green.
- `tornado-test -V ...kernelcontext.api.TestWarpVote`: **4 ran, 0 failed** (`testBallotEvenLanes`, `testAnyAndAll`, `testAllTrue`, `testWarpAggregatedLaneRank`).
- `tornado-test -V --quickPass`: **1035 PASS / 9 FAILED**, against a `develop` baseline of 1031 / 9 on the same machine — the four extra passes are this PR's tests, and the failure set is unchanged.

mikepapadim:
Pushed `c33882230`: four more `TestWarpVote` cases, chosen so each one fails for a *different* wrong lowering rather than just adding coverage.

| test | what it pins down |
|---|---|
| `testMultipleWarpsPerBlock` | 128-thread block = four warps voting differently from each other. The existing tests all use a one-warp block, which cannot tell a warp-wide vote from a block-wide one; a block-wide reduction answers 2 (any but not all) everywhere instead of 3 / 0 per warp. |
| `testDataDependentBallot` | Predicate read from device memory, so the mask has to be built at runtime rather than constant-folded. Each lane reports `bitCount` of its warp's mask and must agree with every other lane in the warp. |
| `testVoteInsideLoop` | Votes issued from inside a loop, one driving a branch and one feeding arithmetic. The `simdAll` branch is the discriminator: a lowering that returned the calling lane's own predicate would fire it on the single step where that lane passes, giving 136 instead of 36. |
| `testWarpLeaderElection` | Ballot over a compound `&&` predicate (data test plus lane test), so it must be materialised as one boolean. Leader is the lowest passing lane, derived from the lowest set bit with `bitCount` — `Integer.numberOfTrailingZeros` is not registered as an intrinsic on this backend, `bitCount` is. |

Verified on RTX 4090 / JDK 21.0.2:

- `TestWarpVote`: **8 ran, 0 failed** (was 4)
- `tornado-test -V --quickPass`: **1040 PASS / 8 FAILED**, against `develop`'s 1031 / 9 on the same machine. Eight of the extra passes are this PR's tests; the ninth is a whitelisted test that flips between runs, which also accounts for 9 → 8 failures. No new failure.
- `./mvnw checkstyle:check` clean

Test-only change — no production code touched since the rebase.

stratika (approved):

PR #997, opened 2026-07-29
Take fewer map and monitor operations on the interpreter's per-object hot path

mikepapadim:
Two small reductions in per-object work on the interpreter's hot path. Both are semantically neutral;
neither produced a measurable speedup on the workload that motivated them, which I want to be upfront
about — they are here because they are strictly less work in code paths executed hundreds of times per
frame, not because they moved a benchmark.

## What the profile showed

Profiling KFusion on the CUDA backend (`beehive-lab/kfusion-tornadovm#42`) — a pipeline of 8 small
task-graphs with ~14 plan executions per frame — JFR attributed **42.9% of JVM execution samples** to:

```
ConcurrentHashMap.get / containsKey
  <- DataObjectState.getDeviceBufferState
  <- TornadoVMInterpreter.resolveObjectState
  <- TornadoVMInterpreter.executeDeAlloc
```

## The changes

1. `DataObjectState.getDeviceBufferState()` did `containsKey` + `put` + `get` — three map operations
   where one suffices. It is now a single `computeIfAbsent`. This method is called for every object of
   every ALLOC / COPY / DEALLOC bytecode.

2. `TornadoVMInterpreter.executeDeAlloc()` now returns early when the buffer is locked. With the
   default `tornado.reuse.device.buffers=true` every buffer is locked, so `deallocate()` returns
   immediately anyway — but the caller still paid two batch-map lookups and a `synchronized` device call
   per object per execution to find that out. The logging path is preserved.

## Measurements

RTX 4090, CUDA 11.5, JDK 21, KFusion on ICL-NUIM `living_room_traj2_loop`, 882 frames:

| | FPS | computation/frame |
|---|---|---|
| before | 225.8 | 3.944 ms |
| after | 225.6 – 228.6 | 3.923 – 3.932 ms |

i.e. **within run-to-run noise**. The JFR attribution turned out to be misleading: those frames are
cheap in wall-clock terms next to the `cuStreamSynchronize` calls they sit beside. For that workload
the real cost is host↔device serialization — 68 stream synchronizations and 112 H2D copies per frame,
against ~0.7 ms of actual kernel time — which is a separate problem (CUDA graph capture).

If the project would rather not carry changes without a demonstrated win, feel free to close this; the
correctness fix in the companion PR is the one that matters.

## Regression check

`tornado-test --quickPass` on an `opencl,cuda` build:

| | tests run | failed |
|---|---|---|
| `develop` | 1026 | 230 |
| this branch | 1026 | 230 |

The 230 pre-existing failures are the CUDA-library and MMA suites running against the OpenCL default
backend of a combined build, identical before and after. Per-class diffing flagged
`branching.TestLoopConditions` and `compute.MMwithBytes` by one test each; both pass on re-run on this
branch and `TestLoopConditions` also fails intermittently on `develop`, so both are pre-existing
flakes.

### DIFF
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/DataObjectState.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/DataObjectState.java
@@ -35,6 +35,12 @@ public class DataObjectState implements ObjectState {
 
     private ConcurrentHashMap<TornadoXPUDevice, XPUDeviceBufferState> deviceStates;
 
+    /** Last returned pair, as one holder so readers always observe a consistent (device, state). */
+    private volatile CachedDeviceState cachedDeviceState;
+
+    private record CachedDeviceState(TornadoXPUDevice device, XPUDeviceBufferState state) {
+    }
+
     public DataObjectState() {
         deviceStates = new ConcurrentHashMap<>();
     }
@@ -44,7 +50,14 @@ public class DataObjectState implements ObjectState {
         if (!(device instanceof TornadoXPUDevice)) {
             throw new TornadoRuntimeException("[ERROR] Device not compatible: " + device.getClass());
         }
-        return deviceStates.computeIfAbsent((TornadoXPUDevice) device, key -> new XPUDeviceBufferState());
+        // Objects are almost always used on the same device twice in a row.
+        final CachedDeviceState cached = cachedDeviceState;
+        if (cached != null && cached.device() == device) {
+            return cached.state();
+        }
+        final XPUDeviceBufferState deviceState = deviceStates.computeIfAbsent((TornadoXPUDevice) device, key -> new XPUDeviceBufferState());
+        cachedDeviceState = new CachedDeviceState((TornadoXPUDevice) device, deviceState);
+        return deviceState;
     }
 
     @Override
@@ -61,6 +74,7 @@ public class DataObjectState implements ObjectState {
     @Override
     public void clear() {
         deviceStates.clear();
+        cachedDeviceState = null;
     }
 
     @Override
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/DataObjectState.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/DataObjectState.java
@@ -44,10 +44,10 @@ public class DataObjectState implements ObjectState {
         if (!(device instanceof TornadoXPUDevice)) {
             throw new TornadoRuntimeException("[ERROR] Device not compatible: " + device.getClass());
         }
-        if (!deviceStates.containsKey(device)) {
-            deviceStates.put((TornadoXPUDevice) device, new XPUDeviceBufferState());
-        }
-        return deviceStates.get(device);
+        // One map operation instead of containsKey + put + get: this is on the hot path of every
+        // ALLOC/COPY/DEALLOC bytecode, and for a pipeline that runs many small task-graphs per frame it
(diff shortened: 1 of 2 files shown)


mikepapadim:
@orionpapadakis do you think can help for llama? can we profile with JFR gpullama3 to see if we have the same behavior there?

tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/DataObjectState.java
```diff
@@ -44,10 +44,10 @@ public XPUDeviceBufferState getDeviceBufferState(TornadoDevice device) {
         if (!(device instanceof TornadoXPUDevice)) {
             throw new TornadoRuntimeException("[ERROR] Device not compatible: " + device.getClass());
         }
-        if (!deviceStates.containsKey(device)) {
-            deviceStates.put((TornadoXPUDevice) device, new XPUDeviceBufferState());
-        }
-        return deviceStates.get(device);
+        // One map operation instead of containsKey + put + get: this is on the hot path of every
+        // ALLOC/COPY/DEALLOC bytecode, and for a pipeline that runs many small task-graphs per frame it
+        // showed up as the single largest host-side cost (43% of JVM samples).
```
mikepapadim:
```suggestion
```

mikepapadim:
Added a third commit — a single-entry cache in `DataObjectState` — and re-profiled properly. The
result is worth stating carefully, because it changes what this PR should be understood to do.

## What the new commit does

An object is almost always used on the same device twice in a row, so the last `(device, state)` pair is
cached and served with one identity compare. The pair lives in one immutable holder read through a
single `volatile`, so a reader always observes a consistent pair; a stale pair fails the identity check
and falls back to the map. `clear()` — the only invalidation point, in the same class — resets it.

## JFR, 1 ms sampling, before vs after (KFusion, 882 frames, RTX 4090)

| | samples | `getDeviceBufferState` | `executeDeAlloc` | `enqueueRead` | NVRTC `CUDAProgram.build` |
|---|---|---|---|---|---|
| before | 3 386 | **11.7%** | 11.5% | 19.4% | 22.3% |
| after | 3 359 | **0.0%** | 0.0% | 21.3% | 22.4% |

The lookup and the whole DEALLOC bookkeeping path disappear from the profile completely.

## But wall-clock time does not move

Three runs each, mean per-frame computation time:

| | run 1 | run 2 | run 3 | mean |
|---|---|---|---|---|
| before | 2.798 ms | 2.810 ms | 2.818 ms | 2.809 ms |
| after | 2.794 ms | 2.796 ms | 2.808 ms | 2.799 ms |

0.4%, i.e. noise — and notice where the freed samples went: `enqueueRead` rose 19.4% → 21.3%. **That
host bookkeeping was hidden behind the blocking device→host read**, so removing it just lets the thread
reach the same wait sooner. It is CPU work genuinely removed (~11% of JVM samples), which matters on a
CPU-constrained host or for power, but it is not a throughput win on this workload and I am not
claiming one.

I think that is still a reasonable thing to merge — strictly less work per bytecode, no behaviour
change — but the decision is yours, and the honest framing is "removes measurable CPU work that is
currently masked by a blocking transfer", not "makes things faster".

## Side effects checked

- Trajectory is **bit-identical**: `compareRuns.py` over 882 frames gives max deviation `0.000e+00`,
  ATE RMSE 0.019406 m unchanged.
- `tornado-test --quickPass` on an `opencl,cuda` build: **1026 run / 230 failed**, and a per-class diff
  against `develop` shows **no differences at all** (the 230 are the pre-existing CUDA-library and MMA
  suites hitting the OpenCL default backend of a combined build).

stratika (approved):

PR #999, opened 2026-07-30
Emit variable declarations in a deterministic order

mikepapadim:
Two runs of the same application on the same machine currently produce **textually different kernel
sources** that are semantically identical. Example, `trackPose` from KFusion, same build, two
consecutive runs:

```diff
-  bool b_194, b_193, b_192, b_191;
+  bool b_192, b_191, b_194, b_193;
```

## Cause

`emitVariableDefs` collects the per-kind declaration groups in a `HashMap<Kind, Set<Variable>>` of
`HashSet`s. Both `Variable` and the platform `Kind` enums hash by identity, so their hash codes differ
in every JVM process, and with them the iteration order of the map and of each set — which is the
order the declarations are emitted in. Present identically in the CUDA, OpenCL and Metal backends
(`CUDABackend.java:241`, `OCLBackend.java:238`, `MetalBackend.java:263`).

## Fix

`LinkedHashMap` + `LinkedHashSet`, so declarations follow insertion order, which is the LIR traversal
order and therefore deterministic. No other behaviour changes.

## Why it matters

- `-Dtornado.opencl.source.dump=True` output can be diffed between runs — right now a diff of two runs
  is noise, which makes codegen changes hard to review.
- Any cache keyed on the generated source is defeated. Concretely, in the companion PR that adds an
  on-disk CUDA module cache, KFusion produced **22 cached modules after the first run and 13 more
  brand-new keys after an identical second run**; with this fix the second run adds none.

## Verification

Dumping every kernel of a full KFusion run twice, `diff -r`:

```
=== source diff across two runs ===
ALL SOURCES BYTE-IDENTICAL
```

`tornado-test --quickPass` on an `opencl,cuda` build (RTX 4090, CUDA 11.5, JDK 21): **1026 run / 230
failed**, unchanged from `develop` (those 230 are the pre-existing CUDA-library and MMA suites hitting
the OpenCL default backend of a combined build). A CUDA-default subset of 78 tests across arrays,
shared buffers, reductions, KernelContext reductions, matrix multiplication and conditionals passes
0 failures.

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/backend/CUDABackend.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/backend/CUDABackend.java
@@ -34,6 +34,7 @@ import static uk.ac.manchester.tornado.runtime.common.TornadoOptions.VIRTUAL_DEV
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


mikepapadim:
related to #1000 no fixed order causes different cache entries for generated kernels - so we recomplie without any reason while we have the binaries already

stratika (approved):

PR #1000, opened 2026-07-30
Cache compiled CUDA modules on disk: 2.2x faster start-up for kernel-heavy applications

mikepapadim:
Every process hands each kernel to NVRTC again. `CUDACodeCache.installSource` always calls
`createProgramWithSource`, and the existing code cache only *dumps* binaries — the load-back path is
reachable only for `CL_DEVICE_TYPE_ACCELERATOR` (`CUDACodeCache.java:172`), so on a GPU it never
triggers. For applications with many kernels this dominates start-up.

**Stacked on #999** — without deterministic kernel source the cache key changes every run
and nothing ever hits.

## Measurements (RTX 4090, CUDA 11.5, JDK 21)

GPULlama3 3B-Q8, 61 tokens, wall clock:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| cold (populates 65 modules) | 11.59 s | | |
| **warm cache** | **5.21 s** | **5.16 s** | **5.17 s** |
| `-Dtornado.cuda.codecache.enable=false` | 11.47 s | 11.44 s | |

**2.2x faster start-up, 6.3 s saved per run**, and steady-state throughput is untouched (51–54 tok/s
either way — that part is kernel execution, not compilation). Cache size for the whole model: 1.3 MB.

KFusion (882 frames): 5.37 s → 4.77 s wall; only ~13 kernels, so the absolute win is smaller, and
per-frame FPS is unchanged (318 → 328, within noise) exactly as expected.

`nsys` on a warm run confirms where the time went: `cuModuleLoadDataEx` is 256 calls totalling
**11.3 ms (0.7%)** — loading cubins is essentially free compared with compiling them.

A note on JFR for anyone re-measuring this: `CUDAProgram.build` covers *both* NVRTC compilation and
`cuModuleLoadDataEx`, so its sample share stays high on a warm run and cannot be used to show the
effect; and JFR's own overhead (allocation sampling plus 1 ms native sampling) roughly doubles this
start-up-bound workload. Wall clock and the `nsys` CUDA API summary are the reliable instruments here.

## Implementation

- `CUDAProgram.getModuleImage()` returns the compiled image via the existing `getBinaries` native.
  (It deliberately does not reuse `dumpBinaries`' device-matching loop: the driver does not implement
  `CL_PROGRAM_DEVICES`, so `getDevices()` returns zeros and that loop never matches. A CUDA program
  holds exactly one image.)
- `installSource` looks the image up first and, on a hit, goes through the existing
  `createProgramWithBinary` path, skipping NVRTC entirely. On a miss it compiles and stores.
- Key is SHA-256 over the kernel source plus device name, device version, driver version, NVRTC
  version and the compiler flags, so a cubin can never be reused across GPUs, driver or toolkit
  upgrades, or different flags.
- Writes go to a temporary file and are then moved into place (`ATOMIC_MOVE`, falling back to a plain
  move), so concurrent processes never observe a partially written module.
- Every failure path — unreadable entry, stale entry, load error — falls back to compiling from
  source. A corrupt cache can only cost time, never correctness.
- Location `TORNADOVM_HOME/var/cuda-codecache/device-<platform>-<device>`, override with
  `-Dtornado.cuda.codecache.dir`. Disable with `-Dtornado.cuda.codecache.enable=false`.

Enabled by default, since that is where the benefit is; happy to flip it to opt-in if you would rather
it prove itself first.

## Side effects checked

- `tornado-test --quickPass` (`opencl,cuda` build): **1026 run / 230 failed**, identical to `develop`,
  no per-class differences.
- CUDA-default subset (78 tests: arrays, shared buffers, reductions, KernelContext reductions, matrix
  multiplication, conditionals) run **twice**: 0 failures both times, 83 modules written on the first
  pass and **zero added on the second**, i.e. every kernel came from the cache and still produced
  correct results.
- GPULlama3 generated text is byte-identical with and without the cache.

### DIFF
--- a/docs/source/flags.rst
+++ b/docs/source/flags.rst
@@ -136,6 +136,8 @@ TornadoVM's **CUDA C** backend targets NVIDIA GPUs: it emits CUDA C, compiled to
    ================================================================  ==================================================================================================================
    ``-Dtornado.cuda.compiler.flags=FLAGS``                           Passes additional flags to NVRTC when compiling the generated CUDA C source (default: none).
    ``-Dtornado.cuda.host.pinning=false``                             Disables pinning host memory for faster host↔device transfers (default: true).
+   ``-Dtornado.cuda.codecache.enable=false``                         Disables the on-disk cache of compiled CUDA modules, so every run re-compiles kernels with NVRTC (default: true).
+   ``-Dtornado.cuda.codecache.dir=PATH``                             Sets the directory used to store the on-disk CUDA module cache (default: ``/var/cuda-codecache``).
    ================================================================  ==================================================================================================================
 
 .. note::
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDACodeCache.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDACodeCache.java
@@ -32,9 +32,14 @@ import static uk.ac.manchester.tornado.runtime.common.Tornado.getProperty;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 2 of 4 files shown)


mikepapadim:
@stratika @orionpapadakis this one is performance-critical for GPULlama3: NVRTC recompilation of all 65 kernels is **67% of the JVM samples** on a 3B-Q8 run, so every process pays ~6.3 s before the first token.

With the cache warm, start-up drops from **11.5 s to 5.2 s (2.2x)** with byte-identical output and unchanged tok/s — the win is entirely on model load, which is exactly what interactive and serving use cases feel.

stratika (commented):
I will move the new flags in the `TornadoOptions` class.

stratika (approved):

PR #1002, opened 2026-07-30
Allocate wait-event lists on demand

mikepapadim:
A task-graph allocates its wait-event matrix eagerly:

```java
events = new int[this.bytecodeResult.getInt()][MAX_EVENTS];   // MAX_EVENTS defaults to 32768
```

That is **131 KB per dependency list**, allocated whether or not the graph ever uses it, filled with
`-1` in the constructor, and then filled again by `initWaitEventList()` on **every execution**. For a
plan made of many small graphs the memset dominates the per-execution cost.

Rows are now created on first append. A row that was never written stays `null`, which the drivers
already treat exactly like a list of `-1` entries — no events to wait on.

Three smaller items in the same area:

- `TornadoExecutionPlan.execute()` built the plan trace string unconditionally, on every execution;
  it is now supplied lazily and materialised only if `TornadoProfilerResult.getTraceExecutionPlan()`
  is actually called.
- `isPersistentObject` allocated a stream pipeline per object per ALLOC bytecode → plain loops.
- `isTaskNamePresent` concatenated `"<graph>.<task>"` for every task on every call → in-place compare
  with `regionMatches`.

## Throughput

GPULlama3 3B-Q8, 256 tokens, warm code cache, RTX 4090 / CUDA 11.5 / JDK 21:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| `develop` | 53.43 | 52.83 | 53.53 tok/s |
| **this branch** | **102.44** | **102.44** | **103.45 tok/s** |

**1.94x**, with byte-identical generated text.

KFusion (882 frames, ICL-NUIM): **316 → 424 FPS**, 2.82 → 2.09 ms per frame, and the trajectory is
**bit-identical** (`compareRuns.py` max deviation `0.000e+00`, ATE RMSE 0.019406 m unchanged).

### Attribution

I built a control with the interpreter change reverted but the other three kept: **53.50 / 51.50 /
53.28 tok/s**, i.e. unchanged. The entire throughput win is the lazy wait-event rows; the other three
items only remove garbage.

## Allocation

JFR `jdk.ObjectAllocationSample`, weight attributed to the deepest TornadoVM frame, same 256-token run:

| site | before | after |
|---|---|---|
| `GridScheduler.toString` | 244.4 MiB | 0 |
| `TornadoVMInterpreter.<init>` | 144.9 MiB | 0 |
| `TornadoTaskGraph.isTaskNamePresent` | 127.5 MiB | 0 |
| `WithGridScheduler.toString` | 23.2 MiB | 0 |
| `TornadoVMInterpreter.isPersistentObject` | 8.0 MiB | 0 |
| **total TornadoVM-attributed** | **777.0 MiB** | **3.0 MiB** |

## Side effects

- `tornado-test --quickPass` on an `opencl,cuda` build: 1026 run, 231 failed against a `develop`
  baseline of 230. The one difference is `branching.TestLoopConditions`, which is **pre-existing
  flaky**: I measured it 10 times on this branch (**3 failures**) and 10 times with the interpreter
  change reverted (**4 failures**). See the note below.
- GPULlama3 output byte-identical; KFusion trajectory bit-identical.

## Aside: `TestLoopConditions` is flaky and the symptom looks real

`testConditionBeforeSingleLoopReturn` fails roughly a third of runs on `develop` with garbage instead
of zero:

```
\_[REASON] expected:<0> but was:<1286959321>
\_[REASON] expected:<0> but was:<-1045729832>
\_[REASON] expected:<0> but was:<-184372750>
```

Values like that mean the assertion reads device memory that was never written, so this is worth
looking at on its own — either the test is missing a transfer or something is not synchronising. Happy
to open a separate issue with the reproduction if useful. (It is what made me double-check this PR
carefully: an early version of the change cleared only the used prefix of each wait list, which *did*
introduce the same symptom, because `resetEventIndexes()` rewinds a list mid-execution and entries can
live beyond the current index. That version is gone — whole rows are cleared, only allocated ones.)

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutionPlan.java
@@ -182,7 +182,7 @@ public sealed class TornadoExecutionPlan implements AutoCloseable permits Execut
      */
     public TornadoExecutionResult execute() {
         tornadoExecutor.execute(executionFrame);
-        TornadoProfilerResult profilerResult = new TornadoProfilerResult(tornadoExecutor, this.getTraceExecutionPlan());
+        TornadoProfilerResult profilerResult = new TornadoProfilerResult(tornadoExecutor, this::getTraceExecutionPlan);
         TornadoExecutionResult executionResult = new TornadoExecutionResult(profilerResult);
         planResults.add(executionResult);
         tornadoExecutor.updateLastExecutedTaskGraph();
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoProfilerResult.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoProfilerResult.java
@@ -17,6 +17,8 @@
  */
 package uk.ac.manchester.tornado.api;
 
+import java.util.function.Supplier;
+
 import uk.ac.manchester.tornado.api.enums.ProfilerMode;
 import uk.ac.manchester.tornado.api.profiler.ProfilerInterface;
 
@@ -41,11 +43,12 @@ import uk.ac.manchester.tornado.api.profiler.ProfilerInterface;
 public class TornadoProfilerResult implements ProfilerInterface {
 
     private final TornadoExecutor executor;
+    private Supplier<String> traceExecutionPlanSupplier;
     private String traceExecutionPlan;
 
-    TornadoProfilerResult(TornadoExecutor executor, String traceExecutionPlan) {
+    TornadoProfilerResult(TornadoExecutor executor, Supplier<String> traceExecutionPlanSupplier) {
         this.executor = executor;
-        this.traceExecutionPlan = traceExecutionPlan;
+        this.traceExecutionPlanSupplier = traceExecutionPlanSupplier;
     }
 
     /**
@@ -228,6 +231,11 @@ public class TornadoProfilerResult implements ProfilerInterface {
     }
 
     public String getTraceExecutionPlan() {
+        // Built on demand: rendering the plan description costs a lot of string garbage and almost no
+        // caller ever asks for it.
+        if (traceExecutionPlan == null && traceExecutionPlanSupplier != null) {
+            traceExecutionPlan = traceExecutionPlanSupplier.get();
+        }
         return traceExecutionPlan;
     }
 
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
@@ -32,7 +32,6 @@ import java.util.BitSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import uk.ac.manchester.tornado.api.GridScheduler;
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
@@ -693,7 +693,7 @@ public class TornadoVMInterpreter {
      * exactly like a list of -1 entries: no events to wait on.
      */
     private int[] waitListFor(int eventId) {
-        return (eventId >= 0 && eventId < events.length) ? waitListFor(eventId) : null;
+        return (eventId >= 0 && eventId < events.length) ? events[eventId] : null;
     }
 
     private int[] waitListForWrite(int eventId) {
--- a/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
+++ b/tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/interpreter/TornadoVMInterpreter.java
@@ -152,17 +152,17 @@ public class TornadoVMInterpreter {
         this.bytecodeResult.getLong(); // Skips bytes not needed
 
         kernelStackFrame = graphExecutionContext.getKernelStackFrame();
-        events = new int[this.bytecodeResult.getInt()][MAX_EVENTS];
+        // Rows are allocated on first use: MAX_EVENTS defaults to 32768, so an eager matrix costs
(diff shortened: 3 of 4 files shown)


mikepapadim:
@stratika @orionpapadakis this is the biggest GPULlama3 win so far: the interpreter was memsetting a `int[dependencies][32768]` wait-event matrix on **every** execution, i.e. once per token step, so decode was paying megabytes of `Arrays.fill` per token.

Making those rows lazy takes 3B-Q8 from **53 to 103 tok/s (1.94x)** with byte-identical output, and KFusion from 316 to 424 FPS with a bit-identical trajectory — no API change, no behaviour change.

mikepapadim:
## JFR CPU profile, before vs after the wait-event change

Two builds differing in **exactly one file** (`TornadoVMInterpreter.java`, `develop` version vs this
branch) so the wait-list change is isolated from the other three items in this PR. Same JFR settings on
both (`settings=profile`, `jdk.ExecutionSample#period=1ms`, `jdk.NativeMethodSample#period=1ms`),
GPULlama3 3B-Q8, RTX 4090 / CUDA 11.5 / JDK 21, 512 generated tokens so decode dominates the recording.

| | before | after |
|---|---|---|
| samples | 19 210 | 15 371 |
| recording duration | 23 s | 19 s |
| tok/s (under JFR) | 41.50 | 60.55 |

### Where the samples went

| frame | before | after |
|---|---|---|
| `java.util.Arrays.fill` (top of stack) | **3 520 (18.3%)** | **0 (0.0%)** |
| `TornadoVMInterpreter.initWaitEventList` (anywhere in stack) | **3 553 (18.5%)** | **0 (0.0%)** |
| `TornadoVMInterpreter.execute` (anywhere in stack) | 9 971 (51.9%) | 6 612 (43.0%) |
| `CUDACommandQueue.flush` (top of stack) | 6 385 (33.2%) | 6 594 (42.9%) |
| `CUDAProgram.build` — NVRTC (anywhere) | 6 269 (32.6%) | 6 157 (40.1%) |
| `CUDACommandQueue.enqueueRead` (anywhere) | 269 (1.4%) | 220 (1.4%) |
| `CUDACommandQueue.enqueueNDRangeKernel` (anywhere) | 159 (0.8%) | 124 (0.8%) |

At a 1 ms sampling period, 3 520 samples is roughly **3.5 s of CPU spent in `Arrays.fill`** across a
23 s run producing 512 tokens — about **7 ms of memset per token**, which is the whole cost of clearing
`int[dependencies][32768]` on every execution. After the change it is not in the profile at all.

Everything else keeps its absolute weight (NVRTC compilation, `flush`, `enqueueRead`,
`enqueueNDRangeKernel` are all within noise of their previous sample counts) and simply grows as a
*share* because the run is shorter — which is what a genuine removal of work looks like, as opposed to
a shift of cost somewhere else.

Two caveats, stated so the numbers are not over-read:

- JFR compresses the win: 41.5 → 60.6 tok/s (1.46x) under instrumentation versus 53 → 103 tok/s
  (1.94x) uninstrumented, because JFR's own overhead is common to both builds.
- `TornadoVMInterpreter.<init>` appears in only 2 CPU samples before the change even though it was
  144.9 MiB in the allocation profile. The eager matrix was mostly an *allocation* cost; the CPU cost
  was the per-execution `initWaitEventList` memset above. The two profiles complement each other.

Recordings: `jfr-before.jfr` / `jfr-after.jfr`, aggregated by attributing each sample to its deepest
TornadoVM frame.

mikepapadim:
CI was right and I was wrong — `build-cuda` failure fixed in the commit above.

## What broke

```java
private int[] waitListFor(int eventId) {
    return (eventId >= 0 && eventId < events.length) ? waitListFor(eventId) : null;  // calls itself
}
```

The helper returned itself instead of `events[eventId]`, so every lookup recursed until the stack
overflowed. `StackOverflowError` carries no message, which is why the suite reported `[REASON] null`
rather than anything useful.

## Why my local testing missed it

Only `withIntraPlanConcurrency()` asks for a wait list — single-stream execution has
`useDependencies == false` and passes `null` without ever calling the helper. So `TestCUDAStreams` was
the *only* suite that could fail, and my local `quickPass` runs on a combined `opencl,cuda` build, which
defaults to OpenCL and skips those CUDA-only tests. Both benchmarks I used (KFusion, GPULlama3) are
single-stream, so they measured the intended win while the multi-stream path was broken. My mistake:
I knew quickPass defaults to OpenCL — I wrote that down while testing #1000 — and still did not run the
CUDA-default sweep for this PR.

## Now verified on this branch

| check | result |
|---|---|
| `TestCUDAStreams` (CUDA default) | **13/13, 0 failures** (was 2) |
| `TestSharedBuffers`, `TestArrays` (CUDA default) | 12/12, 23/23 |
| `tornado-test --quickPass` | 1026 run / **230 failed** = the `develop` baseline exactly |
| `./mvnw checkstyle:check` | 0 violations |
| GPULlama3 3B-Q8, 256 tok | 102.5 / 99.8 tok/s (unchanged by the fix) |
| KFusion 882 frames | 439 FPS, 2.004 ms/frame, ATE 0.019406 m (unchanged) |

The performance claim in the description stands: the fix touches only the multi-stream lookup path,
which neither benchmark exercises.

mikepapadim:
Rebased onto current `develop` and carried the one-line #1007 fix, so this PR's CI can go green.

The `build-cuda` failure you would have seen last is **not from this PR** — `develop` itself does not
compile right now:

```
TestAtomicRmw.java:[153,45] error: cannot find symbol
  symbol:   variable SPIRV
  location: class TornadoVMBackendType
```

#991 landed that test guarding on `TornadoVMBackendType.SPIRV`/`.PTX`, which no longer exist, so every
open PR is red until #1007 merges. This branch now carries that same one-line fix; the commit becomes
redundant and can be dropped when #1007 lands.

## Re-validated after the rebase

`develop` moved (it now has #991, #996, #997, #999, #1004), so the old quickPass baseline of 1026/230 no
longer applies. I rebuilt a fresh baseline from `develop` + #1007:

| | tests run | failed |
|---|---|---|
| `develop` + #1007 (new baseline) | 1033 | 236 |
| **this branch** | 1033 | **236** |

Per-class diff over 127 classes: two entries move in opposite directions — `MMwithBytes` 1→0 and
`branching.TestLoopConditions` 0→1 — both of which I have previously measured as flaky on `develop`
itself (`TestLoopConditions` fails 3–4 times in 10 runs there). Net zero.

Also re-checked the two things that actually matter for this change:

| check | result |
|---|---|
| `TestCUDAStreams` (CUDA default) | **13/13, 0 failures** |
| `TestAtomicRmw` (CUDA default) | 5/5 |
| `./mvnw checkstyle:check` | 0 violations |

mikepapadim:
@stratika

stratika (approved):

PR #1010, opened 2026-07-30
Pass CUDA compiler flags to NVRTC, and add named compile profiles (default|fast|debug|repro)

mikepapadim:


stratika:
Please provide some examples how to test?

I do not understand the following?


| | before | after |
|---|---|---|
| tests failed | 0 of 19 | **19 of 19** |

stratika (approved):
LGTM

PR #1016, opened 2026-08-02
Add KernelContext coverage for globalBarrier, cp.async byte copies and flash attention

mikepapadim:
## What this covers

Three gaps found while auditing the `KernelContext` API against what the CUDA backend actually implements.

### `globalBarrier()` — no test at all

Nothing distinguished it from `localBarrier()`, and its documented semantics were never asserted. New `kernelcontext/api/TestGlobalBarrier` runs on every backend:

- `testGlobalBarrierOrdersGlobalWrites` — every thread writes one global element, then thread 0 of each block reads back the whole block's writes and sums them
- `testGlobalBarrierOrdersLocalWrites` — the same shape through a shared array

Both assert per-block sums, so a barrier that does not order the accesses fails. The scope asserted is the work-group, matching `barrier(CLK_GLOBAL_MEM_FENCE)` on OpenCL and `__syncthreads()` on CUDA — nothing here assumes ordering between blocks.

### `asyncCopyToLocal(int[], int, ByteArray, int)` — untested overload

`TestMatrixMultiplicationMMACpAsync` covers the `HalfFloatArray` overload inside a full MMA kernel; the `ByteArray` overload (the int8 staging path for quantized GEMM) and the `FP8Array` one had no coverage. New `kernelcontext/api/TestAsyncCopyToLocal` isolates the copy: stage 256 bytes into a shared tile with `cp.async`, commit, wait, and compare each 32-bit slot against the host-side little-endian packing. A failure points at the copy rather than at the matrix pipeline.

### `TestFlashAttentionKernelContext` — exists but never runs

The class is in the tree and passes, but it is not in the `tornado-test` registry, so no sweep has ever executed it. Registered.

## Verification

RTX 4090, CUDA 12.6, `make BACKEND=cuda`:

```
$ tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestGlobalBarrier
	testGlobalBarrierOrdersGlobalWrites  ....  [PASS]
	testGlobalBarrierOrdersLocalWrites   ....  [PASS]
Test ran: 2, Failed: 0, Unsupported: 0

$ tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestAsyncCopyToLocal
	testAsyncCopyByteArray               ....  [PASS]
Test ran: 1, Failed: 0, Unsupported: 0

$ tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.api.TestFlashAttentionKernelContext
	testFlashAttention                   ....  [PASS]
Test ran: 1, Failed: 0, Unsupported: 0
```

Tests only; no production code touched. `./mvnw checkstyle:check` clean.

`cp.async` is Ampere and later, so `TestAsyncCopyToLocal` is gated to CUDA with `assertNotBackend(OPENCL)` / `assertNotBackend(METAL)`.

If #1013 lands first, expect a one-line context conflict in `tornado-assembly/src/bin/tornado-test` — both PRs add entries to the same block.


tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/kernelcontext/api/TestAsyncCopyToLocal.java
```diff
@@ -0,0 +1,109 @@
...
+        int[] tile = ctx.allocateIntLocalArray(SLOTS);
+        int lane = ctx.localIdx;
+
+        ctx.asyncCopyToLocal(tile, lane, in, lane * 4);
+        ctx.asyncCopyCommit();
+        ctx.asyncCopyWaitGroup(0);
+        ctx.localBarrier();
+
+        out.set(ctx.globalIdx, tile[lane]);
+    }
```
mikepapadim, line 103: @mairooni @orionpapadakis

mikepapadim: Expanded in `2f69e22cd` — five more tests, taking this from one copy to the full `cp.async` contract.

The original test had a soundness hole: one slot per thread, `asyncCopyWaitGroup(0)`, and each lane read back **its own** slot. A copy that never reached shared memory, or a barrier that failed to publish the tile, still passes that. Two of the three source overloads and every non-zero wait-group argument had no coverage at all.

| Test | What it adds |
|---|---|
| `testAsyncCopyVisibleAcrossLanes` | Each lane reads the **mirrored** lane's slot, closing the self-read hole. The ramp input has no fixed point under the mirror, so the permutation is genuinely checked. |
| `testAsyncCopyMultipleSlotsPerThread` | Four copies per thread batched into one commit group; staged strided (coalesced) / read contiguous, so lanes cross. |
| `testDoubleBufferedPipeline` | The pattern `cp.async` exists for: issue chunk k+1 before consuming chunk k, with `asyncCopyWaitGroup(1)` retiring only the older group. First coverage of a non-zero wait-group argument and of multiple in-flight groups. |
| `testAsyncCopyFP8Array` / `testAsyncCopyHalfFloatArray` | The two uncovered overloads. FP8 uses raw storage bytes — the copy must move bytes without reinterpreting them. |

### Verification

`Test ran: 6, Failed: 0, Unsupported: 0` on RTX 3070 (sm_86) / CUDA 13.0, checkstyle clean. But a passing test proves little on its own here, so I checked two further things:

**The emitted PTX is real** (`--printKernel`). Every kernel lowers to `cp.async.ca.shared.global`, and the shared-tile sizes match intent — `[64]` single, `[256]` wide, `[128]` double-buffered. The fp16 overload addresses with stride `*2` against `*1` for the byte overloads, so the three overloads genuinely differ rather than collapsing to the same code. The pipeline kernel unrolls to exactly 4 `commit_group`, 3 `wait_group 1`, 1 `wait_group 0`, 7 `__syncthreads`.

**The pipeline test is sensitive, not just non-vacuous.** I mutated `asyncCopyWaitGroup(1)` to `(3)` (under-wait: chunk k no longer guaranteed landed) and rebuilt — it fails with `expected:<707142174> but was:<522001177>`. So it detects a wrong wait-group argument rather than merely that the kernel runs. Reverted and re-verified.

**All six execute on device** (`nsys profile --trace=cuda`, `cuda_gpu_kern_sum`): one instance each, 1.1–1.7 µs. Worth confirming explicitly because an unsupported path can bail to sequential host Java and still report a pass. No timings in the test itself — these are correctness-only.

One incidental finding: the FP8 `cp.async` path is **not** Ada-gated. `CUDATensorCoreSupportPhase` sets `hasFP8MMA` only for actual `mma` nodes, so the copy needs sm_80, not sm_89 — `testAsyncCopyFP8Array` runs on this sm_86 card rather than self-skipping.

Backend guard unchanged (OPENCL/METAL excluded), just factored into a shared `assumeCpAsyncBackend()` helper.

mikepapadim: Generated CUDA for `pipelinedStageKernel` (`tornado-test --printKernel`), since the double-buffering claim is only worth as much as what the compiler actually emits.

Reading the unrolled code, the buffer alternation is real — `i_3` = lane (buffer 0 slot), `i_5` = lane+64 (buffer 1 slot), `BYTES` = 256:

| Stage | Copy destination | Source offset | Wait | Read |
|---|---|---|---|---|
| prologue | `adi_2[i_3]` (buf 0) | `lane*4` — chunk 0 | — | — |
| k=0 | `adi_2[i_5]` (buf 1) | `+256` — chunk 1 | `wait_group 1` | `i_7 = adi_2[i_3]` (buf 0) |
| k=1 | `adi_2[i_3]` (buf 0) | `+512` — chunk 2 | `wait_group 1` | `i_9 = adi_2[i_5]` (buf 1) |
| k=2 | `adi_2[i_5]` (buf 1) | `+768` — chunk 3 | `wait_group 1` | `i_11 = adi_2[i_3]` (buf 0) |
| epilogue | — | — | `wait_group 0` | `i_12 = adi_2[i_5]` (buf 1) |

Buffers alternate 0,1,0,1 and every read pairs with the chunk that landed — so the overlap survives compilation rather than being collapsed. Each restage of a buffer is separated from the preceding read of that same buffer by a `__syncthreads()`, so there is no WAR race on the tile. Note also `cvta.to.shared.u64` per copy: the generic-to-shared address conversion `cp.async` requires.

<details>
<summary>Full generated kernel (66 lines)</summary>

```c
extern "C" __global__ void pipelinedStageKernel(long long *_kernel_context, unsigned char *_constant_region, unsigned char *_local_region, int *_atomics, unsigned char *in, unsigned char *out)
{
  unsigned long long ul_0, ul_1, ul_17; 
  int i_3, i_4, i_5, i_6, i_7, i_8, i_9, i_10, i_11, i_12, i_13, i_18, i_19, i_20; 
  long long l_14, l_15, l_16; 

  // BLOCK 0
  ul_0  =  (unsigned long long) in;
  ul_1  =  (unsigned long long) out;
  __shared__ int adi_2[128];
  i_3  =  (threadIdx.x);
  i_4  =  i_3 << 2;
  {
    unsigned __smem;
    asm volatile("{ .reg .u64 u; cvta.to.shared.u64 u, %1; cvt.u32.u64 %0, u; }" : "=r"(__smem) : "l"((char *) adi_2 + (((unsigned) i_3) << 2)));
    asm volatile("cp.async.ca.shared.global [%0], [%1], 4;" :: "r"(__smem), "l"((const char *) ul_0 + 16u + ((long long) i_4) * 1));
  }
  asm volatile("cp.async.commit_group;");
  i_5  =  i_3 + 64;
  i_6  =  i_4 + 256;
  {
    unsigned __smem;
    asm volatile("{ .reg .u64 u; cvta.to.shared.u64 u, %1; cvt.u32.u64 %0, u; }" : "=r"(__smem) : "l"((char *) adi_2 + (((unsigned) i_5) << 2)));
    asm volatile("cp.async.ca.shared.global [%0], [%1], 4;" :: "r"(__smem), "l"((const char *) ul_0 + 16u + ((long long) i_6) * 1));
  }
  asm volatile("cp.async.commit_group;");
  asm volatile("cp.async.wait_group 1;");
  __syncthreads();
  i_7  =  adi_2[i_3];
  __syncthreads();
  i_8  =  i_4 + 512;
  {
    unsigned __smem;
    asm volatile("{ .reg .u64 u; cvta.to.shared.u64 u, %1; cvt.u32.u64 %0, u; }" : "=r"(__smem) : "l"((char *) adi_2 + (((unsigned) i_3) << 2)));
    asm volatile("cp.async.ca.shared.global [%0], [%1], 4;" :: "r"(__smem), "l"((const char *) ul_0 + 16u + ((long long) i_8) * 1));
  }
  asm volatile("cp.async.commit_group;");
  asm volatile("cp.async.wait_group 1;");
  __syncthreads();
  i_9  =  adi_2[i_5];
  __syncthreads();
  i_10  =  i_4 + 768;
  {
    unsigned __smem;
    asm volatile("{ .reg .u64 u; cvta.to.shared.u64 u, %1; cvt.u32.u64 %0, u; }" : "=r"(__smem) : "l"((char *) adi_2 + (((unsigned) i_5) << 2)));
    asm volatile("cp.async.ca.shared.global [%0], [%1], 4;" :: "r"(__smem), "l"((const char *) ul_0 + 16u + ((long long) i_10) * 1));
  }
  asm volatile("cp.async.commit_group;");
  asm volatile("cp.async.wait_group 1;");
  __syncthreads();
  i_11  =  adi_2[i_3];
  __syncthreads();
  asm volatile("cp.async.wait_group 0;");
  __syncthreads();
  i_12  =  adi_2[i_5];
  i_13  =  (blockIdx.x*blockDim.x+threadIdx.x);
  l_14  =  (long long) i_13;
  l_15  =  l_14 + 4L;
  l_16  =  l_15 << 2;
  ul_17  =  ul_1 + l_16;
  i_18  =  i_7 + i_9;
  i_19  =  i_18 + i_11;
  i_20  =  i_19 + i_12;
  *(( int *) ul_17)  =  i_20;
  return;
}  //  kernel
```

</details>

For contrast, the single-copy `stageBytesKernel` emits one `cp.async.ca.shared.global` + `commit_group` + `wait_group 0` against `__shared__ int adi_2[64]`, and `stageHalfFloatKernel` differs only in the source stride (`* 2` rather than `* 1`) — which is what shows the three overloads lower differently instead of collapsing to the same code.

mikepapadim:
@stratika

stratika (approved):

PR #1017, opened 2026-08-02
Correct the hybrid-API documentation against the code on develop

mikepapadim:
## Why

The hybrid-API documentation describes a tree that does not exist. It promises a provider that was never merged, omits one that ships, and names factory methods that are not in the code. Every claim below was checked against `develop` @ b65c2053f.

## What was wrong

| Claim | Reality |
|---|---|
| `HYBRID_API_GUIDE.md`, `docs/source/hybrid-api.rst`: cuTENSOR (`nvidia/cutensor`) is a shipping provider, with a factory table, a troubleshooting row and an install row | No `tornado-cutensor` module on develop. It exists on branch `hybrid-cutensor`, unmerged |
| Both documents list the providers and omit cuSPARSE | `tornado-cusparse` ships (CSR SpMV/SpMM) |
| `HYBRID_API_GUIDE.md`: "18 overloads (`libraryTask` with 1-18 typed arguments)" | 20, since the task-arity work |
| `HYBRID_API_GUIDE.md`: `CuBlasLt::matmulBiasGelu` | Actual name is `ltMatmulGeluBiasFP16`, with a different parameter list |
| `HYBRID_API_GUIDE.md`: `CuFft::cufftC2C` with `CuFft.FORWARD` / `CuFft.INVERSE` | Direction constants do not exist; the factories are `cufftForwardC2C` / `cufftInverseC2C` |
| `HYBRID_API_GUIDE.md` links `HYBRID_API_BRANCHES.md` | Not on develop |
| `docs/source/hybrid-api.rst`: header-only device libraries "plug into" the CUDA-C backend's NVRTC compilation | No CUB integration ships; that is a design note, not a feature |
| `docs/source/programming.rst`: CUDA column says `barrier.sync`, `.shared .s32 array[size]` | PTX-era text; the CUDA-C backend emits `__syncthreads()` and `__shared__ int` |

## What this changes

- **`HYBRID_API_GUIDE.md`**: cuTENSOR section replaced by cuSPARSE, with cuTENSOR called out as living on a branch; arity corrected; cuBLASLt and cuFFT factory names, tables and examples corrected against the source; module lists, install table and the `--add-modules` line corrected; dead link removed.
- **`docs/source/hybrid-api.rst`**: same provider-list correction, plus a note that no CUB integration ships.
- **`docs/source/programming.rst`**: CUDA column fixed, and a new table of which `KernelContext` operations each backend supports, so the limits are discoverable before you hit them.
- **`KernelContext` javadoc**: both barriers say `__syncthreads()` for CUDA rather than `barrier.sync`, and `globalBarrier` states that its scope is the work-group - `__syncthreads()` already makes a block's prior global and shared accesses visible to the block, which is why both fence flags map to it.

Documentation and javadoc only; no behaviour change. `./mvnw checkstyle:check` clean, `make BACKEND=cuda` clean.

The support table in `programming.rst` says an unsupported operation is rejected rather than silently wrong, which is the behaviour after #1013. If that PR is rejected or changed, the wording here should follow it.

### DIFF
--- a/HYBRID_API_GUIDE.md
+++ b/HYBRID_API_GUIDE.md
@@ -2,7 +2,7 @@
 
 The **hybrid API** lets a single TornadoVM `TaskGraph` mix JIT-compiled Java
 tasks (`@Parallel` / `KernelContext`) with calls into vendor-optimized native
-GPU libraries — **cuBLAS, cuBLASLt, cuFFT, cuDNN, CUTLASS, and cuTENSOR**.
+GPU libraries — **cuBLAS, cuBLASLt, cuFFT, cuDNN, cuSPARSE, and CUTLASS**.
 A native call becomes a **library task**: it shares TornadoVM-managed device
 buffers with the surrounding kernels, runs on the same CUDA stream, and is
 captured into CUDA Graphs — so data produced by a JIT kernel feeds a library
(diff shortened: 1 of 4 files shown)


stratika (approved):
LGTM

PR #1018, opened 2026-08-02
Expand unit-test coverage for Java language corner cases and compiler restrictions

mikepapadim:
## Summary

Adds 41 new unit tests to `tornado-unittests`, split into two groups:

- **Gap-fills** (37 tests) for Java-language corner cases that had zero or thin coverage in the existing suite (compound assignment, do-while, labeled break/continue, switch fallthrough, char arithmetic, NaN/Infinity/-0.0, integer overflow wraparound, jagged/ragged arrays).
- **Compiler-boundary regression tests** (4 tests, new `fails/` classes) that pin down previously-undocumented-by-test restrictions in the Graal-based kernel compiler (multi-dimensional array declaration, dynamically-sized array allocation, non-constant `@Parallel` loop stride, OpenCL float atomics).

All tests were run individually against the CUDA backend (device `0:0`) and verified PASS (or the expected exception, for the `fails/` tests). A full `make fast-tests` run afterward shows `971 PASS` (up from a `930 PASS` baseline), the same 7 pre-existing whitelisted failures, and the same 107 unsupported — zero new non-whitelisted regressions.

## Tests added

| File | Test(s) | Purpose |
|---|---|---|
| `numpromotion/TestNumericPromotion.java` | `testCompoundAddAssign`, `testCompoundSubAssign`, `testCompoundMulAssign`, `testCompoundDivAssign`, `testCompoundModAssign`, `testCompoundAndAssign`, `testCompoundOrAssign`, `testCompoundXorAssign`, `testCompoundLeftShiftAssign`, `testCompoundRightShiftAssign`, `testCompoundUnsignedRightShiftAssign` | All 11 compound-assignment operators (`+= -= *= /= %= &= \|= ^= <<= >>= >>>=`) on `byte` — compound assignment includes an implicit narrowing cast that plain binary operators don't, previously untested |
| `numpromotion/Types.java` | `testCharToInt`, `testCharArithmetic` | `char` is unsigned 16-bit — widening to `int` must NOT sign-extend, unlike `byte`/`short`; plus `char` arithmetic with narrowing cast back |
| `numpromotion/TestZeroExtend.java` | `testIntOverflowWraparound`, `testLongOverflowWraparound` | Explicit `Integer.MAX_VALUE + 1` / `Long.MAX_VALUE + 1` two's-complement wraparound assertions (previously only implicit) |
| `branching/TestConditionals.java` | `testSwitchFallthrough`, `testTripleNestedTernary` | A `switch` case with no `break` falling into the next case; ternary nested 3 levels deep (existing tests stopped at 2) |
| `branching/TestLoopConditions.java` | `testDoWhileLoop`, `testDoWhileLoopWithCondition`, `testLabeledBreakOuterLoop`, `testLabeledContinueOuterLoop` | `do-while` loops (file previously only covered `for`); labeled `break`/`continue` targeting an outer loop |
| `math/TestMath.java` | `testFloatPositiveInfinity`, `testFloatNegativeInfinity`, `testFloatIsInfinite`, `testFloatIsNaN`, `testDoublePositiveInfinity`, `testDoubleNegativeInfinity`, `testNegativeZeroFloat`, `testNegativeZeroDouble` | `POSITIVE_INFINITY`/`NEGATIVE_INFINITY`/`isInfinite`/`isNaN` for `float`/`double` (only `signum` was tested before); sign-of-zero via `1.0f / -0.0f == -Infinity` |
| `foundation/TestShorts.java` | `testShortSub`, `testShortMul`, `testShortDiv` | File only had `testShortAdd`; every sibling type (`TestIntegers`, `TestDoubles`, `TestFloats`) already had all 4 ops |
| `logic/TestLogic.java` | `testLogicShortCircuitSideEffect`, `testLogicXorPattern`, `testLogicNestedBoolean` | `&&` short-circuit avoiding a divide-by-zero; boolean `^` (exactly-one-of-two); nested `!`/`&&`/`\|\|` |
| `arrays/TestJaggedArrays.java` *(new)* | `testRaggedRowSums`, `testRaggedRowSumsRandom` | Ragged/jagged host data (`int[][]` with non-uniform row lengths) flattened to a CSR-style buffer + row-offset table; exercises a per-thread loop trip count that varies with the parallel index |
| `fails/MultiDimArrayFail.java` *(new)* | `testMultiDimArrayDeclaration` | `int[][]` declared inside a kernel is rejected (`Error` / Graal `NewMultiArray` not handled) — distinct from the supported flattened `Matrix2DInt`-style host types |
| `fails/DynamicArrayFail.java` *(new)* | `testDynamicSizedArrayAllocation` | A kernel-local array sized by a runtime (non-constant) value is rejected (`TornadoInternalError`) |
| `fails/NonConstantStrideFail.java` *(new)* | `testNonConstantLoopStride` | A `@Parallel` loop with a non-constant stride fails to auto-parallelize (`TornadoBailoutRuntimeException`) |
| `fails/FloatAtomicFail.java` *(new)* | `testFloatAtomicAddOpenCL` | `KernelContext.atomicAdd(FloatArray, ...)` works on CUDA (native `atomicAdd(float*, float)`), explicitly excluded on OpenCL (`atom_add` has no float overload) |

`tornado-assembly/src/bin/tornado-test` is updated to register the one new test class (`TestJaggedArrays`) plus the four new `fails/` classes in `__TEST_THE_WORLD__` so they run under the full suite / `make tests`.

## Test plan

- [x] `JAVA_HOME=<jdk21> make BACKEND=cuda` builds clean
- [x] Each new/modified test class run individually via `tornado-test -V <class>` on CUDA device `0:0` — all pass (or throw the expected exception)
- [x] `./mvnw checkstyle:check` clean on `tornado-unittests` and `tornado-assembly`
- [x] Full `make fast-tests` (CUDA backend): 971 PASS / 7 whitelisted FAILED (unchanged) / 107 UNSUPPORTED (unchanged), exit 0


tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/vectortypes/TestHalfFloats.java
```diff
@@ -927,6 +927,57 @@ public void testInternalSetMethod04() throws TornadoExecutionPlanException {
...
+            HalfFloat subtracted = HalfFloat.sub(added, d.get(i));
+            HalfFloat divided = HalfFloat.div(subtracted, b.get(i));
+            results.set(i, divided);
+        }
+    }
+
+    @Test
+    public void testHalfFloatArithmeticChain() throws TornadoExecutionPlanException {
+        final int size = 64;
+        HalfFloatArray a = new HalfFloatArray(size);
```
mikepapadim, line 945: important for llama fuded kernels

mikepapadim:
@stratika fixed — ready to merge.

`TestMath#testFloatIsNaN` is now guarded on the Metal backend (`af4912f27`).

**Root cause:** `Float.isNaN` isn't registered as an `InvocationPlugin` on *any* backend, so it lowers to its Java bytecode form, `v != v`. MSL compiles with fast-math enabled by default, which permits assuming no NaNs are present — so the self-comparison folds to a constant `false` and `isNaN` reports 0 for NaN input.

That's not a codegen bug, it's the documented consequence of fast-math. Fixing it properly would mean either registering an `isnan` intrinsic for the Metal backend or disabling fast-math there — both behaviour changes well outside a test-coverage PR — so the test is guarded with an explicit reason instead, matching the OpenCL NaN exclusions already in this file.

**Only this one test is affected**, which the failure list confirms:
- `Float.isInfinite` is a magnitude comparison, so it survives fast-math → `testFloatIsInfinite` passes
- `testMathSignumFloatNaN` passes because the Metal preamble routes signum through a `signum_f` shim that calls the real MSL `isnan` intrinsic

**Verified:** all 36 `TestMath` tests pass on CUDA, `./mvnw checkstyle:check` clean. I could not exercise the Metal path locally — no Metal device on this machine — so that guard is unverified on hardware, though it uses the same mechanism as the existing OpenCL guards in the same file. Worth a Metal CI re-run to confirm.

mikepapadim:
Both remaining CI failures fixed — pushed as `074f60fb0` (on top of the earlier `af4912f27` NaN guard).

I reproduced both locally on an NVIDIA OpenCL device rather than guessing from CI. Different causes:

**1. `TestGlobalAtomics#testGlobalAtomicAddFloatAcrossGrid`**

```
unimplemented: In OpenCL, the atom_add function does not support floating point operations.
```

Rejected at sketch time by `OCLGraphBuilderPlugins#registerUnsupportedAtomicAddPlugin`. This one was my oversight — `fails/FloatAtomicFail` *in this same PR* documents exactly that restriction, and I then wrote the sibling test without the guard. Now guarded on OpenCL only; the int-based test in the same class still runs everywhere, CUDA has a native `atomicAdd(float*, float)`, and Metal emulates one with a CAS-loop helper.

**2. `TestFP8ElementwiseCopy#testFP8ElementwiseCopyOutsideMMA`**

```
module java.base does not "opens jdk.internal.foreign" to module tornado.drivers.opencl
```

FP8 is a CUDA-only storage type — `FP8Array` is registered only in the CUDA device dispatch, so elsewhere the runtime falls back to a reflective path over the memory segment. Uses the same `assumeCudaBackend()` helper that the pre-existing `TestFP8` and `TestBFloat16` already use for this exact reason, rather than inventing a new guard.

### Verification

Rather than round-trip with CI I ran the whole suite on both backends locally.

| | OpenCL | CUDA |
|---|---|---|
| `make fast-tests` | exit 0 — **834 PASS, 9 FAILED (all whitelisted), 0 non-whitelisted** | exit 0 — 999 PASS, whitelisted-only |
| `TestFP8ElementwiseCopy` | 1 run, 1 unsupported | 1 run, **0 unsupported** |
| `TestGlobalAtomics` | 2 run, 1 unsupported | 2 run, **0 unsupported** |
| `TestMath` | 36 run, 10 unsupported | 36 run, **0 unsupported** |

The CUDA column is the important one: every guard is a no-op there, so none of this reduces CUDA coverage — the tests still exercise everything they were written to exercise.

Metal remains unverifiable on my machine, so those guards rely on CI. Ready for a re-run.

mikepapadim:
Correction to the CUDA figure in my comment above: I quoted **999 PASS**, but that run predated the two guard commits — I presented it as if it covered current HEAD. It didn't.

Re-ran the full CUDA suite at `074f60fb0`: exit 0, **998 PASS / 8 FAILED (all whitelisted) / 107 UNSUPPORTED, zero non-whitelisted**.

The one-test delta is `reductions.TestReductionsFloats#testComputePi`, a pre-existing whitelisted float-reduction flake — unrelated to this PR (which touches only `TestMath`, `TestGlobalAtomics`, `TestFP8ElementwiseCopy`). It's whitelisted precisely because it's non-deterministic.

Conclusion is unchanged — zero non-whitelisted failures on both CUDA and OpenCL — but the number should be right.

orionpapadakis (approved):

PR #1019, opened 2026-08-02
Fix atomicAdd on local-memory arrays with a runtime-computed index (CUDA backend)

mikepapadim:
## Problem

`KernelContext.atomicAdd(int[] localArray, dynamicIndex, val)` — an atomic add into a **local/shared-memory** array at a **runtime-computed index** — silently dropped the index and always updated element `0`. No exception, no compile error, just a wrong answer.

Confirmed with `--printKernel`; the generated CUDA computes the byte offset for the index and then never uses it:

```c
i_12  =  i_11 + 4;
l_13  =  (long long) i_12;
l_14  =  l_13 << 2;          // byte offset computed...
i_15  =  atomicAdd((int *) adi_2, 1);   // ...and never applied
```

This makes the canonical two-level (per-block shared-memory, then merged to global) histogram — and any other dynamically-bucketed shared-memory atomic — silently produce wrong results. Since a per-block shared-memory histogram is the standard way to avoid hammering global-memory atomics, this also has a real performance cost: the pattern has to be rewritten onto global atomics to be correct.

## Root cause

`CUDAAddressNode#setMemoryAccess` builds addresses differently per memory space:

- **global**: the index is folded into the base pointer up front via `emitAdd`, so consumers receive a ready-to-use address.
- **local/private**: base and index are kept **separate** on the `MemoryAccess` object.

`LoadStmt` (ordinary local-array reads) handles that separated form via its own dedicated code path (`emitIntegerBasedIndexCode`), which is why plain `localArray[i] = x` always worked. But `CUDAUnary.AtomOperation` only ever called `address.emit(crb, asm)` — which emits the bare base value — and never consulted `address.getIndex()`. Every pre-existing local-array atomic in the codebase passes a literal `0` index, so this never surfaced.

## Fix

Emit the index in `AtomOperation#emit` when present. Two subtleties, both established empirically rather than assumed:

1. **Byte-pointer arithmetic, not array subscript.** The index from `CUDAGraphBuilderPlugins#computeAddress` is already a *byte* offset, so it is applied as `(char *) base + off`. Using `base[off]` would make C auto-multiply by `sizeof(element)` and scale an already-scaled offset a second time. (An intermediate attempt using `&(base[off])` produced `CUDA_ERROR_ILLEGAL_ADDRESS`, and a bare `(cast) base[off]` casts the element *value* rather than its address, since `[]` binds tighter than a cast.)
2. **The Panama array-header adjustment must be backed out for local memory only.** That same byte offset includes an off-heap array-header adjustment. It is correct for private-memory `int[]` but wrong for local/shared `int[]`, which is a plain `__shared__` C array with no header. The two spaces are indistinguishable at plugin-registration time (both are just `int[]`), so the correction is applied here, where the base's memory region is known. Removing the header at the plugin level instead was tried and regressed `TestAtomics#testAtomic18_parallel_api_int_array`, which genuinely needs it.

Resulting codegen:

```c
i_14  =  atomicAdd((int *) ((char *) adi_2 + l_13), 1);
```

## Test

Adds `TestLocalArrayDynamicIndexAtomics` — the canonical two-level per-block shared-memory histogram, i.e. exactly the shape that was broken (bucket index derived from input data at runtime).

- [x] New regression test passes (fails on `develop` with `bin 0 expected:<136> but was:<2048>`)
- [x] All 27 pre-existing `TestAtomics` tests still pass (1 unsupported, unchanged)
- [x] `make fast-tests` (CUDA backend, RTX 3070 / sm_86 / CUDA 13.0): **931 PASS, 7 FAILED (all whitelisted), 107 UNSUPPORTED**, exit 0 — zero non-whitelisted failures

### DIFF
--- a/tornado-assembly/src/bin/tornado-test
+++ b/tornado-assembly/src/bin/tornado-test
@@ -189,6 +189,7 @@ __TEST_THE_WORLD__ = [
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestVectorAdditionKernelContext"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.KernelContextWorkGroupTests"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestAtomicRmw"),
+    TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestLocalArrayDynamicIndexAtomics"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.api.TestHalfFloatLocalArray"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.local.memory.TestSwizzledLocalArrays"),
     TestEntry("uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationKernelContext"),
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDAUnary.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDAUnary.java
@@ -38,6 +38,7 @@ import uk.ac.manchester.tornado.drivers.cuda.graal.asm.CUDAAssemblerConstants;
 import uk.ac.manchester.tornado.drivers.cuda.graal.compiler.CUDACompilationResultBuilder;
 import uk.ac.manchester.tornado.drivers.cuda.graal.meta.CUDAMemorySpace;
 import uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDABarrierNode.CUDAMemFenceFlags;
+import uk.ac.manchester.tornado.runtime.TornadoCoreRuntime;
 
 public class CUDAUnary {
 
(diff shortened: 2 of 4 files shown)


mikepapadim:
CI failures fixed — pushed as `20fff3f25`.

The OpenCL/Metal failures were correct, and my original scoping was wrong: **the defect is not CUDA-specific.** `OCLUnary.AtomOperation` has the identical shape — it calls `address.emit()`, which emits only the base and never consults `getIndex()` — and `OCLAddressNode` splits base/index for local and private memory exactly as the CUDA one does, with the same byte-scaled, header-inclusive offset from `computeAddress`.

### OpenCL: fixed and verified on hardware

Same fix, with address-space qualifiers preserved on both casts:

```c
atom_add((volatile __local int *) ((volatile __local char *) adi_2 + (l_14 - 16)), 1);
```

Byte-pointer arithmetic rather than a subscript (the offset is already byte-scaled), and the Panama header subtracted back out for `__local`, which has no such header. The global path is untouched:

```c
atom_add((volatile __global int *) ul_21, i_17);
```

Verified on an NVIDIA OpenCL device:
- the regression test passes
- `TestAtomics` (27) and `TestAtomicRmw` (5) clean
- full `make fast-tests` on the OpenCL backend: **exit 0, 787 PASS, zero non-whitelisted failures**

### Metal: guarded, not guessed at

Metal has the same dropped index, but fixing it is a larger change. `MetalUnary.AtomOperation` hardcodes the `device` address space in its atomic casts (`atomic_fetch_add_explicit((device atomic_int *) ...)`), which is already wrong for a threadgroup array irrespective of the index — MSL enforces strict address-space segregation, so local-memory atomics need the qualifier to vary too. That's separate from the index arithmetic, and I have no Metal hardware to validate it on, so the test is guarded with an explicit reason rather than shipping an unverified codegen change. Worth a follow-up issue if you'd like Metal local atomics working.

stratika (approved):
LGTM

PR #1020, opened 2026-08-02
Add CUDA-samples-derived compute examples with measured results (GEMM, N-body, binomial options, FDTD stencil)

mikepapadim:
## Summary

A runnable benchmark example that exercises the low-precision / tensor-core surface of `KernelContext` end to end, and measures it honestly. Six implementations of the same `C[1024,1024] = A[1024,1024] * B[1024,1024]`, run back-to-back on identical data so the tensor-core and launch-overhead effects are directly comparable:

| # | Stage | KernelContext API exercised |
|---|---|---|
| 1 | fp32 shared-memory tiled (baseline) | `allocateFloatLocalArray`, `localBarrier`, 2D worker grid |
| 2 | fp16 MMA | `mmaFragment`, `mmaLoadA`, `mmaLoadB`, `mma`, `mmaStore` (`M16N8K16`, fp32 accumulate) |
| 3 | bf16 MMA | same fragment plumbing, `mmaBF16` reinterprets the register tuples as bf16 |
| 4 | fp16 MMA + `cp.async` | `asyncCopyToLocal`, `asyncCopyCommit`, `asyncCopyWaitGroup` |
| 5 | fp16 MMA + CUDA Graphs | `withCUDAGraph()` — captures the launch sequence and replays it |
| 6 | FP8 E4M3 MMA | `mmaLoadAFP8`, `mmaLoadBFP8`, `mmaFP8E4M3` (`M16N8K32`) |

Correctness is checked for every stage against an fp32 CPU reference. fp16 lands at `maxErr ≈ 0.006`, bf16 at `≈ 0.07` — the expected trade, since bf16 keeps fp32's exponent range but has only 8 mantissa bits to fp16's 11.

## Measured results

RTX 3070 (sm_86, Ampere), CUDA 13.0, M=N=K=1024, mean of 20 steady-state iterations (warm-up excluded), kernel-only time from `--enableProfiler console` (`TASK_KERNEL_TIME`):

```
kernel                 device time    vs fp32
gemmFp32Tiled            1336 us       1.00x
gemmFp16MMA               773 us       1.73x
gemmBf16MMA               767 us       1.74x
gemmFp16MMACpAsync        327 us       4.08x
```

`nsys stats --report cuda_gpu_kern_sum` independently agrees on the three kernels it recorded (1197 / 715 / 292 µs).

FP8 reports as **skipped** on this hardware: FP8 tensor-core MMA needs compute capability ≥ 8.9 (Ada/Hopper), and sm_86 is Ampere.

## Two measurement findings, documented rather than papered over

**1. End-to-end wall-clock badly understates the kernel speedup.** The naive numbers show only ~1.2–1.6x, versus the real 1.7–4.1x. Each execution copies the 4 MB fp32 result back to the host, which is identical work for every stage and compresses all the ratios toward 1. The example prints both figures and says explicitly which one to trust, with the command to reproduce the kernel-only measurement.

**2. `nsys` silently omitted a kernel that definitely ran.** `gemmBf16MMA` appears nowhere in the nsys kernel summary across repeated runs, despite the kernel demonstrably executing — its cubin is emitted to the code cache, its results carry bf16-specific rounding error (0.0699, distinct from fp16's 0.0060), and TornadoVM's own profiler times it at 767 µs. Worth knowing as a profiler caveat: if a kernel goes missing from an nsys report, cross-check with `--enableProfiler console` before concluding it never ran. This is called out in the example's javadoc.

## A trap the FP8 stage has to defend against

Without `-Dtornado.recover.bailout=False` (which `tornado-test` sets but a plain `tornado` example run does not), an unsupported FP8 kernel does **not** throw. TornadoVM prints a `[Bailout]` warning and silently re-runs the method as sequential host Java — where `KernelContext` indices are all zero, so nothing meaningful is computed and the "kernel" appears absurdly fast. Timed naively this reports a fictional **33 TFLOP/s** on hardware that cannot do FP8 at all.

`FP8GemmStage` therefore validates the output before printing any number, and reports the stage as skipped if the device kernel never ran. Both failure modes (exception vs. silent bailout) are handled, since which one occurs depends on a JVM flag the example cannot assume.

## How to run

```bash
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.kernelcontext.matrices.LowPrecisionGemmBenchmark

# kernel-only timing
tornado --enableProfiler console -m tornado.examples/...LowPrecisionGemmBenchmark

# full timeline
nsys profile --trace=cuda,nvtx -o lowprec-gemm tornado -m tornado.examples/...LowPrecisionGemmBenchmark
nsys stats --report cuda_gpu_kern_sum lowprec-gemm.nsys-rep
```

## Test plan

- [x] `make BACKEND=cuda` builds clean
- [x] All six stages run; five execute on device and validate against the fp32 reference, FP8 correctly reports skipped on sm_86
- [x] Kernel times cross-validated with two independent tools (TornadoVM profiler + nsys)
- [x] Examples-only change — no runtime, driver, or API code touched, so no test-suite impact

### DIFF
--- a/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/kernelcontext/compute/BinomialOptions.java
+++ b/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/kernelcontext/compute/BinomialOptions.java
@@ -1,0 +1,372 @@
+/*
+ * Copyright (c) 2026 APT Group, Department of Computer Science,
+ * The University of Manchester.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.manchester.tornado.examples.kernelcontext.compute;
+
+import java.util.Random;
+
+import uk.ac.manchester.tornado.api.GridScheduler;
+import uk.ac.manchester.tornado.api.KernelContext;
+import uk.ac.manchester.tornado.api.TaskGraph;
+import uk.ac.manchester.tornado.api.TornadoExecutionPlan;
+import uk.ac.manchester.tornado.api.WorkerGrid1D;
+import uk.ac.manchester.tornado.api.annotations.Parallel;
+import uk.ac.manchester.tornado.api.enums.DataTransferMode;
+import uk.ac.manchester.tornado.api.exceptions.TornadoExecutionPlanException;
+import uk.ac.manchester.tornado.api.types.arrays.FloatArray;
+
+/**
+ * European call pricing with a Cox-Ross-Rubinstein binomial tree -- a
+ * <b>block-per-work-item wavefront</b>, which is a parallel decomposition nothing else in the
+ * TornadoVM examples uses.
+ *
+ * <p>
+ * Every other {@code KernelContext} example here maps one thread to one output element and uses
+ * the block purely as a staging area (tiled GEMM, tiled N-body, reductions). This one inverts
+ * that: <b>one block prices one option</b>, and the threads inside the block cooperate on that
+ * single option's tree. Each thread owns one tree node; the whole block then walks the tree
+ * backwards from expiry to the present, collapsing one level per iteration:
+ *
+ * <pre>
+ * leaf:  every thread computes its own terminal payoff
+ * for level = STEPS down to 1:
+ *     tree[tid] = myValue          // publish
+ *     localBarrier()               // everyone's value is visible
+ *     if (tid &lt; level)
+ *         myValue = pu*tree[tid+1] + pd*tree[tid]   // collapse one level
+ *     localBarrier()               // nobody overwrites before all reads land
+ * thread 0 holds the option price
+ * </pre>
+ *
+ * <p>
+ * Note the loop runs {@code STEPS} iterations with <b>two barriers each</b> -- 510 barriers per
+ * block, versus the single barrier per K-tile in the GEMM kernels. The active thread count shrinks
+ * by one every level, so the block finishes almost entirely idle. That is inherent to the
+ * algorithm, and it is exactly why this shape is interesting to have as an example: it stresses
+ * barrier throughput and divergence rather than memory bandwidth or FLOPs.
+ *
+ * <p>
+ * <b>Why both barriers are load-bearing.</b> The update reads {@code tree[tid]} and
+ * {@code tree[tid+1]} and writes {@code tree[tid]}. Drop the second barrier and thread
+ * {@code tid} can overwrite {@code tree[tid]} while thread {@code tid-1} is still reading it as
+ * its {@code tid+1} neighbour -- a classic in-place-sweep race. Keeping the new value in a
+ * register ({@code myValue}) and only publishing it at the top of the next iteration is what makes
+ * the two-barrier form correct.
+ *
+ * <p>
+ * A {@code @Parallel} one-option-per-thread version is included as a baseline. It needs no shared
+ * memory and no barriers at all, but each thread must hold the whole tree -- and on a GPU that is
+ * fatal: see its javadoc, the launch is rejected outright.
+ *
+ * <h2>Measured result</h2>
+ *
+ * <p>
+ * RTX 3070 (sm_86), CUDA 13.0, 2048 options, 255 tree steps, mean of 10 steady-state iterations:
+ *
+ * <pre>
+ *   block-per-option (GPU, kernel-only)   35265 us
+ *   sequential Java (1 CPU core)          44561 us
+ *   -> 1.25x
+ * </pre>
+ *
+ * <p>
+ * 1.25x over a single CPU core is a poor return for 524288 GPU threads, and the reason is
+ * structural rather than fixable by tuning. The kernel is <b>barrier- and divergence-bound</b>,
+ * not compute- or bandwidth-bound:
+ *
+ * <ul>
+ * <li>2048 blocks x 510 barriers is over a million block-wide synchronisations, and there is no
+ * arithmetic to hide them behind -- each level does two multiplies and an add per node.</li>
+ * <li>The active thread count falls by one per level, from 255 down to 1. Averaged over the
+ * sweep only half the block is ever doing useful work, and the final levels run a 256-thread
+ * block to compute a handful of values.</li>
+ * <li>Kernel-only time (35.3 ms) is essentially the whole end-to-end time (35.6 ms), so data
+ * movement is not the issue either.</li>
+ * </ul>
+ *
+ * <p>
+ * The honest summary: a binomial tree of this depth is a bad fit for one-block-per-option on this
+ * hardware. The real CUDA sample amortises the cost differently -- far more options in flight and
+ * a cache-blocked sweep so the tree does not have to be walked level-by-level in lockstep. This
+ * example is kept for the pattern, not the throughput: it is the only place in the TornadoVM
+ * examples where a block cooperates on a single work item, and it shows what that costs.
+ *
+ * <p>
+ * How to run:
+ * </p>
+ *
+ * <pre>
+ * tornado -m tornado.examples/uk.ac.manchester.tornado.examples.kernelcontext.compute.BinomialOptions
+ * </pre>
+ *
+ * <p>
+ * Kernel-only timing:
+ * </p>
+ *
+ * <pre>
+ * tornado --enableProfiler console -m tornado.examples/uk.ac.manchester.tornado.examples.kernelcontext.compute.BinomialOptions
+ * </pre>
+ */
+public class BinomialOptions {
+
+    /** Threads per block == tree leaves. STEPS is one less, so leaf index tid maps 1:1. */
+    private static final int BLOCK = 256;
+    private static final int STEPS = BLOCK - 1;
+
+    private static final int NUM_OPTIONS = 2048;
+    private static final int ITERATIONS = 10;
+
+    private static final float RISK_FREE = 0.02f;
+    private static final float VOLATILITY = 0.30f;
+
+    /**
+     * One block per option, one thread per tree node.
+     */
+    public static void binomialKernelContext(KernelContext ctx, FloatArray spot, FloatArray strike, FloatArray years, FloatArray callResult) {
+        int option = ctx.groupIdx;
+        int tid = ctx.localIdx;
+
+        float s = spot.get(option);
+        float k = strike.get(option);
+        float t = years.get(option);
+
+        float dt = t / STEPS;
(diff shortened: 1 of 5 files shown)


mikepapadim:
Extended with an **int8 tensor-core stage**, completing the precision matrix (fp32 / fp16 / bf16 / int8 / FP8).

### Updated kernel-only results

RTX 3070 (sm_86), CUDA 13.0, M=N=K=1024, mean of 20 steady-state iterations via `--enableProfiler console`:

```
kernel                 device time    vs fp32
gemmFp32Tiled            1334 us       1.00x
gemmBf16MMA               771 us       1.73x
gemmFp16MMA               757 us       1.76x
gemmInt8MMA               446 us       2.99x
gemmFp16MMACpAsync        327 us       4.09x
```

int8 lands between the 16-bit formats and the `cp.async`-pipelined fp16 kernel, which is the expected shape: it halves the operand bytes *and* doubles the K-depth per instruction (`M16N8K32` vs `M16N8K16`), but still stages its tiles through the same synchronous global→shared path that `cp.async` avoids.

int8 reports **`maxErr=0.0000`** — integer GEMM is exact, so that is a real correctness check on the fragment layout, not a rounding measurement.

### FP8 tile layout corrected

The FP8 B-tile packing was wrong: it packed 4 consecutive J-values per int, while the validated int8 kernel — same `M16N8K32` shape, same 8-bit fragment tiling — packs **2 K-values × 2 J-values**. FP8 can't be validated on this Ampere hardware (needs compute capability ≥ 8.9), so deriving its layout from the passing int8 path is the only sound basis available. The code now states that provenance explicitly rather than implying the layout was verified.

### On the multi-stream part of the ask

Investigated and documented rather than implemented, because there is nothing to implement against: TornadoVM's CUDA backend **already** runs transfers and compute on separate, NVTX-labelled CUDA streams (`DEFAULT` / `H2D` / `COMPUTE` / `D2H`, see `CUDACommandQueue`), but that pipeline is managed internally. There is no user-facing stream API on `TornadoExecutionPlan`, and `withConcurrentDevices()` governs multi-*device* execution, not multi-stream within one GPU. The closest user-visible control over that pipeline is `DataTransferMode` per argument — whose cost this benchmark already quantifies (it is exactly what makes the end-to-end numbers transfer-dominated).

mikepapadim:
Final addition: an **XOR-swizzled fp16 MMA stage** (`swizzleStoreFp16Stride32` + `mmaLoadBSwizzled`), which covers the last part of the `KernelContext` MMA surface the benchmark hadn't touched.

### Complete results

RTX 3070 (sm_86), CUDA 13.0, M=N=K=1024, kernel-only time via `--enableProfiler console`, mean of 20 steady-state iterations:

```
kernel                 device time    vs fp32
gemmFp32Tiled            1336 us       1.00x
gemmBf16MMA               772 us       1.73x
gemmFp16MMASwizzled       764 us       1.75x
gemmFp16MMA               761 us       1.76x
gemmInt8MMA               449 us       2.98x
gemmFp16MMACpAsync        329 us       4.06x
```

### The swizzle stage is a negative result, and it's kept deliberately

764 vs 761 µs is inside run-to-run noise — swizzling buys nothing here. That's worth shipping rather than dropping, because the *reason* is the useful part: swizzling pays off when the shared-memory layout would otherwise make a warp's `ldmatrix` reads collide on one bank, but this benchmark's plain path already stores B int-packed (two fp16 per 32-bit word), which sidesteps most of that on its own. The swizzled variant keeps B as native fp16 elements and spends the swizzle undoing the conflicts *that* layout reintroduces. The two roughly cancel.

Documented so a reader reaches for `swizzleStoreFp16Stride32` when their tile layout actually conflicts, rather than as a reflexive optimisation. Correctness is unchanged (`maxErr=0.0060`, identical to plain fp16).

### PR is complete

The precision matrix is fully covered (fp32 / fp16 / bf16 / int8 / FP8), both concurrency mechanisms are addressed (CUDA Graphs implemented; multi-stream investigated and documented as having no user-facing API), and every number is cross-validated by two independent profilers. Ready for review.

mikepapadim:
Consolidated: the tiled N-body example (originally #1021) now lives in this PR, so all CUDA-samples-derived compute examples land together. #1021 is closed in favour of this one.

### What's in the PR now

**1. `LowPrecisionGemmBenchmark`** (+ `FP8GemmStage`) — seven GEMM implementations through `KernelContext`, timed back-to-back:

```
kernel                 device time    vs fp32
gemmFp32Tiled            1336 us       1.00x
gemmBf16MMA               772 us       1.73x
gemmFp16MMASwizzled       764 us       1.75x
gemmFp16MMA               761 us       1.76x
gemmInt8MMA               449 us       2.98x
gemmFp16MMACpAsync        329 us       4.06x
```
plus FP8 E4M3, which correctly reports as skipped on sm_86 (needs compute capability ≥ 8.9).

Covers `mma` / `mmaBF16` / `mmaInt8` / `mmaFP8E4M3`, the swizzle helpers, `cp.async`, and `withCUDAGraph()`.

**2. `NBodyTiled`** — all-pairs N-body, naive vs shared-memory tiled, bit-exact against a sequential reference:

```
nBodyNaive     25648 us
nBodyTiled     25411 us     1.01x
```

Covers `allocateFloatLocalArray` + double `localBarrier` outside a GEMM.

### Both examples carry documented negative results

Neither the XOR swizzle (1.75x vs plain fp16's 1.76x) nor the N-body tiling (1.01x) buys anything on this hardware. Both are kept with the *reason* documented — the swizzle because the int-packed layout already avoids the bank conflicts it exists to fix, the N-body tiling because the kernel is SFU-bound on `sqrt` rather than memory-bound, and Ampere's cache already supplies the reuse. They're useful precisely because they show when a textbook optimisation doesn't transfer.

### Measurement issues caught along the way

- End-to-end wall-clock understated the GEMM kernel speedups by 2-3x (D2H transfer dominated) — found with nsys.
- A naive FP8 timing reported a fictional **33 TFLOP/s** on hardware that can't do FP8, because an unsupported kernel silently bails out to sequential host Java rather than throwing. `FP8GemmStage` now validates output before printing any number.
- The FP8 tile layout was wrong (4 consecutive J-values per int); corrected against the validated int8 path, since FP8 can't be validated on sm_86.
- `nsys` silently omitted `gemmBf16MMA` from its kernel summary despite the kernel demonstrably running — cross-checked with `--enableProfiler console`.
- **The existing `NBody` example has a data race**: in-place position updates with no grid-wide barrier between force accumulation and integration. Easy to mistake for float drift; ruled out because the 0.18 deviation was ~4 orders of magnitude larger than the entire per-step position update. `NBodyTiled` is double-buffered and reproduces the reference exactly.

mikepapadim:
Added a third example: **`BinomialOptions`** — Cox-Ross-Rubinstein binomial-tree option pricing.

This one is here for a **parallel decomposition the repo doesn't otherwise have**: one block prices one option, each thread owns one tree node, and the block walks the tree backwards collapsing one level per iteration. Every other `KernelContext` example maps one thread to one output element and uses the block purely as a staging area — this inverts that.

Both barriers per level are load-bearing: the update reads `tree[tid]` and `tree[tid+1]` and writes `tree[tid]`, so dropping the second lets thread `tid` clobber `tree[tid]` while thread `tid-1` still needs it as its neighbour. Keeping the new value in a register and publishing only at the top of the next iteration is what makes it correct.

### Measured — and it's a third negative result

RTX 3070, CUDA 13.0, 2048 options, 255 steps, kernel-only:

```
block-per-option (GPU)    35265 us
sequential Java (1 core)  44561 us   ->  1.25x
```

1.25x for 524288 GPU threads is a poor return, and it's structural rather than tunable:

- 2048 blocks × 510 barriers is **over a million block-wide syncs**, with only two multiplies and an add per node to hide them behind
- The active thread count decays 255 → 1, so half the block is idle on average and the last levels run a 256-thread block to compute a handful of values
- Kernel time (35.3 ms) is essentially the whole end-to-end time (35.6 ms), so data movement isn't the cause

Kept for the pattern, not the throughput. The real CUDA sample amortises this differently — far more options in flight and a cache-blocked sweep instead of lockstep level-by-level.

### The `@Parallel` baseline deliberately does not run

One-option-per-thread needs a 256-float private tree = **1 KB per thread**, which exceeds the per-thread local-memory budget; the launch is rejected with `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES`.

A rejected launch leaves the output untouched and therefore *times as absurdly fast* (0.1 ms — it never ran). Same trap as the FP8 stage earlier in this PR, so it gets the same treatment: the result is validated before any timing is printed, and the failure is reported instead. That constraint is precisely what motivates the block-cooperative version — the tree is too big to be private per thread, but fits in shared memory when a block shares one copy.

### PR now contains

| Example | Pattern |
|---|---|
| `LowPrecisionGemmBenchmark` (+`FP8GemmStage`) | fp16/bf16/int8/FP8 tensor cores, cp.async, CUDA Graphs, swizzle |
| `NBodyTiled` | Shared-memory tiled all-pairs |
| `BinomialOptions` | Block-per-work-item wavefront |

mikepapadim:
Added a fourth example: **`Fdtd3dStencil`** — radius-4 symmetric star stencil (25-point), naive vs XY shared-memory tiled. It's the repo's first *tiled* stencil; the existing `stencil` benchmark is an untiled 7-point Jacobi.

I picked this specifically as the candidate **most likely to show a positive result** — stencils are genuinely memory-bound (~25 loads per ~25 FLOPs), which is exactly the case where shared-memory tiling should pay. It doesn't:

```
fdtdNaive   155.3 us   2148 GiB/s effective
fdtdTiled   183.2 us   1821 GiB/s effective   0.85x
```

Both bit-correct against the sequential reference (`maxDev = 1e-6`).

### The bandwidth number settles it

At `6*RADIUS+2` accesses per interior point, the naive kernel is moving the equivalent of **2148 GiB/s** — about **five times** the RTX 3070's ~448 GB/s of DRAM bandwidth. A kernel can't exceed DRAM bandwidth, so most of those accesses never reach DRAM: **L1/L2 is already absorbing the stencil's reuse.** Manual staging then buys no traffic reduction while adding two barriers per Z step, halo indexing, and extra work for the first `RADIUS` threads of each row and column.

### The through-line across this PR

Four ports, four cases where a textbook CUDA optimisation doesn't pay on Ampere:

| Optimisation | Result | Why |
|---|---|---|
| XOR swizzle (GEMM) | 1.75x vs 1.76x | int-packed layout already avoids the bank conflicts |
| Tiled N-body | 1.01x | SFU-bound on `sqrt`, not memory-bound |
| Block-per-option binomial | 1.25x vs 1 CPU core | barrier- and divergence-bound |
| Tiled FDTD stencil | 0.85x | cache already absorbs the reuse |

The N-body and FDTD results share one cause and it's worth stating plainly: **on Ampere-class hardware the cache hierarchy has largely absorbed the reuse that manual shared-memory tiling was invented to provide.** These optimisations were formulated for GPUs whose caches were far weaker relative to their arithmetic. Tiling still matters when the working set genuinely doesn't fit or the access pattern defeats the cache — but it's no longer a reflexive win, and it needs measuring rather than assuming.

All four examples are kept, with the reasoning documented in each. They're useful precisely as calibration: they show *where* the classic advice still applies and where it has quietly expired.

### Also a sharp lesson on end-to-end timing

FDTD is an extreme case: wall-clock is ~11 ms per execution, of which the kernel is **~0.16 ms** — the rest is copying the 16 MB result volume back. End-to-end suggests ~7x over sequential Java; the kernels are closer to **500x**. For a short kernel over a large array, end-to-end timing measures PCIe, not the GPU.

stratika (approved):

PR #1021, opened 2026-08-03
Add shared-memory tiled N-body example

mikepapadim:
## Summary

Ports the classic CUDA all-pairs N-body **shared-memory tiling** optimisation that the existing `kernelcontext.compute.NBody` example omits (that one is naive-only — every thread streams all N bodies straight from global memory). The new example runs naive and tiled back-to-back on identical data so the comparison is direct.

```
for each tile:
    shared[localIdx] = body[tile * TILE + localIdx]    // one cooperative load
    localBarrier()
    for j in 0..TILE:  accumulate force from shared[j] // TILE reuses per load
    localBarrier()
```

Global loads per thread drop from N to N/TILE; the arithmetic is unchanged.

## The tiling does not help on Ampere — and the example says so

RTX 3070 (sm_86), CUDA 13.0, 16384 bodies, TILE=256, kernel-only via `--enableProfiler console`, mean of 10 steady-state iterations:

```
nBodyNaive     25648 us
nBodyTiled     25411 us     1.01x
```

A 1% difference is noise. Two reasons, both documented in the example:

1. **The kernel isn't memory-bound to begin with.** Both variants sustain only ~210 GFLOP/s against an RTX 3070's multi-TFLOP fp32 peak, which points at the special-function unit: every pair interaction costs one `sqrt`, and 268M of them in ~25 ms sits near the SFU throughput ceiling. Relieving memory pressure can't speed up a kernel waiting on the transcendental pipeline.
2. **Ampere's L1/L2 already provides the reuse** that tiling was invented to provide, for a working set this small. The textbook win for this optimisation was measured on GPUs whose caches were far weaker than their arithmetic.

The tiled kernel is kept regardless: it's still the correct way to write this when N or the body record outgrows the cache, and the naive/tiled pair is a clean demonstration of `allocateFloatLocalArray` + double `localBarrier` outside a GEMM. The point is that the speedup shouldn't be assumed without measuring.

## A data race worth flagging

Both kernels here are **double-buffered** — they read `posIn`/`velIn` and write `posOut`/`velOut`, never both.

An in-place version is a data race: there is no grid-wide barrier between the force loop and the integration step, so a thread can overwrite a body's position while other threads are still reading it for their own force sums. **The existing `NBody` example has this structure** (and validates with a `0.1` tolerance, which absorbs it).

This is easy to misread as floating-point drift. It isn't — I initially assumed exactly that, and the numbers ruled it out: the observed 0.18 deviation was far larger than the entire per-step position update (~3e-5), so it couldn't be rounding. With double-buffering both kernels reproduce the sequential reference **exactly** (`maxDev = 0.00000` — bit-exact, since the summation order is identical).

The sequential reference also had to be fixed to compute forces against a *snapshot* of the input positions, for the same reason: updating in place lets body *i+1* see body *i*'s already-integrated position, which is a different algorithm.

## Test plan

- [x] `make BACKEND=cuda` builds clean
- [x] Both kernels validate bit-exactly against the sequential Java reference (`maxDev = 0.00000`)
- [x] Timing separates first-run (JIT/NVRTC) from steady-state; kernel-only figures cross-checked with `--enableProfiler console`
- [x] Examples-only change — no runtime, driver, or API code touched

### DIFF
--- a/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/kernelcontext/compute/NBodyTiled.java
+++ b/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/kernelcontext/compute/NBodyTiled.java
@@ -1,0 +1,397 @@
+/*
+ * Copyright (c) 2026 APT Group, Department of Computer Science,
+ * The University of Manchester.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.manchester.tornado.examples.kernelcontext.compute;
+
+import uk.ac.manchester.tornado.api.GridScheduler;
+import uk.ac.manchester.tornado.api.KernelContext;
+import uk.ac.manchester.tornado.api.TaskGraph;
+import uk.ac.manchester.tornado.api.TornadoExecutionPlan;
+import uk.ac.manchester.tornado.api.WorkerGrid1D;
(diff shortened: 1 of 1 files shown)


mikepapadim:
Consolidated into #1020 so all CUDA-samples-derived compute examples review together. The tiled N-body commit is cherry-picked there unchanged; nothing is lost.

PR #1022, opened 2026-08-05
[cuda] Skip the per-launch kernel stack-frame upload when the device copy is unchanged (~2x on short task graphs)

mikepapadim:
Two changes to the same thing: the fixed cost the host pays per operation, independent of how many bytes move. They were #1022 and #1023; #1023 is folded in here because its value is only measurable on top of this one, and reviewing them apart means reading the same profile twice.

## How this was found

JFR (1 ms sampling, `stackdepth=192`, `DebugNonSafepoints`) plus Nsight Systems on `develop` @`ba28f1526`. A `saxpy` task graph over 512 floats — one kernel, one H2D, one blocking D2H — spent **17.46 µs per `execute()` while the GPU did 3.31 µs of work**. Per iteration the runtime issued **4 stream operations, 12 event calls and 3 `cuStreamSynchronize`** to move 2 KB and run a 1.1 µs kernel.

Worth recording from the same profile: the Java bytecode interpreter and its data structures account for **~0.9% of wall time** (118 `jdk.ExecutionSample` vs 13,545 `jdk.NativeMethodSample`, and most of the Java samples are the benchmark harness's own `Arrays.sort`). The cost is in the per-operation CUDA dispatch sequence, not in interpreting bytecodes.

## Commit 1 — skip the per-launch kernel stack-frame upload

The 24-byte kernel-argument stack frame is re-uploaded on **every** kernel launch, even though its contents are identical across iterations in the common case (no `KernelContext`, so all three slots are zero). Nsight Systems puts that copy at **2.00 µs per launch on an RTX 4090 — the same cost as the 2 KB user-data copy in the same graph**, because the price is the JNI round trip, the `cuMemcpyHtoDAsync` and an event pair, not the bytes.

`CUDAKernelStackFrame` now keeps a snapshot of the slots it last placed on the device, together with the execution-plan id that put them there. `enqueueWrite` compares the host frame against that snapshot and returns `-1` — the established "no event" marker in dependency lists, already skipped by `CUDAEventPool.serialiseEvents` — instead of issuing the transfer when they match.

`CUDAInstalledCode` falls back to the caller's dependency list when there is no write event, so dependencies are never dropped: previously the launch chained onto the write event, which transitively carried the caller's dependencies.

The skip is refused when the plan runs with intra-plan concurrency (`isMultiStreamEnabled`), where the launch may issue on a different queue from the one that carried the write and the write event is load-bearing for ordering.

## Commit 2 — skip stream synchronisation when nothing has been enqueued

A task-graph execution ends with a blocking device-to-host read-back, then `flush()`, then `waitOn()` — **three `cuStreamSynchronize` per execution, two of them issued against a stream that is already empty**.

`cuda_queue_t` gains a `pending` flag, set by every stream enqueue (both transfer helpers, `cuLaunchKernel`, every `cuEventRecord`, `cuStreamWaitEvent`, `cuGraphLaunch`) and cleared on drain. `clFlush`, `clFinish` and the blocking-transfer sync all route through a single `sync_stream()` helper that returns immediately when nothing has been enqueued since the last drain — in which case the stream is empty by construction and the driver round trip buys nothing.

The second half of that commit matters as much as the flag: the transfer helpers now record their completion event **before** draining rather than after. Recording afterwards re-armed `pending`, so the following `flush()` still synchronised an otherwise idle stream. With the reorder the sync covers the event record too, and `pending` is genuinely clear afterwards. The semantics are unchanged — the event still marks the end of the copy.

`sync_stream()` only ever skips when the flag says nothing has been enqueued since the previous successful drain. Every path that puts work on the stream sets it, including `cuStreamWaitEvent` (a wait node is stream work) and `cuGraphLaunch` (a replayed graph is stream work like any other). `cuMemcpyDtoD` in `CUDANativeCommandQueue.cpp` is the synchronous variant and leaves nothing outstanding. Capture is unaffected: the helpers already refuse to synchronise a capturing stream.

## Numbers

RTX 4090, JDK 21.0.2, base `develop` @`ba28f1526`. Per-`execute()` median reported by `BenchmarkDriver` (which already discards the first 30 iterations), `saxpy 100000 512`:

| build | median |
|---|---|
| `develop` | 17.46 µs |
| commit 1 only | 8.75 µs |
| commit 2 only | 17.44 µs |
| **both (this PR)** | **8.35 / 8.36 / 8.60 µs** |

**2.09x.** The second commit is worth nothing on its own on this workload and **4.2% once the first lands** — which is the argument for shipping them together: the removed syncs only surface as time when the dominant per-launch cost is gone.

nsys, 20,000 iterations, per iteration:

| | `develop` | this PR |
|---|---|---|
| NVTX `H2D 24 B` instances | 20,000 (2.00 µs avg) | **1** |
| `cuStreamSynchronize` calls | 60,000 (3/iter) | **20,000 (1/iter)** |
| `cuEventCreate` / `Record` / `Destroy` | 4 each | 3 each |
| NVTX `D2H 2.0 KB` avg | 7.81 µs | 5.63 µs |

The saving is larger than the 2.00 µs the copy itself cost: removing it also removes an event pair and shortens the queue that the blocking read-back waits behind.

Across the small-size benchmarks the win tracks how dispatch-dominated the graph is, and nothing regresses:

| benchmark (size 512) | `develop` | this PR | speed-up |
|---|---|---|---|
| `saxpy` | 18.33 µs | 9.38 µs | 1.95x |
| `stencil` | 17.24 µs | 13.24 µs | 1.30x |
| `blackscholes` | 22.56 µs | 20.39 µs | 1.11x |
| `nbody` | 40.22 µs | 39.22 µs | 1.03x |
| `dft` | 95.83 µs | 94.18 µs | 1.02x |
| `blurFilter 3000 256` (3 tasks) | 266.59 µs | 262.56 µs | 1.015x |

**Controls — GPU-bound work unaffected:**

| workload | `develop` | this PR |
|---|---|---|
| `saxpy 2000 16777216` | 7.542 / 7.530 / 7.518 ms | 7.456 / 7.392 / 7.425 ms |
| `nbody 500 16384` | 1.7222 / 1.7239 / 1.7260 ms | 1.7261 / 1.7252 / 1.7270 ms |

## Testing

- `tornado-test --quickPass`, CUDA backend: same failures as the `develop` baseline on this machine (86 vs 85, the difference being the known-flaky `MMwithBytes`, 4/4 in isolated re-runs).
- CUDA-graph capture exercised specifically, since both changes touch state that must survive stream capture (`cuGraphLaunch` sets `pending`; the deferred frame write must still see a valid frame): `TestCuBlas` 12/12, `TestCuFft` 8/8, `TestExecutor` 7/7.

## Known limitation

A generated kernel with a non-void return writes `frame[0]` through `CUDAReturnSlot`, so its device copy could drift from the host snapshot. This is only reachable via the internal JIT test path — the `TaskGraph.task(...)` functional interfaces are all `void` — and nothing on the CUDA path reads the frame back.

## Reproduction

```bash
tornado --jvm="-Dtornado.benchmarks.skipserial=True" \
  --params="saxpy 100000 512"

nsys profile -t cuda --sample=none --cpuctxsw=none -o saxpy512 \
  tornado --jvm="-Dtornado.benchmarks.skipserial=True" \
  --params="saxpy 20000 512"
nsys stats --report cuda_api_sum saxpy512.nsys-rep
```

Part of #1028. Supersedes #1023.

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDADeviceContext.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDADeviceContext.java
@@ -386,7 +386,7 @@ public class CUDADeviceContext implements CUDADeviceContextInterface {
      * Multi-queue issue is disabled while the plan's stream is capturing into a CUDA graph:
      * capture records a single stream, and touching other streams would invalidate it.
      */
-    private boolean isMultiStreamEnabled(long executionPlanId) {
+    public boolean isMultiStreamEnabled(long executionPlanId) {
         return intraPlanConcurrencyPlans.contains(executionPlanId) && !isStreamCapturing(executionPlanId);
     }
 
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/CUDAInstalledCode.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/CUDAInstalledCode.java
@@ -226,9 +226,16 @@ public class CUDAInstalledCode extends InstalledCode implements TornadoInstalled
         if (deviceContext.deferKernelContextWriteIfCapturing(executionPlanId, kernelArgs)) {
             waitEvents = events;
         } else {
-            internalEvents[0] = kernelArgs.enqueueWrite(executionPlanId, events);
-            waitEvents = internalEvents;
-            updateProfilerKernelContextWrite(executionPlanId, internalEvents[0], meta, kernelArgs);
+            int kernelContextWriteEventId = kernelArgs.enqueueWrite(executionPlanId, events);
+            if (kernelContextWriteEventId == -1) {
+                // The frame was already on the device unchanged, so there is no write event to
+                // chain the launch onto: keep the caller's dependencies instead of dropping them.
+                waitEvents = events;
+            } else {
+                internalEvents[0] = kernelContextWriteEventId;
+                waitEvents = internalEvents;
+                updateProfilerKernelContextWrite(executionPlanId, kernelContextWriteEventId, meta, kernelArgs);
+            }
         }
 
         int task;
@@ -314,7 +321,9 @@ public class CUDAInstalledCode extends InstalledCode implements TornadoInstalled
         setKernelArgs(oclKernelStackFrame, atomicSpace, meta);
         if (!deviceContext.deferKernelContextWriteIfCapturing(executionPlanId, oclKernelStackFrame)) {
             int kernelContextWriteEventId = oclKernelStackFrame.enqueueWrite(executionPlanId);
-            updateProfilerKernelContextWrite(executionPlanId, kernelContextWriteEventId, meta, oclKernelStackFrame);
+            if (kernelContextWriteEventId != -1) {
+                updateProfilerKernelContextWrite(executionPlanId, kernelContextWriteEventId, meta, oclKernelStackFrame);
+            }
         }
 
         if (meta == null) {
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/mm/CUDAKernelStackFrame.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/mm/CUDAKernelStackFrame.java
@@ -57,6 +57,19 @@ public class CUDAKernelStackFrame extends CUDAByteBuffer implements KernelStackF
     /** Off-heap host buffer backing {@code buffer}, or 0 when allocation fell back to heap. */
     private long hostPointer;
 
+    /**
+     * Snapshot of the frame slots as they currently sit in device memory, and the execution plan
+     * that put them there. The frame is re-uploaded on every launch even though its contents are
+     * identical across iterations for the common case (no {@code KernelContext}, so all slots are
+     * zero), which costs a {@code cuMemcpyHtoDAsync} plus an event pair per launch. Comparing the
+     * host slots against this snapshot lets an unchanged frame skip the transfer entirely.
+     */
+    private final long[] deviceSlots = new long[RESERVED_SLOTS];
+
+    private boolean deviceSlotsValid;
+
+    private long deviceSlotsPlanId;
+
     CUDAKernelStackFrame(long bufferId, int numArgs, CUDADeviceContext device) {
         super(device, bufferId, 0, FRAME_BYTES);
         this.callArguments = new ArrayList<>(numArgs);
@@ -90,6 +103,7 @@ public class CUDAKernelStackFrame extends CUDAByteBuffer implements KernelStackF
     @Override
     public void invalidate() {
         isValid = false;
+        deviceSlotsValid = false;
         if (hostPointer != 0) {
             // Unregister (native side synchronises the context first) BEFORE freeing, so
             // no stale pin can survive the host buffer.
@@ -100,13 +114,22 @@ public class CUDAKernelStackFrame extends CUDAByteBuffer implements KernelStackF
         deviceContext.getPlatformContext().releaseBuffer(toBuffer());
     }
 
-    /** Async device write from the off-heap (pinned) host frame: no per-launch stream sync. */
+    /**
+     * Async device write from the off-heap (pinned) host frame: no per-launch stream sync.
+     * Returns {@code -1} when the device copy already holds these slots and the transfer was
+     * skipped; that value is the established "no event" marker in dependency lists.
+     */
     @Override
     public int enqueueWrite(long executionPlanId, final int[] events) {
         if (hostPointer == 0) {
             return super.enqueueWrite(executionPlanId, events);
         }
-        return deviceContext.enqueueWriteBuffer(executionPlanId, toBuffer(), getOffset(), FRAME_BYTES, hostPointer, 0, events);
+        if (deviceCopyIsUpToDate(executionPlanId)) {
+            return -1;
+        }
+        int event = deviceContext.enqueueWriteBuffer(executionPlanId, toBuffer(), getOffset(), FRAME_BYTES, hostPointer, 0, events);
+        snapshotDeviceCopy(executionPlanId);
+        return event;
     }
 
     /** Blocking device write from the off-heap host frame (used by the CUDA-graph capture flush). */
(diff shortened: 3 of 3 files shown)


mikepapadim:

|---|---|---|---|

mikepapadim:

mikepapadim:
## Rebased onto `develop` — and the JNI→FFM migration changed what this PR can carry

@stratika @mairooni @orionpapadakis — flagging this one as **critical perf**, because the regression it fixes now ships in a release (details at the bottom). Would appreciate a look.

### What the rebase did

`develop` had moved 282 commits since the merge base and the PR was `dirty`. It is now rebased and **mergeable** again, but two of the three commits did not survive, so the PR is deliberately smaller than it was:

| Commit | Outcome |
|---|---|
| `[cuda] Skip the per-launch kernel stack-frame upload when the device copy is unchanged` | **Rebased cleanly.** Java-only; all three target files still exist. This is the PR now. |
| `[cuda] Skip stream synchronisation when nothing has been enqueued since the last drain` | **Dropped — cannot be rebased.** |
| `[docs] Add a tornadovm-perf-campaign skill` | **Dropped — already on `develop`**, in a newer revision than this branch carried. |

Previous head was `b083138e4`, for anyone who wants the original three commits.

### Why the stream-sync commit had to go

Its entire implementation surface no longer exists. `2a680067f [cuda] Remove the tornado-cuda JNI module now that the backend runs entirely on FFM` deleted all four files it patched (`CUDACommandQueue.cpp`, `CUDAContext.cpp`, `CUDAGraph.cpp`, `cuda_jni.h`), so it conflicts as modify/delete against every one of them.

It needs a **re-implementation against FFM, not a conflict resolution**, and that is a design call rather than a mechanical port, which is why I have not just rewritten it here:

- The change tracked a `pending` flag on `cuda_queue_t`, set wherever work was enqueued and cleared on drain, gating `cuStreamSynchronize`.
- The FFM shape maps cleanly at the call sites — `clFlush` and `clFinish` both funnel into `CUDACommandQueue.synchronizeStream(long)`, which is exactly the pair the C++ patched.
- The obstacle is state ownership: `CUDAHandles.Queue` is a `record`, so there is nowhere to hang a mutable per-queue flag without either converting it to a class or keeping side state keyed by stream. `CUDACommandQueue` already has a separate `dirty`/`pollDirty` pair, but that is per-plan join bookkeeping for secondary queues under intra-plan concurrency, not the same thing — reusing it would conflate the two.

Happy to do that port if you tell me which way you want the state held; otherwise it probably belongs in its own PR against FFM.

### One caveat on the commit that remains

The rebase is clean and I verified every symbol it leans on still exists on `develop` (`isMultiStreamEnabled`, now widened to `public`, `hostPointer`, `enqueueWriteBuffer`, `RESERVED_SLOTS`). **I was not able to run a build to confirm it compiles** — `mvnw -P cuda-backend` fails in `tornado-api` on this machine for unrelated toolchain reasons (`-source 8`, then JVMCI module visibility). So treat "rebases cleanly" as exactly that, and let CI be the judge.

The correctness argument is unchanged and worth restating, since it is the subtle part: the skip is claimed only when the plan issues on a single stream (`!isMultiStreamEnabled`), because under intra-plan concurrency the launch may run on a different queue from the one that carried the write, and skipping would drop the ordering the write event provides.

### New finding: this regression ships in v6.0.0

The evidence in #1028 was all gathered on `develop`. I have now measured the **released** SDK, and the per-execution dispatch sequence is still there — v6.0.0 was cut on 2026-09-02 with this PR still open, so users have it today.

On `6.0.0-jdk22plus-cuda`, RTX 4090 / driver 565.57.01 / CUDA 12.6.85 / JDK 25.0.2, profiling one task graph with one trivial kernel and differencing 10 → 100 executions under `nsys` (so fixed start-up cancels and the slope is the true per-execution cost):

| CUDA API | per execution | ns per execution |
|---|---|---|
| `cuStreamSynchronize` | **3.00** | 2,751 |
| `cuLaunchKernel` | 1.00 | 2,256 |
| `cuMemcpyDtoHAsync_v2` | 1.00 | 1,535 |
| `cuMemcpyHtoDAsync_v2` | **1.00** | 1,232 |
| `cuEventCreate` | 3.00 | 1,127 |
| `cuCtxSetCurrent` | 3.00 | 211 |
| `cuEventRecord` | 3.00 | 179 |
| `cuEventDestroy_v2` | 3.00 | 128 |
| `cuStreamIsCapturing` | 1.00 | 36 |

Call counts are exactly linear in execution count. That single per-execution H2D is the 24-byte kernel argument stack frame — finding #1 of #1028, which this PR removes. The three `cuStreamSynchronize` are finding #2, which is the commit that just had to be dropped.

**~9.5 µs of CUDA API time per execution**, on a graph whose kernel is trivial.

Two things I got wrong first time and corrected, in case they save someone else the detour:

- A single trace **cannot** separate start-up from per-execution cost. I initially read a fixed start-up cost as per-execution and got a number roughly 12x too large. Differencing two execution counts is what makes it sound.
- On a real workload, most of what looks like sync overhead is not overhead. On a warp-shuffle reduction demo, the 54,446 ns/execution in `cuStreamSynchronize` is the host legitimately blocking on the kernel — its two kernels average 54,792 ns, matching to 0.6%. Excluding that genuine wait, real per-execution driver overhead is ~8.3 µs.

Minimal reproducer (~50 lines, no `BenchmarkRunner`) and the raw `nsys` output are here, if it is useful for validating the fix:
https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/results/raw/25-host-dispatch-breakdown

For contrast, on the same 40 launches the hand-written CUDA equivalent of that demo makes **41 memory-transfer calls against TornadoVM's 1,620**, with no events and no stream synchronisation at all.

mikepapadim:
## Now built and tested — no regressions against a same-session `develop` baseline

My previous comment said I could not run a build and asked you to let CI be the judge. That caveat is retired: the build failure was mine, not the tree's. `./mvnw -P cuda-backend` activates only the backend profile, so no `jdkNN` profile supplies `<source>`/`<target>` or the JVMCI `--add-exports` — hence `-source 8` and then a module-visibility error. Two symptoms, one cause. The working invocation on a JDK 25 `JAVA_HOME` is `make jdk22plus BACKEND=cuda` (not `make BACKEND=cuda`, whose `build` target hardcodes `--jdk jdk21`).

### Method

`cfdd7c498`'s parent is exactly `origin/develop` @ `5bbe99cb0`, so this is a clean one-commit A/B. Both sides built and tested in the same session, same machine, same toolchain, same command — RTX 4090, driver 565.57.01, CUDA 12.6.85, JDK 25.0.2, `tornado 6.0.1-dev`, CUDA backend.

```
make jdk22plus BACKEND=cuda && source setvars.sh && tornado-test --quickPass
```

### Result

| | classes | tests ran | failed | unsupported |
|---|---|---|---|---|
| `develop` @ `5bbe99cb0` | 168 | 1164 | 7 | 77 |
| this PR @ `cfdd7c498` | 168 | 1164 | 7 | 77 |

Both builds are `BUILD SUCCESS` with zero skipped or failed modules.

Identical in aggregate, and identical per class. I then re-ran the four failing classes with `-V` on both builds and diffed the failing **method** names, since matching counts alone would not rule out a different test failing:

```
reductions.TestReductionsFloats                        :: testComputePi
kernelcontext.matrices.TestMatrixMultiplicationKernelContext :: mxm2DKernelContext01
kernelcontext.matrices.TestMatrixMultiplicationKernelContext :: mxm2DKernelContext02
compute.ComputeTests                                   :: testMandelbrot
compute.ComputeTests                                   :: testJuliaSets
api.TestDevices                                        :: test04
api.TestDevices                                        :: test06
```

`diff` between the two runs is empty. All seven are pre-existing on `develop` and none are in a path this PR touches.

I flagged `TestMatrixMultiplicationKernelContext` to myself as the one to watch — 2 of 3 failing, and the closest class in the suite to anything this PR changes. It fails identically on `develop`, so it is unrelated, but it is probably worth someone looking at separately.

### Graph-capture paths specifically

This PR touches state that has to survive stream capture, so those deserved explicit exercise rather than trusting the aggregate:

| | |
|---|---|
| `TestCuBlas` | 14 ran, 0 failed |
| `TestCuFft` | 8 ran, 0 failed |
| `TestExecutor` | 7 ran, 0 failed (not part of `--quickPass`; run separately) |

### Unchanged

The correctness argument from the previous comment stands: the skip is claimed only when the plan issues on a single stream (`!isMultiStreamEnabled`), because under intra-plan concurrency the launch may run on a different queue from the one that carried the write.

The dropped stream-sync commit still needs a re-implementation against FFM rather than a conflict resolution, and still needs your call on where the per-queue `pending` state should live given `CUDAHandles.Queue` is a `record`.

mikepapadim:
@stratika @mairooni can we merge this?

stratika (approved):
LGTM

PR #1023, opened 2026-08-05
[cuda] Skip stream synchronisation when nothing has been enqueued since the last drain (3 syncs per execution to 1)

mikepapadim:
## Summary

A task-graph execution ends with a blocking device-to-host read-back, then `flush()`, then `waitOn()` — **three `cuStreamSynchronize` per execution, two of them issued against a stream that is already empty**.

`cuda_queue_t` gains a `pending` flag, set by every stream enqueue (both transfer helpers, `cuLaunchKernel`, every `cuEventRecord`, `cuStreamWaitEvent`, `cuGraphLaunch`) and cleared on drain. `clFlush`, `clFinish` and the blocking-transfer sync all route through a single `sync_stream()` helper that returns immediately when nothing has been enqueued since the last drain — in which case the stream is empty by construction and the driver round trip buys nothing.

The second half of the change matters as much as the flag: the transfer helpers now record their completion event **before** draining rather than after. Recording afterwards re-armed `pending`, so the following `flush()` still synchronised an otherwise idle stream. With the reorder the sync covers the event record too, and `pending` is genuinely clear afterwards. The semantics are unchanged — the event still marks the end of the copy, and syncing after recording it is if anything the more obvious ordering.

## Numbers

RTX 4090, JDK 21.0.2, base `develop` @`ba28f1526`. `saxpy 100000 512`, per-`execute()` median, 5 runs.

**On its own**, on this workload, the removed syncs were near-no-ops and the wall-clock is unchanged:

| build | runs (µs) | median |
|---|---|---|
| `develop` | 17.46, 17.46, 17.48, 17.50, 17.55 | 17.46 µs |
| this PR alone | 17.43, 17.43, 17.45, 17.45, 17.45 | 17.44 µs |

What it does do on its own is halve the syscall traffic — nsys, 20 000 iterations:

| | `develop` | this PR alone |
|---|---|---|
| `cuStreamSynchronize` calls | 60 000 (3/iter) | **20 000 (1/iter)** |

The value shows up once the dominant per-launch cost is gone. Stacked on top of #1022 (skip the unchanged kernel stack-frame upload), where the iteration is 8.6 µs rather than 17.5 µs, the same two syncs are worth a measurable **4.2 %**:

| build | median |
|---|---|
| `develop` | 17.46 µs |
| + kernel-frame skip | 8.75 µs |
| + kernel-frame skip + this PR, event recorded *after* the sync | 8.54 µs |
| + kernel-frame skip + this PR (event recorded *before* the sync) | **8.38 µs** |

I am submitting it separately because it stands on its own as a correctness-neutral reduction in driver round trips, and because the `pending` flag is the piece any future work on stream/queue handling will want to build on.

## Controls — GPU-bound work unaffected

| workload | `develop` | with this PR stacked |
|---|---|---|
| `saxpy 2000 16777216` | 7.542 / 7.530 / 7.518 ms | 7.456 / 7.392 / 7.425 ms |
| `nbody 500 16384` | 1.7222 / 1.7239 / 1.7260 ms | 1.7261 / 1.7252 / 1.7270 ms |

## Testing

- `tornado-test -V --quickPass`: **1031 PASS / 9 FAILED**, identical to `develop` on the same machine. All failures whitelisted.
- CUDA-graph path exercised specifically, since `pending` has to survive stream capture and `cuGraphLaunch` (which is why `cuGraphLaunch` sets it): `TestCuBlas` 12/12, `TestCuFft` 8/8, `TestExecutor` 7/7.

## Note on correctness

`sync_stream()` only ever skips when the flag says nothing has been enqueued since the previous successful drain. Every path that puts work on the stream sets it, including `cuStreamWaitEvent` (a wait node is stream work) and `cuGraphLaunch` (a replayed graph is stream work like any other). `cuMemcpyDtoD` in `CUDANativeCommandQueue.cpp` is the synchronous variant and leaves nothing outstanding. Capture is unaffected: the helpers already refuse to synchronise a capturing stream.

## Reproduction

```bash
tornado --jvm="-Dtornado.benchmarks.skipserial=True" \
  --params="saxpy 100000 512"

nsys profile -t cuda --sample=none --cpuctxsw=none -o saxpy512 \
  tornado --jvm="-Dtornado.benchmarks.skipserial=True" \
  --params="saxpy 20000 512"
nsys stats --report cuda_api_sum saxpy512.nsys-rep
```

### DIFF
--- a/tornado-drivers/cuda-jni/src/main/cpp/source/CUDACommandQueue.cpp
+++ b/tornado-drivers/cuda-jni/src/main/cpp/source/CUDACommandQueue.cpp
@@ -88,6 +88,7 @@ static jlong record_event(cuda_queue_t *queue) {
     LOG_CUDA_AND_VALIDATE("cuEventCreate", result);
     result = cuEventRecord(ev->event, queue->stream);
     LOG_CUDA_AND_VALIDATE("cuEventRecord", result);
+    queue->pending = true;
     return (jlong) ev;
 }
 
@@ -107,9 +108,24 @@ static void wait_events(JNIEnv *env, cuda_queue_t *queue, jlongArray array) {
         cuda_event_t *ev = (cuda_event_t *) raw[i + 1];
         if (ev != nullptr) {
             cuStreamWaitEvent(queue->stream, ev->event, 0);
+            queue->pending = true;
         }
     }
     env->ReleasePrimitiveArrayCritical(array, raw, JNI_ABORT);
+}
+
+/*
+ * Drains the stream, unless nothing has been enqueued since it was last drained - in which
+ * case the stream is empty by construction and cuStreamSynchronize would only cost a driver
+ * round trip. Never called while capturing: a sync invalidates the capture.
+ */
+static void sync_stream(cuda_queue_t *queue) {
+    if (!queue->pending) {
+        return;
+    }
+    CUresult result = cuStreamSynchronize(queue->stream);
+    LOG_CUDA_AND_VALIDATE("cuStreamSynchronize", result);
+    queue->pending = false;
 }
 
 /*
@@ -163,6 +179,7 @@ static jlong end_event(cuda_event_t *ev, cuda_queue_t *queue) {
     LOG_CUDA_AND_VALIDATE("cuEventCreate(end)", result);
     result = cuEventRecord(ev->event, queue->stream);
     LOG_CUDA_AND_VALIDATE("cuEventRecord(end)", result);
+    queue->pending = true;
     return (jlong) ev;
 }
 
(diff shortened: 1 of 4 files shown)


mikepapadim:

|---|---|---|---|

mikepapadim:

mikepapadim:
Folded into #1022 and closing this one.

Reason: on its own this change is worth **nothing measurable** on the workload it was profiled against (17.44 µs vs 17.46 µs for `saxpy 100000 512`), and **4.2%** once #1022's per-launch copy is gone. Reviewing the two apart means reading the same profile twice and judging this one against a number it cannot produce alone. #1022 now carries both commits, unchanged, with the combined measurement (17.46 → 8.35 µs, **2.09x**) and the standalone figures for each.

Nothing is dropped: the `pending` flag, the `sync_stream()` helper and the record-before-drain reorder are the same commit, cherry-picked as-is onto #1022's branch.

PR #1024, opened 2026-08-05
[profiler] Harvest device timestamps after the stream drains instead of blocking on every operation (profiler-on cost -34%)

mikepapadim:
## Summary

With `-Dtornado.profiler=True` (`--enableProfiler`), every transfer and every kernel launch did `resolveEvent → waitForEvents → getElapsedTime` **inline**, so the host waited for each operation to complete before it enqueued the next one. JFR on a profiler-on run put **51.9 % of wall time inside `CUDAEvent.clWaitForEvents`** (`cuEventSynchronize`).

Two consequences, both bad:

1. **Turning the profiler on made a short task graph 3.3× slower** — 17.46 µs → 57.32 µs per `execute()` for `saxpy` over 512 floats.
2. **The numbers it reported were wrong.** It gave `copyInAvg = 12.9 µs` for a 2 KB host-to-device copy that Nsight Systems measures at **0.65 µs** — the reported span was mostly the serialisation the profiler itself introduced.

A profiler that changes the thing it measures by that much is not usable for the workloads people most want to profile.

## Change

The reads are deferred to a point where the stream has already been drained and every event is complete, so reading a timestamp costs nothing. Two queues, one per layer:

- **`TornadoVMInterpreter`** (backend-neutral): `deferProfilerSample(eventIndex, Consumer<Event>)` replaces the five inline read sites — copy-in ×2, copy-out ×2, atomics copy-in. Harvested in the run epilogue, immediately after `flush()`.
- **`CUDADeviceContext`** (CUDA-only): `deferProfilerRead(Runnable)` for the three driver sites — `CUDAKernelScheduler.updateProfiler`, `CUDAInstalledCode.submitSequential`, `CUDAInstalledCode.updateProfilerKernelContextWrite`. Harvested in `flush()` and `sync()`.

Both queues harvest early once 256 samples are outstanding, so a long run cannot hold references to event slots that the backend event pool has since recycled.

Library-task timing is deliberately left inline: it measures host wall time bounded by stream markers, so it genuinely needs the synchronisation.

Nothing about what is recorded changes — the same metrics are accumulated into the same `ProfilerType` slots, just read later. Profiler results are consumed after execution, so the deferral is not observable.

## Numbers

RTX 4090, JDK 21.0.2, base `develop` @`ba28f1526`. `saxpy 100000 512`, per-`execute()` median, 5 runs.

| build | profiler OFF | profiler ON | ratio |
|---|---|---|---|
| `develop` | 17.46 µs | **57.32 µs** | 3.28× |
| this PR | 17.47 µs | **37.77 µs** (37.62, 37.72, 37.77, 37.83, 37.85) | 2.16× |

**Profiler-on cost down 34 %.** The default path is untouched, as expected — this PR only changes code behind `TornadoOptions.isProfilerEnabled()`.

JFR on the profiler-on run after the change: `clWaitForEvents` drops from **51.9 % → 1.1 %** of wall.

Stacked with the two dispatch-overhead PRs in this series the profiler-on figure reaches 32.61 µs.

## The reported metrics get closer to the truth

`saxpy 20000 512`, `-Dtornado.profiler=True -Dtornado.log.profiler=True`, against Nsight Systems as ground truth:

| metric | `develop` | this PR | nsys |
|---|---|---|---|
| `kernelMin` | 2336 ns | 2368 ns | kernel 1141 ns |
| `kernelAvg` | 10 663 ns | 9097 ns | — |
| `copyInAvg` | 12 899 ns | 4624 ns | H2D 646 ns |

`copyInAvg` moves about 3× closer to the real copy time. The residual inflation is queue wait inside the start→end event span, not host blocking.

## Remaining profiler overhead, for the record

Still ~2.2× versus profiler-off, and it is not the deferral. With timing enabled each operation needs `CU_EVENT_DEFAULT` events rather than `CU_EVENT_DISABLE_TIMING`: nsys measures `cuEventRecord` at **615 ns** with timing against **137 ns** without, six of them per iteration, and the timing events also stretch the terminal `cuStreamSynchronize` from 3.0 µs to 14.0 µs.

Halving that by reusing operation *N−1*'s end event as operation *N*'s start is **incompatible with deferred harvesting** — shared boundary events get re-recorded before anyone reads them. Cutting it further would mean an opt-in timing granularity (for example kernels only, leaving `copyIn`/`copyOut` untimed), which is a behavioural change and out of scope here.

## Testing

- `tornado-test -V --quickPass`: **1031 PASS / 9 FAILED**, identical to `develop` on the same machine. All failures whitelisted.
- Profiler-on runs to exercise the deferred harvest specifically: `TestArrays` 23/23, `TestReductionsIntegersKernelContext` 12/12.

Note that the `TornadoVMInterpreter` half is backend-neutral and therefore also affects OpenCL and Metal; the deferral is mechanical there (same reads, later), and the CUDA-specific queue is confined to `CUDADeviceContext`.

## Reproduction

```bash
# profiler on / off
tornado --jvm="-Dtornado.benchmarks.skipserial=True -Dtornado.profiler=True" \
  --params="saxpy 100000 512"

# JFR attribution
tornado --jvm="-XX:StartFlightRecording=filename=prof.jfr,dumponexit=true \
  -XX:FlightRecorderOptions=stackdepth=192 -XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints \
  -Dtornado.benchmarks.skipserial=True -Dtornado.profiler=True" \
  --params="saxpy 300000 512"
jfr print --events jdk.NativeMethodSample prof.jfr
```

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDADeviceContext.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDADeviceContext.java
@@ -203,6 +203,7 @@ public class CUDADeviceContext implements CUDADeviceContextInterface {
         // host against up to 7 queues per execution.
         joinRoleQueues(executionPlanId, commandQueue);
         commandQueue.finish();
+        harvestDeferredProfilerReads();
     }
 
     @Override
@@ -941,6 +942,39 @@ public class CUDADeviceContext implements CUDADeviceContextInterface {
     public void flush(long executionPlanId) {
         CUDACommandQueue commandQueue = getCommandQueue(executionPlanId);
         commandQueue.flush();
+        harvestDeferredProfilerReads();
+    }
+
+    /**
+     * Profiler read-outs postponed until the stream has been drained. Reading a device
+     * timestamp waits for its event, so doing it inline after each launch pins the host to
+     * the device for the whole run; the reads are batched here and run once the work is
+     * known complete, where they cost nothing.
+     */
+    private final List<Runnable> deferredProfilerReads = Collections.synchronizedList(new ArrayList<>());
+
+    /** Bound on outstanding reads, so a long run cannot outlive the backing event slots. */
+    private static final int MAX_DEFERRED_PROFILER_READS = 256;
+
+    public void deferProfilerRead(Runnable read) {
+        deferredProfilerReads.add(read);
+        if (deferredProfilerReads.size() >= MAX_DEFERRED_PROFILER_READS) {
+            harvestDeferredProfilerReads();
+        }
+    }
+
+    public void harvestDeferredProfilerReads() {
+        if (deferredProfilerReads.isEmpty()) {
+            return;
+        }
+        List<Runnable> pending;
+        synchronized (deferredProfilerReads) {
+            pending = new ArrayList<>(deferredProfilerReads);
+            deferredProfilerReads.clear();
+        }
+        for (Runnable read : pending) {
+            read.run();
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


mikepapadim:

|---|---|---|---|

mikepapadim:

PR #1025, opened 2026-08-05
[cuda] Propagate CUDA driver failures to Java instead of only logging them

mikepapadim:
## The bug

When a CUDA driver call fails, the JNI **prints the error and carries on**. `LOG_CUDA_AND_VALIDATE` is a logging macro, despite the name — it writes to stdout and returns:

```c
CUresult result = cuLaunchKernel(...);
LOG_CUDA_AND_VALIDATE("cuLaunchKernel", result);   // prints, does not throw
return end_event(ev, queue);                        // hands Java a normal event handle
```

Java never learns the launch failed. Execution continues, the read-back copies whatever was already in the output buffer, and the run reports success.

The Java side is already wired for this — `CUDACommandQueue.enqueueNDRangeKernel` and friends declare `throws CUDAException` and convert it into `TornadoBailoutRuntimeException` — but **nothing in the JNI has ever thrown `CUDAException`**, so the whole path was dead code.

## Reproduction

A `KernelContext` task with a 2D worker grid whose Y extent exceeds the hardware's `gridDim.y` limit of 65535, so the *launch* fails (the compile succeeds). The kernel writes `1.0f` into every element of an array initialised to `0.0f`:

```java
WorkerGrid worker = new WorkerGrid2D(size, 100_000_000);
worker.setLocalWork(64, 1, 1);
...
plan.withGridScheduler(grid).execute();
System.out.println("out[0] = " + out.get(0));   // kernel writes 1.0
```

Run with `-Dtornado.recover.bailout=False`:

**Before (develop @`ba28f1526`)**
```
[TornadoVM-CUDA-JNI] ERROR : cuLaunchKernel -> Returned: 1 (CUDA_ERROR_INVALID_VALUE)
RESULT: execute() returned normally
RESULT: out[0] = 0.0 (kernel writes 1.0)
```

**After**
```
[TornadoVM-CUDA-JNI] ERROR : cuLaunchKernel -> Returned: 1 (CUDA_ERROR_INVALID_VALUE)
RESULT: threw TornadoBailoutRuntimeException: Bailout from LAUNCH Bytecode:
Reason: cuLaunchKernel failed: CUDA_ERROR_INVALID_VALUE (1) - invalid argument
```

The failure mode is quiet and it produces plausible-looking numbers. I hit it while screening NVRTC compiler flags: a cubin built with `--Ofast-compile=max` loads but cannot launch (`CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES`), and `nbody 500 16384` cheerfully reported **`median(ns)=4.89e+04` against a real 1.72 ms** — a 35× "speed-up" produced entirely by no kernel running. Anyone benchmarking would have believed it.

## Change

A single helper in `cuda_jni.h`:

```c
static inline bool tornado_throw_cuda_exception(JNIEnv *env, const char *call, CUresult result);
```

It raises `uk.ac.manchester.tornado.drivers.cuda.exceptions.CUDAException` with the call name, the symbolic error name, the numeric code and the driver's own description, and reports whether an exception is pending on return — what actually happened, not what was attempted, since callers use that answer to decide whether to abandon their result. An exception already pending is left alone: it is the first failure and the useful one, and `ThrowNew` on top of one is undefined.

A second helper, `tornado_discard_event`, frees a boxed event whose handle is about to be dropped. A JNI method returning with an exception pending has its return value discarded by the JVM, so an event handed back on that path would never reach the Java side that releases it.

Applied to:

- `cuLaunchKernel` — the event is still recorded before throwing, so the queue's bookkeeping stays consistent
- `cuMemcpyHtoDAsync` / `cuMemcpyDtoHAsync`, and the `cuStreamSynchronize` that follows a blocking transfer
- `cuStreamSynchronize` in `clFlush` / `clFinish`

The transfer helpers now report failure through a `CUresult *failure` out-parameter rather than throwing in place. That matters: the Java-array transfer paths hold a `GetPrimitiveArrayCritical` region across the copy, and `ThrowNew` allocates, so the throw has to happen **after** `ReleasePrimitiveArrayCritical`. The off-heap paths hold no critical region and throw directly.

`LOG_CUDA_AND_VALIDATE` is left in place, so the diagnostic line still appears before the exception.

A later pass extended the same rule to the six remaining sites on the execution path that discard their `CUresult`: `record_event`/`begin_event`/`end_event` (a failed `cuEventCreate`/`cuEventRecord` still returned a boxed event whose handle was never created, so every wait list and profiler timestamp built on it referred to nothing), `clCreateCommandQueue` (a failed `cuCtxSetCurrent` produced a queue bound to whatever context was current), `clWaitForEvents` (a failed `cuEventSynchronize` let the caller read a buffer the device may still be writing), and `cuMemcpyDtoD` (a failed device-to-device copy left the destination stale). The rule the audit produced — one route per call: an exception where the result is discarded, a checked status where the signature already returns one, never both — lives in `tornado_report_cuda_failure`.

## Scope

Deliberately limited to the operations whose failure silently corrupts results and whose Java wrappers already declare `throws CUDAException`. Allocation, module loading and NVRTC compilation already surface their failures through other paths — the ptxas error for an over-sized shared-memory kernel, for instance, is already reported correctly as a `TornadoBailoutRuntimeException`.

## Testing

`TestLaunchFailureReporting.testRejectedLaunchIsReported` covers the reproduction above. The trigger is the `gridDim.y` cap rather than anything hardware-specific: 65535 blocks in Y is the limit on **every** CUDA compute capability, so the launch is rejected on any NVIDIA GPU whatever its shared memory, register file or driver version, while the kernel itself still compiles and loads.

That matters — an earlier version of this test provoked the failure with a 64 KB static `__shared__` array, and on the CI GPU (sm_120, NVRTC targeting compute_90) that launch is *accepted*, so the test failed because nothing had failed.

- **Negative control**: with `develop`'s JNI sources in place the test fails (`output stayed 0.0` — the buffer the rejected kernel never wrote); with this PR's JNI it passes. Same build, same machine, only the JNI swapped.
- `tornado-test --quickPass`: **0 failures**, 188 s.
- No measurable cost — the added work is one comparison against `CUDA_SUCCESS` per call. `saxpy 100000 512`: 17.459 / 17.511 / 17.451 µs against a `develop` median of 17.46 µs.

Tested on RTX 4090, CUDA driver via NVRTC 12.6, JDK 21.0.2.


mikepapadim:
Found while auditing NVRTC compiler flags against the generated CUDA-C (related to #1010). That audit also turned up three flags that are **byte-for-byte no-ops** on TornadoVM's generated code — verified by hashing the emitted cubin, not by timing:

| flag | cubin identical to baseline? |
|---|---|
| `--restrict` | yes — no-op |
| `--extra-device-vectorization` | yes — no-op |
| `--fmad=false` | yes — no-op |

The cause is visible in the codegen: every pointer is laundered through `unsigned long long` and parameters are `unsigned char *`, so no aliasing or vectorization information survives for the first two; and the generator already emits explicit `fma()` calls, so there is no `a*b+c` contraction left for `--fmad=false` to disable.

That bears on #1010: its `fast` profile pairs `--use_fast_math` (which does work — `dft` compile 27.6 → 19.9 ms, cubin 6776 → 5480 B) with `--extra-device-vectorization` (which does nothing), and its `repro` profile is `--fmad=false`, which cannot deliver reproducibility while the FMAs come from the code generator rather than from contraction.

It also answers the open question in #1010's description about the `-lineinfo` size delta: **compile time unchanged, cubin 2.2× on `nbody` (5352 → 11 560 B) and 2.4× on `dft` (6776 → 16 568 B)**.

One flag looks worth adding unconditionally: **`--minimal`** gives a byte-identical cubin with a 5 % faster compile on `nbody` and 10 % on `dft`, and is available from CUDA 11.5 so it clears the `build-cuda` CI floor. Happy to send that separately.

mikepapadim:

mikepapadim:
Folded in the test from #988, which covers the same ground with a case this PR was missing: `TestLaunchFailureReporting.testRejectedLaunchIsReported` — a `KernelContext` task whose 2D worker grid exceeds the hardware `gridDim.y` limit, so the *launch* fails while the compile succeeds, and the failure must arrive in Java as a bailout rather than as a normal return with untouched output.

Ported unchanged apart from dropping its `assertNotBackend(TornadoVMBackendType.SPIRV)` line — that backend no longer exists on `develop`, and the file did not compile with it. Passing on CUDA; quickPass **85 failures, equal to the baseline**; checkstyle clean.

#988 is now redundant and I am closing it.

mikepapadim:
Pushed `0fd0ed88a`: the rest of the audit.

I classified all **60** `LOG_CUDA_AND_VALIDATE` sites in the CUDA JNI by what happens to the `CUresult`: **28** return it to Java as a status the Java side already checks (the CUDA-graph family, `createBuffer`, `cuMemHostRegister`), **7** are cleanup paths and **13** are device enumeration — both deliberately log-only — and **12** discard it on the execution path. Six of those were this PR already; the remaining six are now closed:

- `record_event`/`begin_event`/`end_event` — a failed `cuEventCreate`/`cuEventRecord` still returned a boxed event whose handle was never created, so every wait list and profiler timestamp built on it referred to nothing;
- `clCreateCommandQueue` — a failed `cuCtxSetCurrent` produced a queue bound to whatever context was current;
- `clWaitForEvents` — a failed `cuEventSynchronize` let the caller read a buffer the device may still be writing (reported *after* the `GetPrimitiveArrayCritical` region, since `ThrowNew` allocates);
- `cuMemcpyDtoD` — a failed device-to-device copy left the destination stale.

New `tornado_report_cuda_failure` helper carries the rule the audit produced: one route per call — exception where the result is discarded, checked status where the signature already returns one, never both.

Verified: `TestLaunchFailureReporting` 1/1, `TestCuBlas` 12/12, `TestCuDnn` 9/9, `TestExecutor` 7/7, quickPass 85 = baseline, `saxpy 512` unchanged at 17.48 µs, checkstyle clean.

mikepapadim:
Pushed `fcc32afe9` + `e8239f996`, and merged the current `develop`. Two things: the CI failure, and the review comments.

### The CI failure

`TestLaunchFailureReporting#testRejectedLaunchIsReported` failed on both CUDA test legs (`jdk25`, `jdk27`) with *"a launch the driver rejects must not be reported as success (output stayed 1.0)"*. The test was not testing what it claimed on that machine: it provoked the failure with a 64 KB static `__shared__` array, and on the CI GPU (sm_120, NVRTC targeting compute_90 and the driver JIT-ing) that launch is **accepted**. The kernel ran, the output was correct, and the test failed because nothing had failed. The `jdk21` leg failed separately and unrelatedly — a truncated `sdk.tgz` artifact (`gzip: stdin: unexpected end of file`), infrastructure only.

The trigger is now the one the PR description has always described: a 2D worker grid whose Y extent needs more than 65535 blocks. `gridDim.y` is capped at 65535 on **every** CUDA compute capability, so the launch is rejected with `CUDA_ERROR_INVALID_VALUE` on any NVIDIA GPU regardless of shared memory, register file, or driver version — while the kernel itself still compiles and loads, which is the path under test.

Verified as a negative control rather than assumed: with `develop`'s JNI sources in place the test **fails** (`output stayed 0.0` — the buffer the rejected kernel never wrote), and with this PR's JNI it **passes**. Same build, same machine, only the JNI swapped.

### Review comments

**`tornado_throw_cuda_exception` returned `true` without having thrown.** Callers use the answer to decide whether to abandon their result, so a failed `FindClass`/`ThrowNew` could make a JNI method return an empty handle with no exception to explain it. It now returns `env->ExceptionCheck()` — what actually happened, not what was attempted — and returns early when an exception is already pending rather than calling `ThrowNew` on top of one, which is undefined. That second half also closes a real path: `begin_event` can throw and let the operation continue to `end_event`, which would then have thrown again.

**Events leaked on the exception paths.** A JNI method returning with an exception pending has its return value discarded by the JVM, so the `cuda_event_t` handed back there never reaches the Java side that frees it. New `tornado_discard_event` helper; applied to `cuLaunchKernel` and all four transfer paths (Java-array and off-heap, read and write). `record_event`/`end_event` now route their own cleanup through it too, which fixes a pre-existing leak in the same lines: they did `delete ev` without `cuEventDestroy(ev->start)`, so a failure after `begin_event` had recorded a timing event dropped that `CUevent`.

### Verified

- `TestLaunchFailureReporting` 1/1 on RTX 4090, and 0/1 against unmodified `develop` JNI (negative control above).
- `tornado-test --quickPass`: **0 failures**, 188 s.

stratika (approved):

PR #1027, opened 2026-08-06
[runtime] Clear only the written prefix of each wait-list row between executions (1.57x on dependency-tracked decode)

mikepapadim:
## Problem

`TornadoVMInterpreter.initWaitEventList()` runs at the top of *every* `execute()` and clears each allocated wait-list row in full. A row is `new int[MAX_EVENTS]`, and `tornado.max.events` defaults to **32768** (128 KB per row), so the clear is 32768 stores per dependency slot per execution regardless of how many entries were actually used — typically a handful.

The full-row clear was deliberate: `resetEventIndexes()` rewinds an event list in the middle of an execution, so live entries can sit beyond the current `eventsIndexes` value and a clear bounded by that index would leave stale event ids behind.

This dominates host time for workloads that run many short executions per iteration. Profiled with JFR on a Qwen1.5-MoE decode loop (GPULlama3.java, CUDA backend, RTX 4090), `Arrays.fill` under `initWaitEventList` was **51% of decode host CPU** when dependency tracking is on.

## Fix

Track a per-row high-water mark (`eventsWrittenPrefix`) of every write since the last clear, and clear only that prefix. Everything past the mark is still `-1` from allocation, and the mark — unlike `eventsIndexes` — is never rewound by `resetEventIndexes()`, so the stale-entry hazard the original comment describes cannot occur.

Rows that are never written stay `null` and cost nothing, as before.

## Measurements

Qwen1.5-MoE-A2.7B-Chat Q8_0, CUDA backend, RTX 4090, 128 generated tokens, 2-3 runs per configuration:

| Configuration | Before | After |
|---|---|---|
| default (`tornado.vm.deps=False`) | 72.4 / 71.5 tok/s | 70.9 / 72.2 tok/s |
| `-Dtornado.vm.deps=True` | 46.4 / 45.9 tok/s | **72.2 / 72.3 / 71.9 tok/s (1.57x)** |
| `-Dtornado.vm.deps=True -Dtornado.max.events=1024` | 72.1 / 72.4 tok/s | 71.9 tok/s |

The third row is the workaround this patch makes unnecessary: shrinking `tornado.max.events` used to be the only way to get the dependency-tracking path back to parity. The default path is unaffected because rows are allocated lazily, so nothing is cleared when dependency tracking is off.

## Testing

`tornado-test --quickPass` on the CUDA backend, JDK 21:

- before: 85 failures
- after: 85 failures
- after, with `-Dtornado.vm.deps=True`: 85 failures

Identical per-class results (all pre-existing failures: Metal-only simdgroup tests, multi-backend virtualization tests, half-float vector types). `TestArrays`, `TestMultipleTasksSingleDevice` and `TestBatches` also pass with `-Dtornado.vm.deps=True`.

## Backend/OS tested


mikepapadim:

mairooni (approved):
LGTM

Issue #1028, opened 2026-08-06
[perf] Host-side runtime overhead: where the time actually goes, the fixes, and what is left


mikepapadim:
## Context

For small and medium task graphs, and for anything that runs many short executions per iteration (LLM decode, filters, iterative solvers), TornadoVM's wall-clock is dominated by **host-side per-operation cost**, not by GPU work. This issue collects the profiling campaign that quantified that cost, the PRs that remove it, and what is left.

A result worth stating up front, because it redirects where optimisation effort belongs:

> The Java bytecode interpreter and its data structures account for **~0.9% of wall time** on a dispatch-bound graph (118 `jdk.ExecutionSample` vs 13,545 `jdk.NativeMethodSample`, and most of the Java samples are the benchmark harness's own `Arrays.sort`). The cost is in the **per-operation CUDA dispatch sequence and runtime bookkeeping**, not in interpreting bytecodes.

## Method

- **JFR** — 1 ms sampling, `stackdepth=192`, `-XX:+DebugNonSafepoints`; samples phase-split by filtering stacks (decode vs start-up).
- **Nsight Systems** — `nsys profile -t cuda --sample=none --cpuctxsw=none`, then `nsys stats --report cuda_api_sum` for per-iteration API call counts, used as ground truth for what the driver actually did.
- **Benchmarks** — `BenchmarkRunner saxpy 100000 512` (dispatch-bound: one kernel, one H2D, one blocking D2H, 2 KB moved, 1.1 µs kernel) for per-`execute()` medians; `saxpy 2000 16777216` and `nbody 500 16384` as GPU-bound controls; GPULlama3.java on Qwen1.5-MoE-A2.7B-Chat Q8_0 as the real end-to-end workload.
- **Hardware** — NVIDIA RTX 4090, Ubuntu 24.04, JDK 21.0.2, CUDA backend.

## What the profiles found

On `develop`, a `saxpy` over 512 floats spent **17.46 µs per `execute()` while the GPU did 3.31 µs of work**. Per iteration the runtime issued **4 stream operations, 12 event calls and 3 `cuStreamSynchronize`** to move 2 KB and run a 1.1 µs kernel. Four separate causes, each with its own PR:

| # | Finding | PR |
|---|---|---|
[...]

PR #1029, opened 2026-08-06
[runtime] Read non-terminal copy-outs asynchronously instead of blocking on each one (1.27x on an 8-output graph)

mikepapadim:
## The bug

`TornadoVMGraphCompiler` patches only the **last** copy-out of a task graph into `TRANSFER_DEVICE_TO_HOST_ALWAYS_BLOCKING`:

```java
if (code[position] == TornadoVMBytecodes.TRANSFER_DEVICE_TO_HOST_ALWAYS.value()) {
    code[position] = TornadoVMBytecodes.TRANSFER_DEVICE_TO_HOST_ALWAYS_BLOCKING.value();
}
```

The point of that patch is to add a wait that the other copy-outs do not have. Except they do: the handler for the *non*-blocking bytecode calls the same blocking device method as the blocking one.

```java
// TornadoVMInterpreter.transferDeviceToHost - the TRANSFER_DEVICE_TO_HOST_ALWAYS handler
int readEvent = interpreterDevice.streamOutBlocking(...);
```

`transferDeviceToHost` and `transferDeviceToHostBlocking` differ only in logging and in `resetEventIndexes`. So a task graph with N outputs performs **N blocking reads**, each one a full host round trip to the device, when only the last one needs to block.

## Change

`transferDeviceToHost` now takes `streamOut` (the asynchronous read that already exists on every backend) whenever the blocking variant is not doing something extra. `streamOutBlocking` additionally handles atomics, under-demand partial copies and batch chunks — `enqueueRead` supports none of those — so the async path is guarded:

```java
final boolean canReadAsync = !objectState.isAtomicRegionPresent() //
        && objectState.getPartialCopySize() == 0 //
        && sizeBatch <= 0;
```

Outputs are still visible when `execute()` returns: the terminal `..._BLOCKING` read and the sync in `TornadoTaskGraph.waitOn()` both drain the queue, and the reads are issued in order behind the work that produces them. Nothing about the visibility contract changes — only how many times the host stops to wait for it.

While here: `OCLTornadoDevice.streamOut` and `MetalTornadoDevice.streamOut` passed `events == null` as their `useDeps` argument, i.e. inverted — with a dependency list present they would drop it and return `-1` instead of the read event. The CUDA implementation has it the right way round (`events != null`). That path was unreachable before this patch, since nothing called `streamOut` for a copy-out; it is reachable now, so it is fixed here.

## Numbers

RTX 4090, JDK 21, CUDA backend, `develop` @`4d90e2e9f`. A task graph with one input and N outputs of 4096 floats each, 20,000 timed executions after 2,000 warm-up, per-`execute()` median:

| outputs | `develop` | this PR | speed-up |
|---|---|---|---|
| 1 | 19.43 µs | 19.43 µs | 1.00x |
| 2 | 35.80 µs | 25.91 µs | **1.38x** |
| 4 | 53.29 µs | 43.86 µs | 1.22x |
| 8 | 92.35 µs | 72.89 µs | **1.27x** |

One output is unchanged, as expected — that copy-out is the terminal one and stays blocking. Every additional output saves roughly 2.4 µs, which is the host round trip that was being paid per copy.

The probe checks the output values after the loop; `mismatches=0` in every configuration.

## Testing

- `tornado-test --quickPass`, CUDA backend: **85 failures, identical to the `develop` baseline on this machine** (all pre-existing: Metal-only simdgroup tests, multi-backend virtualization tests, half-float vector types).
- The OpenCL and Metal `streamOut` fixes are compile-checked but not run here (no OpenCL/Metal device on this machine); they restore the documented contract of the method rather than changing behaviour of an exercised path.

## Backend/OS tested

Part of #1028.


mikepapadim:
Added `TestMultipleCopyOuts` (registered in `tornado-test`), 5 tests, all passing on the CUDA backend:

| test | what it covers |
|---|---|
| `testFourOutputsInOneGraph` | four outputs, so three non-terminal copy-outs are read asynchronously and all four must be complete when `execute()` returns |
| `testMultipleOutputsAcrossExecutions` | eight executions with a different input each time — a read that has not landed shows up as the *previous* execution's output, which a single-shot test cannot catch |
| `testMixedOutputTypes` | `FloatArray` + `IntArray` outputs in one graph, so the reads go through different buffer wrappers |
| `testPartialCopyOutWithSecondOutput` | `UNDER_DEMAND` partial copy via `DataRange` next to an eager output — covers the `partialCopySize != 0` guard that keeps the blocking read |
| `testBatchedCopyOut` | `withBatch("1MB")` over 3 executions — covers the `sizeBatch > 0` guard |

**Pre-existing bug found while writing these**: a *batched* graph with **two** outputs fails on `develop` with

```
Cannot invoke "uk.ac.manchester.tornado.api.memory.XPUBuffer.toBuffer()" because the return value of "uk.ac.manchester.tornado.runtime.common.XPUDeviceBufferState.getXPUBuffer()" is null
```

It reproduces with the interpreter change reverted, so it is unrelated to this PR and I left the batched test single-output rather than encoding the failure. I can open a separate issue for it if useful.

mikepapadim:
Ran this on **OpenCL** as well as CUDA, and it found a real hole in the first version of the patch — now fixed in `2c6997765`.

### What OpenCL exposed

A task graph with two **`Matrix2DFloat`** outputs returned **all zeros** for the non-terminal one. It passes on CUDA and passes on OpenCL with the interpreter change reverted, so it was this PR.

Cause: `XPUBuffer.enqueueRead` is not equivalent to `XPUBuffer.read` for every buffer implementation. On OpenCL a `Matrix2DFloat` is not in the `instanceof` chain of `OCLTornadoDevice.createDeviceBuffer`, so it falls through to `OCLFieldBuffer` — whose asynchronous read copies something else entirely. The blocking read was hiding that.

### Fix

An explicit capability instead of an assumption:

```java
// XPUBuffer
default boolean supportsAsyncRead() {
    return false;
}
```

Opted in by the memory-segment wrappers on all three backends (CUDA, OpenCL, Metal), which do implement the two reads over the same region. Everything else — field buffers, vector wrappers, on-heap array wrappers — keeps the blocking read, as before this PR. The interpreter guard now reads:

```java
final boolean canReadAsync = objectState.getXPUBuffer().supportsAsyncRead()
        && !objectState.isAtomicRegionPresent()
        && objectState.getPartialCopySize() == 0
        && sizeBatch <= 0;
```

Conservative on purpose: opting a wrapper in is a one-line change once its `enqueueRead` is verified against its `read`.

### Re-verified

- **CUDA**: `TestMultipleCopyOuts` 6/6 (added `testTwoMatrixOutputs`); multi-output timings unchanged — 1 output 19.41 µs, 4 outputs 44.02 µs, 8 outputs 68.63 µs vs `develop`'s 19.43 / 53.29 / 92.35 µs. `tornado-test --quickPass` 86 vs the 85 baseline, the one difference being the known-flaky `MMwithBytes` (4/4 in three isolated re-runs).
- **OpenCL** (NVIDIA + Intel platforms on this box): `TestMultipleCopyOuts` 6/6, including the matrix case that failed before the fix. Aggregate quickPass is not a usable signal on this machine — `develop` itself fails 271 there, e.g. `TestArrays` 4/23 — so I compared per class on the same device instead.
- **Metal**: not runnable here (Linux). The capability defaults to off, so nothing changes for any wrapper that has not opted in.

### Related, not fixed here

`OCLTornadoDevice.streamOutBlocking` and `MetalTornadoDevice.streamOutBlocking` pass `events == null` as their `useDeps` argument — inverted, where CUDA passes `events != null`. That is the *blocking* path, so with `tornado.vm.deps=True` or intra-plan concurrency those two backends drop the wait list on every terminal copy-out. Same bug class as the `streamOut` one this PR fixes, but on a live path; I have a one-line change for it and did not include it here because I cannot validate the deps path properly on this machine's OpenCL. Say the word and I will send it separately.

mairooni (approved):
LGTM. Validated this also against `GPULlama3.java` (branch: [tornadovm-jdk27](https://github.com/beehive-lab/GPULlama3.java/tree/feat/tornadovm-jdk27))

PR #1036, opened 2026-08-07
[api] TornadoExecutionPlan.transferToDevice(): ad-hoc copy-in without running a kernel to force it

mikepapadim:
## What this adds

An explicit copy-in, so that getting data onto the device at a chosen moment stops being something you have to fake by running a kernel you do not want.

```java
plan.transferToDevice();              // every input of the plan, no task runs
plan.transferToDevice(weights);       // just this object, now
plan.transferToDevice(a, b, c);       // several
```

The copy-in counterpart of `TornadoExecutionResult.transferToHost(...)`. The data is on the device when the call returns.

- **No arguments** — walks the plan's bytecode for its data transfers only: `ALLOC` and the copy-ins run; `LAUNCH`, the copy-outs, the barrier and the CUDA-graph operations are skipped. (`DEALLOC` deliberately still runs — it is a no-op for the locked buffers holding the uploaded data, and skipping it lets every task-graph of a plan hold its buffers at once, which is enough to exhaust the device on a plan of many graphs.)
- **With arguments** — uploads exactly those objects, through whichever task-graphs of the plan take them as parameters; graphs that do not know an object ignore it. An object with no device buffer yet gets one, so it works before the plan has ever run, and it works for `FIRST_EXECUTION` objects that would otherwise never be uploaded again.

## Why

There is no way today to put data on the device without running something. Applications therefore run work they do not need, purely for the side effect. GPULlama3.java's `forceCopyInReadOnlyData()` runs the activation graph, then **every layer graph**, then the logits graph, on zeroed state, so that the model weights get copied in:

```java
public void forceCopyInReadOnlyData() {
    state.wrapX.clear();
    state.positionHolder.init(0);
    executionPlan.withGraph(taskGraphLayout.activationIdx())...execute();
    for (int layer = 0; layer < config.numberOfLayers(); layer++) {
        executionPlan.withGraph(taskGraphLayout.layerIdx(layer))...execute();
    }
    executionPlan.withGraph(taskGraphLayout.logitsIdx())...execute();
}
```

With this PR that whole method is `executionPlan.transferToDevice();`, and refreshing one buffer mid-run — new weights, a new embedding row, a table swapped between executions — is `executionPlan.transferToDevice(thatObject)` instead of a task graph whose kernel exists only to make the transfer happen.

## Numbers

RTX 4090, JDK 21, CUDA backend. 64 MB of `FIRST_EXECUTION` weights plus a 64 MB `EVERY_EXECUTION` input, after `withPreCompilation()`:

| | `transferToDevice()` | `transferToDevice(weights)` | first `execute()` | second `execute()` |
|---|---|---|---|---|
| no up-front upload | — | 17.34 ms (allocates + uploads) | 63.95 ms | 7.63 ms |
| whole-plan upload first | 34.51 ms | **3.71 ms** (buffer exists: pure H2D) | 52.86 ms | 7.63 ms |
| without either | — | — | 78.95 ms | 7.50 ms |

The first execution loses exactly the upload it no longer has to do, and a targeted refresh of an already-allocated 64 MB buffer costs 3.71 ms — the copy and nothing else.

**GPULlama3.java, Qwen1.5-MoE-A2.7B-Chat Q8_0** — I actually made the replacement (`forceCopyInReadOnlyData()` becomes `executionPlan.transferToDevice()` when CUDA graphs are off; with them on the pass is also the capture, so it stays) and measured both paths in the same build, same 16 GB device heap, two runs each:

| | copy-in phase | decode | wall |
|---|---|---|---|
| dummy forward pass | 3673 / 3698 ms | 83.53 / 83.20 tok/s | 8.52 / 8.58 s |
| `transferToDevice()` | **3529 / 3504 ms** | 76.38 / 76.25 tok/s | 8.56 / 8.55 s |

Read that carefully, because it is not a win yet:

- the copy-in phase is **~4.6% faster** — the phase is dominated by the upload itself (`cuMemHostRegister` plus PCIe, 3.4 s of it in an nsys capture), so skipping the dummy kernels was never going to buy much;
- **decode looks 8% slower**, and that is the same work moving rather than appearing: the dummy pass installs and first-launches every kernel, which the API path leaves for the first real token. The gap shrinks from 8% at 128 tokens to ~3.7% at ~450 tokens, which is what a fixed ~150 ms one-off looks like;
- **wall-clock is identical**, and the plan needs ~2 GB more device heap this way (fails at 14 GB, fine at 16 GB) because nothing releases between graphs when no execution drives them.

So for that application today it is a wash in time and a clear win in intent: a method that ran a full forward pass for a side effect becomes one line that says what it means. The remaining piece — a warm-up that installs code without transferring, so neither path pays first-launch inside the timed loop — is worth doing separately, and is easier to see now that the transfer exists as its own operation.

The microbenchmark above is the honest picture of the mechanism; this is the honest picture of its worth on a start-up bound by something else.

## The `withGraph(i)` interaction, fixed here

`TornadoExecutor.selectGraph(i)` replaces the executor's graph list with the selected graph (keeping the full list in `subgraphList`), so a plan driven graph-by-graph — which is exactly how a pipeline runs — would have had its upload applied only to whichever graph was selected last. Both forms of `transferToDevice` now iterate every task-graph of the plan regardless of the current selection: an upload is about the plan's data, not about which graph runs next. Caught by the pipeline tests below, which fail without it.

## Where the transfers-only mode lives

Rebased on the interpreter dispatch refactor now on `develop`, which replaced the 24-arm `if`/`else` chain in `execute()` with a `switch` over the opcode enum and one handler method per bytecode. Its own comment notes that in the old shape "adding an execution mode meant editing every one of them" — so the transfers-only mode is applied at the dispatch point rather than spread across the handlers: an `EnumSet` of the bytecodes such a pass walks past (`LAUNCH`, both device-to-host transfers, `BARRIER`, the four CUDA-graph operations), checked once before the switch. Their operands are still consumed via `skipBytecodeOperands` — the reads are positional — but nothing is issued.

That routing exposed a latent gap in `skipBytecodeOperands`: it had **no branch for `CUDA_GRAPH_END_CAPTURE`**, because both of its existing callers (`preCompileLaunchesInCaptureRegion`, `skipToAfterEndCapture`) intercept that opcode before ever calling it. Sending it there left the 4-byte `graphId` in the stream and decoded everything after it from the wrong offset:

```
[REASON] [ERROR] TornadoVM Bytecode not recognized
```

Fixed by making the helper total for every opcode reachable in the loop (`INIT` is consumed in the bytecode header, before the loop starts). The five CUDA-graph tests listed below cover it — four of them fail without the fix.

## Testing

`TestOnDemandCopyIn`, **13 tests**, registered in `tornado-test`:

- upload-then-execute matches execute alone; the call runs no task (the output is untouched until the execution);
- a `FIRST_EXECUTION` input really is uploaded by the call — changing the host array afterwards does not change the result, while an `EVERY_EXECUTION` input still refreshes;
- targeted upload of one object **updates a `FIRST_EXECUTION` object that the runtime would never re-upload**, both before the first execution and between executions;
- several objects in one call, with objects the plan does not know ignored;
- a targeted upload reaches every task-graph of the plan that takes the object;
- applies to a plan of two task-graphs; repeated calls are harmless; composes with `withPreCompilation()`;
- **under CUDA graphs**: a transfers-only pass over a plan that captures and replays, the same with `KernelContext` kernels driven by a `GridScheduler`, and a targeted upload into an already-captured graph whose `FIRST_EXECUTION` weights nothing else would refresh.

`TestCopyInPipelines`, **7 tests** — the shapes this API exists for:

- update the input of a producer that persists its result on the device, and see the change through the consumer;
- overwrite a device-resident persisted buffer from the host mid-run (the case an application fakes with a task whose only purpose is the transfer);
- an LLM-shaped pipeline — a per-step graph, four chained layer graphs passing state through `persistOnDevice`/`consumeFromDevice`, a read-out graph — with the layer weights placed by `transferToDevice()` instead of a warm-up run, then one layer's weights swapped mid-run;
- the same pipeline under `withIntraPlanConcurrency()` (multi-stream);
- an upload between two executions of the same graph, 16 times, each one visible in that iteration's result;
- **persist/consume under CUDA graphs**, in both plain and `KernelContext` form: the weights placed by `transferToDevice()` rather than a warm-up run, so the transfers-only pass runs over bytecode containing the capture and launch operations, and swapped again once the graphs are captured.

Hybrid-API (library task) coverage, CUDA-only and skipped elsewhere by the existing guards:

- `TestCuBlas.testSgemmWithAdHocWeightUpload` — a `FIRST_EXECUTION` GEMM operand placed by an upload and then swapped between executions (14/14 in the class);
- `TestCuBlas.testSgemmPipelineWithUpFrontUpload` — `cublasSgemm` result persisted on the device and consumed by a second graph, plan uploaded up front;
- `TestCuDnn.testReluWithAdHocInputUpload` and `TestCuDnn.testReluThenTanhPipelineWithUpFrontUpload` — the same two shapes through cuDNN (11/11 in the class).

The five CUDA-graph tests are guarded with `assertNotBackend(OPENCL)`, so on OpenCL they report `UNSUPPORTED` rather than passing silently.

### Full suite vs. a same-build `develop` baseline

`tornado-test` (all 170 classes, 175 class-runs), RTX 4090, JDK 21, run on this branch and on `develop` built the same way:

| | this branch | `develop` baseline |
|---|---|---|
| **CUDA** | 11 failures / 6 classes | **11 / the same 6** |
| **OpenCL** | 8 failures / 6 classes | **8 / the same 6** |

The failing classes are identical on both sides and none touch the copy-in paths: `TestReductionsFloats`, `TestMatrixMultiplicationKernelContext`, `TestProfiler` (CUDA only), `ComputeTests`, `TransformerKernelsTest`, `TestDevices`, plus `TestMultipleTasksMultipleDevices` on OpenCL. `TransformerKernelsTest` is flaky rather than deterministic — it passed once and failed three consecutive re-runs on `develop`, same method and same float tolerance miss (`testReductionOneBlockWithLayer`, expected 0.8730353, got 0.9308163).

## Backends/OS tested

- OpenCL backend, same machine.
- Metal not runnable here (Linux); the change is backend-neutral — it drives the existing interpreter, `TornadoXPUDevice.streamIn` and `sync`.

Part of #1028.

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestCopyInPipelines.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestCopyInPipelines.java
@@ -74,6 +74,16 @@ public class TestCopyInPipelines extends TornadoTestBase {
         }
     }
 
+    /**
+     * CUDA graphs are a CUDA-backend feature. On every other backend these tests report the
+     * configuration as unsupported rather than failing - a JUnit Assume would be counted as a PASS
+     * by the test runner, so the typed *NotSupported exceptions are what the guard has to raise.
+     */
+    private void requireCudaGraphs() {
+        assertNotBackend(TornadoVMBackendType.OPENCL);
+        assertNotBackend(TornadoVMBackendType.METAL);
+    }
+
     public static void copyWithContext(KernelContext context, FloatArray input, FloatArray output) {
         int index = context.globalIdx;
         output.set(index, input.get(index));
@@ -327,7 +337,7 @@ public class TestCopyInPipelines extends TornadoTestBase {
      */
     @Test
     public void testPersistConsumePipelineUnderCudaGraphs() throws TornadoExecutionPlanException {
-        assertNotBackend(TornadoVMBackendType.OPENCL);
+        requireCudaGraphs();
 
         FloatArray weights = new FloatArray(SIZE);
         weights.init(2.0f);
@@ -382,7 +392,7 @@ public class TestCopyInPipelines extends TornadoTestBase {
      */
     @Test
     public void testPersistConsumePipelineUnderCudaGraphsWithKernelContext() throws TornadoExecutionPlanException {
-        assertNotBackend(TornadoVMBackendType.OPENCL);
+        requireCudaGraphs();
 
         FloatArray weights = new FloatArray(SIZE);
         weights.init(2.0f);
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestCopyInPipelines.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestCopyInPipelines.java
@@ -1,0 +1,305 @@
+/*
+ * Copyright (c) 2026, APT Group, Department of Computer Science,
+ * The University of Manchester.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package uk.ac.manchester.tornado.unittests.api;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+import uk.ac.manchester.tornado.api.ImmutableTaskGraph;
+import uk.ac.manchester.tornado.api.TaskGraph;
+import uk.ac.manchester.tornado.api.TornadoExecutionPlan;
+import uk.ac.manchester.tornado.api.annotations.Parallel;
+import uk.ac.manchester.tornado.api.enums.DataTransferMode;
+import uk.ac.manchester.tornado.api.exceptions.TornadoExecutionPlanException;
+import uk.ac.manchester.tornado.api.types.arrays.FloatArray;
+import uk.ac.manchester.tornado.unittests.common.TornadoTestBase;
+
+/**
+ * {@link TornadoExecutionPlan#transferToDevice(Object...)} in the pipelines it exists for: chains of
+ * task-graphs that hand data to each other through {@code persistOnDevice} and
+ * {@code consumeFromDevice}, where the data a stage needs has to be put on the device at a chosen
+ * moment rather than by running something.
+ *
+ * <p>The pipeline shape here is the one an LLM decode loop uses: a small per-step graph, a run of
+ * layer graphs chained on the device, and a final graph that copies one result back.
+ *
+ * <p>
+ * How to run?
+ * </p>
+ * <code>
+ * tornado-test -V uk.ac.manchester.tornado.unittests.api.TestCopyInPipelines
+ * </code>
+ */
+public class TestCopyInPipelines extends TornadoTestBase {
+
+    private static final int SIZE = 4096;
+    private static final int LAYERS = 4;
+
+    public static void scale(FloatArray input, FloatArray output, float alpha) {
+        for (@Parallel int i = 0; i < input.getSize(); i++) {
+            output.set(i, alpha * input.get(i));
+        }
+    }
+
+    public static void scaleInPlace(FloatArray state, FloatArray weights) {
+        for (@Parallel int i = 0; i < state.getSize(); i++) {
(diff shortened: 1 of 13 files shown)


tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/api/TestCopyInPipelines.java
```diff
@@ -0,0 +1,305 @@
...
+            // New weights, uploaded explicitly: the device copy of a FIRST_EXECUTION object would
+            // otherwise keep the values it was given the first time.
+            weights.init(3.0f);
+            plan.transferToDevice(weights);
+
+            plan.withGraph(0).execute();
+            plan.withGraph(1).execute();
+            for (int i = 0; i < SIZE; i++) {
+                assertEquals(6.0f, result.get(i), 0.001f);
+            }
```
mikepapadim, line 137: this is the pattern we didnt support so far and we need @mairooni @stratika @orionpapadakis

PR #1037, opened 2026-08-07
[api] Apply on-demand transfers to every task-graph of the plan, not just the selected one

mikepapadim:
## The bug

`TornadoExecutionResult.transferToHost(...)` silently does nothing — and the caller reads whatever the host array held before — when the plan is being driven one graph at a time and the requested object belongs to a graph that is not the selected one.

```java
plan.withGraph(0).execute();
TornadoExecutionResult result = plan.withGraph(1).execute();
result.transferToHost(outputOfGraph0, outputOfGraph1);

// outputOfGraph0[0] == 0.0   <- never copied back, no error
// outputOfGraph1[0] == 10.0
```

`TornadoExecutor.selectGraph(i)` replaces `immutableTaskGraphList` with the selected graph (keeping the full list in `subgraphList`), and `transferToHost` iterates `immutableTaskGraphList`. So the graph that owns the object is never asked.

That combination is not exotic: driving a plan graph-by-graph is how a pipeline runs — each stage selected in turn — and `UNDER_DEMAND` outputs exist precisely so results can be collected afterwards, at which point the "current" graph is whichever ran last.

## Fix

A transfer names an object, not a graph, so it applies to every task-graph of the plan:

```java
private List<ImmutableTaskGraph> allTaskGraphs() {
    return subgraphList != null ? subgraphList : immutableTaskGraphList;
}
```

used by `transferToHost(Object...)` and `partialTransferToHost(DataRange)`. Graphs that do not know an object already skip it (`argumentsLookUp`), so widening the loop cannot transfer anything extra — it only stops the owning graph from being skipped.

Deliberately limited to the two on-demand transfer paths, which are the ones whose failure mode is a wrong number rather than an error. `freeDeviceMemory`, `resetDevice` and friends keep their current behaviour.

## Testing

New `TestOnDemandTransferSelection` (3 tests, registered in `tornado-test`), passing on **CUDA and OpenCL**:

- an object owned by an unselected graph is copied back (fails on `develop`: reads 0.0);
- the same through a partial `DataRange` transfer in two halves;
- asking the selected graph for its own object still works.

`tornado-test --quickPass` on CUDA: 86 vs the 85-failure baseline, the difference being the known-flaky `MMwithBytes` (4/4 in isolated re-runs). `./mvnw checkstyle:check` clean.

## Backends/OS tested

- OpenCL backend, same machine.
- Metal not runnable here (Linux); the change is in the backend-neutral API layer.

Found while making the same rule hold for the copy-in direction in #1036. Part of #1028.

### DIFF
--- a/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutor.java
+++ b/tornado-api/src/main/java/uk/ac/manchester/tornado/api/TornadoExecutor.java
@@ -147,13 +147,23 @@ class TornadoExecutor {
     }
 
     void transferToHost(Object... objects) {
-        immutableTaskGraphList.forEach(immutableTaskGraph -> immutableTaskGraph.transferToHost(objects));
+        allTaskGraphs().forEach(immutableTaskGraph -> immutableTaskGraph.transferToHost(objects));
     }
 
     void partialTransferToHost(DataRange dataRange) {
         // At this point we compute the offsets and the total size in bytes.
         dataRange.materialize();
-        immutableTaskGraphList.forEach(immutableTaskGraph -> immutableTaskGraph.transferToHost(dataRange.getArray(), dataRange.getOffset(), dataRange.getPartialSize()));
+        allTaskGraphs().forEach(immutableTaskGraph -> immutableTaskGraph.transferToHost(dataRange.getArray(), dataRange.getOffset(), dataRange.getPartialSize()));
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


mikepapadim:
Folded into #1032 and closing this one.

Same entity: what the copy-out paths are supposed to do. #1032 now carries the contract tests, the OpenCL/Metal `useDeps` inversion and this graph-selection fix, with `TestOnDemandTransferSelection` intact — the three tests still fail on `develop` (they read 0.0) and pass on CUDA and OpenCL with the fix. Cherry-picked unchanged.

PR #1040, opened 2026-08-07
[runtime] Dispatch interpreter bytecodes through a switch over named handlers

mikepapadim:
Refactor only — no behaviour change, no new tests. `tornado-test --quickPass` and the dispatch-bound benchmark are both unchanged.

## The shape being removed

`TornadoVMInterpreter.execute()` was **245 lines**: setup, then a `while` loop whose body was a **24-arm `if (op == X.value()) ... else if ...` chain**. Every arm decoded its own operands inline, repeated the `isWarmup` skip test, and assigned `lastEvent`. Three consequences:

- **Dispatch cost grew with position.** `END` — the most executed opcode after the transfers — sat near the bottom of the chain, so reaching it meant ~20 comparisons.
- **Adding an execution mode meant editing every arm.** Adding a transfers-only pass in #1036 touched **12 of them**, purely to add `|| transfersOnly` next to `isWarmup`. That is the classic sign of behaviour that wants to be a policy, not a repeated condition.
- **Nothing had a name.** The 15 behaviours were anonymous blocks in one method, so nothing could be read, tested or referenced on its own.

## Change

```java
bytecodeLoop: while (bytecodeResult.hasRemaining()) {
    final TornadoVMBytecodes bytecode = TornadoVMBytecodes.fromValue(bytecodeResult.get());
    if (bytecode == null) {
        throwErrorInterpreter(op);
    }
    switch (bytecode) {
        case ALLOC -> lastEvent = handleAlloc(logBuilder, lastEvent, isWarmup);
        case LAUNCH -> lastEvent = handleLaunch(logBuilder, lastEvent, isWarmup);
        ...
    }
}
```

- **`TornadoVMBytecodes.fromValue(byte)`** — an opcode-indexed lookup array built once (opcodes are contiguous, 10–30). An unknown byte returns `null`, which the interpreter reports instead of guessing: previously an out-of-step stream fell to the final `else` only after failing 24 comparisons.
- **15 named handlers**, one per bytecode, each returning the new `lastEvent`. Operand decoding stays inside its handler — the reads are positional and must stay in order — but each behaviour now has a name, a signature and a body you can read without scrolling.
- `execute()` drops from **245 to 116 lines**, and is now setup + loop + epilogue.

The `switch` over an enum compiles to a `tableswitch`, so dispatch is one indexed jump regardless of opcode.

The `END` case keeps its `break` via a labelled loop rather than a sentinel return, so the control flow reads the same as before.

## Testing

Behaviour-preserving refactors get verified, not argued:

- `tornado-test --quickPass` on CUDA: **85 failures, equal to the `develop` baseline**.
- OpenCL, per class on the same device: `TestBatches` 29/29, `TestMultipleTasksSingleDevice` 4/4, `TestArrays` 23 run / 4 failed — all identical to baseline.
- `saxpy 100000 512`, the dispatch-bound benchmark that would notice a regression here: **17.41 / 17.51 / 17.44 µs** against a 17.46 µs baseline. No measurable change either way, which is the expected result: the chain was ~20 predictable comparisons against a ~17 µs iteration.

## What this unblocks

The transfers-only mode in #1036 becomes a property of the handlers that care rather than a condition repeated 12 times, and the same table can back a bytecode printer and a verifier instead of the third and fourth hand-written `switch` over opcodes. Neither is in this PR.

## Backends/OS tested

CUDA and OpenCL on NVIDIA RTX 4090, Ubuntu 24.04, JDK 21. The interpreter is backend-neutral, so Metal runs the same code path.

Part of #1028.


mairooni (approved):
LGTM, thanks!

PR #1043, opened 2026-08-19
Fix emitter miscompile of nested short-circuit booleans

mikepapadim:
# Fix: CUDA emitter miscompiles nested short-circuit boolean expressions (OQ-17)

Built and tested against upstream `develop` at `99549c986` (fetched and rebased fresh before
starting), CUDA backend only, RTX 4090 / sm_89.

## Symptom

`(c == 3) || ((c == 2) && (v == -1))` — the Game-of-Life B3/S23 survival rule, and any similarly
shaped `A || (B && C)` — disagreed with identical host Java on the CUDA backend. Also affected:
deeper nestings (`A || (B && C && D)`) and, independently, plain conjunction chains used directly
as an `if` condition without any enclosing `||` (`0 < idx && idx < n && 0 < i && i < n`).

## Minimal reproducer

`tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/logic/TestLogic.java`:
`testLogicOrOfAnd` (exhaustive over 256 inputs: 16 neighbour counts x {DEAD, ALIVE}), plus two
follow-up variants added while chasing this down: `testLogicOrOfAndAnd` (`A || (B && C && D)`,
exhaustive over 64 inputs) and `testLogicOrOfAndOr` (`A || (B && (C || D))`, exhaustive over 64
inputs).

```
tornado-test -V uk.ac.manchester.tornado.unittests.logic.TestLogic#testLogicOrOfAnd
    [REASON] index 0: count=0 cell=0 expected:<0> but was:<-1>
```

## Root cause

`tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/CUDANodeLIRBuilder.java`

Graal represents both `&&` and `||` as `ShortCircuitOrNode(x, xNegated, y, yNegated)`, meaning
`(xNegated ? !x : x) || (yNegated ? !y : y)`. `B && C` is encoded as `ShortCircuitOrNode(B, true,
C, true)` (`!B || !C`, so `NOT(that) == B && C`), and nested conjunctions/disjunctions chain
further `ShortCircuitOrNode`s with a mix of negated and non-negated legs.

Two related bugs, both in the same "compute this node's value" helpers:

**Bug 1 (`emitNegatedLogicNode`, was line 361).** This method computes `NOT(node)` and ended with
`setResult(node, result)` — caching the *negated* value under `node`'s own identity in
`nodeOperands`, a single-slot `NodeMap<Value>` per node (confirmed via `javap` on
`tornado-graal-23.1.0.jar` — there is no separate cache for a node's negated form). A later
*plain*-value lookup of the same node (`operandOrConjunction`, used when a `ShortCircuitOrNode`
leg is shared between its own materialization and a second, enclosing negation) then silently
read back the stale negated value. Confirmed directly with `-Dtornado.logger.buildlir=True`
(existing instrumentation, not added for this): the leaf nodes for `c==2` and `v==-1` each got
their negated form cached while the inner `B && C` node materialized itself, and that stale
negated form was then reused — instead of recomputed — when the outer `||` asked for `B && C`'s
plain value a second time.

Fix: don't write to the shared cache from this method; return the computed `Value` directly to
the caller, which is all every call site actually needs.

**Bug 2 (`emitLogicNode`'s `ShortCircuitOrNode` branch, was line ~420).** The sibling method
computing a node's *plain* value had the mirror-image bug: it always used both legs' plain values
(`operandOrConjunction(x)`, `operandOrConjunction(y)`), ignoring `isXNegated()`/`isYNegated()`
entirely. This is only correct when neither leg is negated. It was masked on `develop` because
`emitIf` calls `emitLogicNode` directly (bypassing the operand cache) to recompute an `if`
condition, and on the buggy `develop` code Bug 1's stale cache happened to already hold the
*correct* negated value for the leg that needed it — two bugs cancelling out by coincidence, for
this one shape. Fixing Bug 1 alone removed that accidental crutch and exposed Bug 2 as a new
failure: `TestSignedComparisonsCodegen#testSigned04` (`0 < idx && idx < n && 0 < i && i < n` as an
`if` condition, no `||` involved at all) started failing after the Bug 1 fix alone, caught by the
before/after full-suite diff. **This is why the fix touches both branches — it was not
found in the original diagnosis (`.work/prs/01-diagnosis.md`) and was located only by running the
full suite after the first fix.**

Fix: mirror Bug 1's fix — respect `isXNegated()`/`isYNegated()` on each leg, calling the new
`negatedOperand()` helper (parallel to the existing `operandOrConjunction()`, but always
recomputes via `emitNegatedLogicNode` since there is no cached negated form to reuse) when a leg
is negated.

Both `emitShortCircuitOrNode` (the third place a `ShortCircuitOrNode` gets a value, used when the
node is scheduled as its own floating LIR value) and `getProcessedOperand` were already correct —
they already dispatch on `isXNegated()`/`isYNegated()` per leg. Only the two on-demand
recomputation helpers had the bug.

## Before / after generated CUDA

`A || (B && C)` (`testLogicOrOfAnd`), `--printKernel`:

```c
// before (wrong)
b_14  =  i_10 != 2 || i_12 != -1;
b_15  =  i_10 == 3 || i_10 != 2 && i_12 != -1;   // NOT(B) && NOT(C), should be B && C
i_16  =  (b_15 == true) ? -1 : 0;

// after (fixed)
b_14  =  i_10 != 2 || i_12 != -1;
b_15  =  i_10 == 3 || i_10 == 2 && i_12 == -1;   // correct
i_16  =  (b_15 == true) ? -1 : 0;
```

`A || (B && C && D)` (`testLogicOrOfAndAnd`, only reachable after the Bug 1 fix - it failed before
that fix was in place too, at a different index):

```c
// after (fixed)
b_17  =  i_13 != -1 || i_15 != 1;
b_18  =  i_11 != 2 || b_17;
b_19  =  i_11 == 3 || i_11 == 2 && i_13 == -1 && i_15 == 1;   // correct
i_20  =  (b_19 == true) ? -1 : 0;
```

`0 < idx && idx < n && 0 < i && i < n` as a plain `if` condition (`testSigned04`), regression from
the Bug 1-only fix, resolved by the Bug 2 fix:

```c
// Bug-1-fix-only (wrong): b_6 should be NOT(isInBounds); idx-in-range leg is not negated
b_6  =  i_2 >= 0 && i_2 < 7 || b_4;

// after both fixes (correct): idx-in-range leg is properly negated
b_6  =  i_2 < 0 || i_2 >= 7 || b_4;
```

## Test evidence

`tornado-test -V uk.ac.manchester.tornado.unittests.logic.TestLogic` (9 tests, including the 3 new
ones): 9/9 PASS after the fix (all 3 new ones fail without it - `testLogicOrOfAndAnd` fails with
only the Bug 1 fix applied, `testLogicOrOfAnd`/`testLogicOrOfAndOr` fail with neither fix).

Full CUDA unit suite (`tornado-test --verbose --quickPass`, matching `make fast-tests`' core step),
diffed before vs. after:

| | before (this branch's parent, `develop`) | after (both fixes) |
|---|---|---|
| PASS | 1040 | 1043 (+3: the new tests) |
| FAILED | 8 | 8 (identical set, see below) |
| UNSUPPORTED | 90 | 90 |

Failed-test set is byte-identical before and after - all 8 are pre-existing, already
`__TORNADO_TESTS_WHITE_LIST__` entries in `tornado-assembly/src/bin/tornado-test`, untouched by
this change:
`TestReductionsFloats#testComputePi`,
`TestMatrixMultiplicationKernelContext#mxm2DKernelContext01`/`02`,
`ComputeTests#testMandelbrot`/`testJuliaSets`,
`TransformerKernelsTest#testReductionOneBlockWithLayer`,
`TestDevices#test04`/`test06`.

An intermediate build (Bug 1 fix only, before Bug 2 was found) newly failed
`TestSignedComparisonsCodegen#testSigned04` - this is exactly why the full-suite before/after diff
matters, not just the targeted reproducer; that regression is fixed in this same branch, not
carried over.

## Backends tested / not tested

CUDA only, per this session's scope. **Not built or tested:** OpenCL
(`tornado-drivers/opencl/.../OCLNodeLIRBuilder.java`) and Metal
(`tornado-drivers/metal/.../MetalNodeLIRBuilder.java`) contain the same
`emitNegatedLogicNode`/`emitLogicNode`/`ShortCircuitOrNode` structure, read but not touched or
verified in this change. If confirmed, they would need the identical fix, but that is a separate,
unverified claim - not included here.


mairooni (approved):
I applied the fix to the OpenCL and Metal backends as well. The CI pipeline passes. LGTM

PR #1051, opened 2026-08-28
[cuda] Cache the Graal compilation result per device so a new execution plan does not re-run the front end

mikepapadim:
## Problem

The in-memory code cache is scoped per execution plan:
`CUDADeviceContext` keeps a `Map<executionPlanId, CUDACodeCache>`, and plan teardown clears the
entry. A new `TornadoExecutionPlan` therefore starts empty and re-runs the entire Graal front
end for tasks it has already compiled.

Profiling a single-task graph on an RTX 4090 (n=32768), on a freshly created plan:

| Phase | Time |
|---|---|
| `TASK_COMPILE_GRAAL_TIME` | **3.28 ms** |
| `TASK_CODE_GENERATION_TIME` | 0.34 ms |
| `TASK_COMPILE_DRIVER_TIME` (NVRTC) | 0.19 ms |
| `TASK_KERNEL_TIME` (actual GPU work) | **0.005 ms** |

The front end costs roughly **620x the kernel it produces**. NVRTC is not the problem - the
on-disk cubin cache already covers it, and disabling it (`-Dtornado.cuda.codecache.enable=false`)
changes nothing measurable (2.777 vs 2.962 ms). The repeated cost is Graal-side.

This penalty is a fixed ~2.8 ms per plan regardless of problem size, so it hits hardest exactly
where TornadoVM is otherwise cheapest. It matters for any code that builds many short-lived
plans, which is the natural shape for a library that maps an existing API onto TornadoVM.

## Change

Installed code stays per execution plan - plan teardown invalidates the underlying module, and
sharing module handles across concurrent plans would not be safe. What is shared is the *result
of the front end*: the generated source is cached once per device and every later plan pays only
the install step.

`CUDAInstalledCode` already takes `executionPlanId` as a launch-time parameter and holds no
plan-specific state, and `CUDAContext` is created per platform, so this does not change any
module or stream lifetime.

New option `tornado.compilation.share` (default `True`) restores the old behaviour when set to
`False`.

## Results

Cost of creating a plan, running one task, and closing it (RTX 4090, CUDA 12.6, JDK 21):

| n | before | after | speedup |
|---|---|---|---|
| 4 096 | 2.886 ms | 1.226 ms | **2.35x** |
| 32 768 | 2.847 ms | 1.238 ms | **2.30x** |
| 262 144 | 3.349 ms | 1.745 ms | **1.92x** |
| 1 048 576 | 4.927 ms | 3.304 ms | **1.49x** |

`TASK_COMPILE_GRAAL_TIME` drops from 3.28 ms to 0.0011 ms. The reused-plan path is unchanged
(0.046 -> 0.046 ms at n=4096, 0.733 -> 0.727 ms at n=1M), i.e. no cost is added to code that
already reuses a plan. Residual fresh-plan cost is now buffer allocation, module install and
stream setup rather than Graal.

## Correctness

Two traps, both covered by the new tests:

1. **Shape analysis folds concrete loop bounds into the parallel domain**, so compilation is
   specialised to argument *sizes*, not just types. A type-only key silently under-computes: a
   4096-element array served from a 1024-element entry left 3072 elements stale. Array lengths
   and scalar values are part of the key.
2. **`meta.setDomain(...)` is a side effect of the front end** and is skipped once a domain is
   set. A cache hit that does not restore it launches the task with no iteration space, which
   measured ~48x slower at 1M elements. The domain is carried in the cache entry and restored.

The key is conservative: it covers task id, method, backend/device index, batch configuration
and argument shape, and **returns `null` for any argument whose shape cannot be characterised
exactly** - images, matrices and vector types carry their own dimensions, so rather than
enumerate them and risk a collision, those tasks compile as they did before. An earlier revision
that lacked this opt-out regressed `TestImages` (+11) and `vectortypes.TestFloats` (+1); with it,
the failing set is identical to baseline.

It deliberately does not consult `shouldCompile()`: `TornadoVMInterpreter` calls
`forceCompilation()` unconditionally for the last task of every graph, so honouring it would
disable the cache for every single-task graph. The states that genuinely invalidate code (batch
threads change, and a batch consuming the thread id) are functions of the batch configuration,
which the key includes, and `updateBatchThreads()` has already run by that point.

## Testing

- New `uk.ac.manchester.tornado.unittests.compiler.TestSharedCompilationCache` (3 tests, pass):
  repeated plans at one size, plans at differing sizes, and plans differing only in a scalar
  loop bound.
- `tornado-test --quickPass` on CUDA/RTX 4090, both arms of `tornado.compilation.share` on the
  same build: **1162 ran, 8 failed, 77 unsupported in both**, with an identical set of failing
  classes (all pre-existing: `TestDevices`, `ComputeTests`, `TransformerKernelsTest`,
  `TestMatrixMultiplicationKernelContext`, `TestReductionsFloats`).

## Notes

OpenCL and Metal have the same per-plan cache structure (`Map<Long, OCLCodeCache>`,
`Map<Long, MetalCodeCache>`) and the same opportunity. This PR is CUDA-only so the approach can
be reviewed on hardware I can actually measure on; the other backends are a natural follow-up.

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/runtime/CUDATornadoDevice.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/runtime/CUDATornadoDevice.java
@@ -64,11 +64,13 @@ import uk.ac.manchester.tornado.api.types.arrays.Int8Array;
 import uk.ac.manchester.tornado.api.types.arrays.IntArray;
 import uk.ac.manchester.tornado.api.types.arrays.LongArray;
 import uk.ac.manchester.tornado.api.types.arrays.ShortArray;
+import uk.ac.manchester.tornado.api.types.arrays.TornadoNativeArray;
 import uk.ac.manchester.tornado.api.memory.NativeArrayLayouts;
 import uk.ac.manchester.tornado.drivers.common.TornadoBufferProvider;
 import uk.ac.manchester.tornado.drivers.cuda.CUDABackendImpl;
 import uk.ac.manchester.tornado.drivers.cuda.CUDACodeCache;
 import uk.ac.manchester.tornado.drivers.cuda.CUDACommandQueue;
+import uk.ac.manchester.tornado.drivers.cuda.CUDACompiledKernel;
 import uk.ac.manchester.tornado.drivers.cuda.CUDADeviceContext;
 import uk.ac.manchester.tornado.drivers.cuda.CUDADeviceContextInterface;
 import uk.ac.manchester.tornado.drivers.cuda.CUDATargetDevice;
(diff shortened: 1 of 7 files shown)


tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/runtime/CUDATornadoDevice.java
```diff
@@ -267,11 +269,41 @@ private TornadoInstalledCode compileTask(long executionPlanId, SchedulableTask t
...
+            // earlier plan on this device already produced it, so a new plan pays only the
+            // install step below.
+            //
+            // This deliberately does not consult shouldCompile(). The interpreter calls
+            // forceCompilation() unconditionally for the last task of every graph, so honouring
+            // it would disable this cache for any single-task graph. The states that genuinely
+            // invalidate generated code - a change of batch threads, and a batch that consumes
+            // the thread id - are both a function of the batch configuration, which
+            // compilationCacheKey() includes; updateBatchThreads() has already run by the time
+            // this executes, so the key reflects the batch about to be compiled.
```
mikepapadim:
```suggestion
```

PR #1055, opened 2026-08-28
[curand] Add a cuRAND library-task provider for device-side random number generation

mikepapadim:
## What

A cuRAND library-task provider, so a TornadoVM program can generate random numbers **into a buffer
that is already on the device** instead of generating them on the host and uploading them.

```java
TaskGraph graph = new TaskGraph("rng")
        .transferToDevice(DataTransferMode.FIRST_EXECUTION, buffer)
        .libraryTask("normals", CuRand::generateNormal, buffer, offset, count, 0.0f, 1.0f)
        .persistOnDevice(buffer);
```

Follows the existing cuBLAS/cuSPARSE/cuTENSOR layout: `tornado-curand` for the Java side and
`tornado-drivers/curand-jni` for the JNI, registered in the aggregator poms, the assembly and
`tornado.py`.

Covers `curandGenerateNormal` / `curandGenerateNormalDouble` and
`curandGenerateUniform` / `curandGenerateUniformDouble`, for `FloatArray` and `DoubleArray`,
writing at an element offset so a caller can fill one slice of a larger buffer.

## Why

Monte-Carlo simulation on the GPU is often limited by getting the random numbers there rather than
by the arithmetic. Measured on a finmath-lib Black-Scholes valuation ported to TornadoVM
(RTX 4090, 1M paths, 40 time steps):

| phase | CPU | GPU |
|---|---|---|
| model arithmetic | 483.7 ms | 23.3 ms |
| Brownian increments | 430.4 ms | 1031.6 ms |

The arithmetic is 20.8x faster and completely hidden, because the increments were still generated
by a host RNG and then uploaded - 160 MB of transfers, on top of paying the same host RNG cost the
CPU baseline pays. With this provider a 1M-element draw of normals costs **0.198 ms**, against
about 25.8 ms per draw for generate-on-host-and-upload: **~130x** on that phase.

## Details worth review

- **The generator is bound to the execution plan's CUDA stream** (`curandSetStream`), so generation
  is ordered against the kernels and transfers of the same task graph, exactly as the cuBLAS
  provider does.
- **The seed is applied once, when the generator is created, not per dispatch.** cuRAND restarts
  the sequence whenever a seed is set, so re-seeding on each call would hand out the same block of
  numbers every time. `testSuccessiveExecutionsDiffer` pins this.
- **The destination is declared `READ_WRITE`, not `WRITE_ONLY`.** A call typically fills a slice of
  a larger buffer; declaring it write-only tells TornadoVM the whole buffer is produced here and
  the region outside the slice would stop being valid. `testGenerateIntoSlice` pins this.
- Element offsets are converted to byte offsets in the native layer, so the Java side never deals
  in element sizes. `getDevicePointer` already points past the TornadoVM array header.

## Testing

New `uk.ac.manchester.tornado.unittests.curand.TestCuRand`, 6 tests, all passing on an RTX 4090
(CUDA 12.6, JDK 21): distribution mean and standard deviation for normals, mean and range for
uniforms, a non-default mean/stddev, generation into a slice with the rest of the buffer left
intact, successive executions producing different numbers, and the double-precision path.

`tornado-test --quickPass` with `BACKEND=cuda`, against unmodified `develop` on the same machine:
**1168 vs 1162 ran (the 6 new tests), 8 failed, 77 unsupported in both**, with an identical set of
failing classes (all pre-existing). `./mvnw checkstyle:check` clean.

CMake locates cuRAND from the CUDA toolkit the same way the cuBLAS module locates cuBLAS, and warns
and falls back to `-lcurand` if it cannot.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -816,6 +816,7 @@
         <module>tornado-api</module>
         <module>tornado-annotation</module>
         <module>tornado-cublas</module>
+        <module>tornado-curand</module>
         <module>tornado-cufft</module>
         <module>tornado-cudnn</module>
         <module>tornado-cusparse</module>
--- a/tornado-assembly/assembly.xml
+++ b/tornado-assembly/assembly.xml
@@ -178,6 +178,7 @@
                 <include>io.github.beehive-lab:tornado-benchmarks</include>
                 <include>io.github.beehive-lab:tornado-unittests</include>
                 <include>io.github.beehive-lab:tornado-cublas</include>
+                <include>io.github.beehive-lab:tornado-curand</include>
                 <include>io.github.beehive-lab:tornado-cufft</include>
                 <include>io.github.beehive-lab:tornado-cudnn</include>
                 <include>io.github.beehive-lab:tornado-cusparse</include>
--- a/tornado-assembly/pom.xml
+++ b/tornado-assembly/pom.xml
@@ -107,6 +107,11 @@
                 </dependency>
                 <dependency>
                     <groupId>${project.groupId}</groupId>
+                    <artifactId>tornado-curand</artifactId>
+                    <version>${project.version}</version>
+                </dependency>
+                <dependency>
+                    <groupId>${project.groupId}</groupId>
                     <artifactId>tornado-cufft</artifactId>
                     <version>${project.version}</version>
                 </dependency>
--- a/tornado-assembly/src/bin/tornado.py
+++ b/tornado-assembly/src/bin/tornado.py
@@ -71,6 +71,7 @@ __CUDA_MODULE__ = "tornado.drivers.cuda"
 __ENABLE_NATIVE_ACCESS__ = "--enable-native-access="
 __FFM_MODULE__ = "tornado.runtime"
 __CUBLAS_MODULE__ = "tornado.cublas"
+__CURAND_MODULE__ = "tornado.curand"
 __CUFFT_MODULE__ = "tornado.cufft"
 __CUDNN_MODULE__ = "tornado.cudnn"
 __CUSPARSE_MODULE__ = "tornado.cusparse"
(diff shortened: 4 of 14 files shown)


mikepapadim:
## Rebased onto `develop`, and the binding is now FFM rather than JNI

Two changes since the last push.

### 1. Rebased onto `3ca741b49`

The branch was 60 commits behind and `dirty`. Both conflicts were module lists, and **neither side could be taken verbatim**: `develop` has migrated `cuda-jni`, `cublas-jni` and `cufft-jni` to FFM and deleted those directories, while this branch re-added all three alongside its own `curand-jni`. Taking this side would have registered three modules that no longer exist. Resolved by keeping only what this PR introduces.

### 2. The binding is now FFM, and `curand-jni` is gone

Given `develop` has moved every other provider onto `java.lang.foreign`, landing this one on JNI would have meant merging it and immediately migrating it. It is now FFM-native, and the JNI module is **deleted rather than bypassed**:

- `tornado-drivers/curand-jni/` — removed (pom, `CMakeLists.txt`, `tornado-curand.cpp`)
- `tornado-drivers/pom.xml` — module entry dropped
- `tornado-assembly/pom.xml`, `assembly.xml` — dependency and libs include dropped

The C++ shim was a pure passthrough, so nothing was lost. Ten `native` methods became nine downcall handles resolved from `libcurand.so.10`. Two pieces of logic moved into Java:

```java
// curandCreateGenerator writes the handle through an out-parameter
try (Arena arena = Arena.ofConfined()) {
    MemorySegment handle = arena.allocate(C_POINTER);
    if ((int) CURAND_CREATE_GENERATOR.invokeExact(handle, rngType) != CURAND_STATUS_SUCCESS) {
        return 0;
    }
    return handle.get(C_POINTER, 0).address();
}

// the slice offset, previously C pointer arithmetic
CURAND_GENERATE_UNIFORM.invokeExact(generator, devicePtr + elementOffset * Float.BYTES, n);
```

**No new build wiring.** `--enable-native-access` is already granted to `tornado.runtime`, where `FFMSupport` performs every restricted lookup on the providers' behalf, and `tornado.curand` is already in the launcher's `--add-modules` list. `tornado.py` even carries a comment saying the providers "use `java.lang.foreign` rather than a JNI library of their own" — cuRAND was the one case contradicting it.

### 3. Test availability guard

`TestCuRand` had no `@Before` guard, unlike every other library-task suite, so it would **fail** rather than skip when the default device is not CUDA or cuRAND is missing. Added, along with the `CuRandLibraryProvider.isAvailable()` accessor it needs, mirroring `CusparseLibraryProvider`.

### Verification

RTX 4090 (sm_89), driver 565.57.01, CUDA 12.6.85, JDK 25.0.2, `make jdk22plus BACKEND=cuda`:

```
BUILD SUCCESS
tornado-test -V uk.ac.manchester.tornado.unittests.curand.TestCuRand
Test ran: 6, Failed: 0, Unsupported: 0
```

`testGenerateIntoSlice` is the one that matters for the port — it exercises the element-offset arithmetic that moved out of C++. And the built SDK now contains only `tornado-curand.jar`, with no `libtornado-curand` native library, confirming the module is genuinely gone rather than merely unused.

I amended rather than adding a second commit, so the PR reads as "add a cuRAND provider (FFM)" instead of adding a JNI module and then deleting it. Previous head was `1c81965bd` if you want the pre-FFM state.

orionpapadakis (approved):

PR #1058, opened 2026-08-28
Replace JNI with the Java FFM API in the backends and library tasks

mikepapadim:
Removes the hand-written JNI libraries from the OpenCL and CUDA backends and from the
cuBLAS, cuBLASLt, cuFFT and cuSPARSE library-task providers, calling the native libraries
directly through `java.lang.foreign`. Draft: two providers remain, for a reason given below.

### What is gone

| module | before | after |
|---|---|---|
| `tornado-drivers/opencl-jni` | 3,130 lines of C++ | deleted |
| `tornado-drivers/cuda-jni` | 3,198 lines of C++ | deleted |
| `tornado-drivers/cublas-jni` | 419 lines of C++ | deleted |
| `tornado-drivers/cufft-jni` | 158 lines of C++ | deleted |
| `tornado-drivers/cusparse-jni` | 203 lines of C++ | deleted |
| `tornado-drivers/cudnn-jni` | 536 lines of C++ | **313 deleted**, SDPA shim kept |

An OpenCL SDK now ships an **empty `lib/`**. A CUDA SDK ships two libraries where it used to
ship six, and one of those two is down to four JNI entry points. Roughly 7,400 lines of
C/C++/Objective-C removed.

### Test results

One box (RTX 4090, driver 565.57, CUDA 12.6, JDK 21), `tornado-test --quickPass`, each
backend against a same-commit build of `develop`:

| backend | baseline | this branch |
|---|---|---|
| OpenCL | 1167 ran / 7-8 failed / 274 unsupported | **1167 / 7-8 / 274** |
| CUDA | 1164 ran / 8 failed / 77 unsupported | **1164 / 8 / 77** |

Same failing classes in both columns. `make checkstyle` is green.

The OpenCL range is not hedging: `TransformerKernelsTest#testReductionOneBlockWithLayer` is
**flaky on OpenCL on `develop` itself**, measured over 20 runs per side on the same box:

| | pass | fail |
|---|---|---|
| `develop` | 9 | 11 |
| this branch | 13 | 7 |

Roughly a coin flip either way; the difference is about 1.3σ at n=20, so the two are
indistinguishable. Whichever way that one lands decides whether an OpenCL sweep reports 7 or 8.
Worth someone looking at separately: the test's input is deterministic (`new Random(7)`, and
JUnit gives each method a fresh instance), so the variation is in the device-side reduction, not
the data — it is a `KernelContext` reduction over local memory, and it also fails consistently on
the CUDA backend in both baseline and branch. Pre-existing, and unrelated to this PR.

Library tasks additionally exercised directly — all pass:
`TestCuBlas` (14), `TestCuBlasLt` (6), `TestCuFft` (8), `TestCusparse` (11), `TestCuDnn` (11),
`TestNvtx` (4), plus the nine `tornado-cublas` example/benchmark mains, including CUDA-graph
capture, the TF32 and FP16 GemmEx paths, and the cuBLASLt fused GELU+bias epilogue
(100 TFLOP/s, results matching).

### Performance

Measured on an RTX 4090 (driver 565.57, CUDA 12.6, JDK 21), same machine, builds alternated.
Detail in the comments; the summary is that **real inference gets faster and nothing gets slower
once warm**.

[GPULlama3.java](https://github.com/beehive-lab/GPULlama3.java) `--bench`, Llama-3.2-3B-Q8_0,
`-p 512 -n 128 -r 5`, **3 full invocations per configuration**, same application jar on both SDKs:

| configuration | JNI (develop) | FFM (this branch) | delta |
|---|---|---|---|
| **tg128** decode, default | 105.09 t/s `[104.92-105.22]` | 108.45 t/s `[108.33-108.57]` | **+3.2%** |
| **pp512** prefill, default | 63.46 t/s `[63.35-63.54]` | 64.62 t/s `[64.56-64.65]` | **+1.8%** |
| tg128, `--cuda-graphs` | 134.97 t/s | 135.08 t/s `[134.97-135.14]` | +0.1% |
| pp512, `--cuda-graphs` | 73.18 t/s | 73.22 t/s `[73.12-73.30]` | +0.1% |

Ranges do not overlap on the default path. The gain appears where per-call host cost is on the
critical path — decode issues many small kernels per token, faster than the GPU needs them — and
disappears under `--cuda-graphs`, which collapses that host work into a single graph launch. Note
this branch also removed per-call work while porting (interned NVTX labels, a reused scratch slot
instead of an arena per call), so the honest claim is that *the FFM issue path as written* is
cheaper than the JNI one it replaces, not that the binding mechanism alone accounts for it.

Synthetic dispatch benchmarks show **parity**: 28.59 vs 28.60 us per dispatch at steady state, six
runs. There, each `execute()` blocks on the GPU — JFR puts 82.9% of host time in
`cuStreamSynchronize` — so issue cost is masked and neither build can pull ahead.

The one cost is **cold start: ~13.5 ms, paid once**, with dispatches 10-1,000 running 14-22% slower
while the method handles link and the issue path JITs. Steady state is reached at ~10,000
dispatches. For scale, the first dispatch costs 72 ms on *both* builds, dominated by Graal and NVRTC
compilation.

### Bugs this surfaced

Three real defects the JNI layer had been hiding, all fixed here:

- **Buffer overrun in the OpenCL device-to-device copy.** `mapOnDeviceMemoryNDRegion` took
  `headerSize` as an element count but subtracted it from `sizeDest` as a byte count, so every
  copy ran three floats past the end of the destination mapping. C does not bounds-check a
  mapped pointer; `MemorySegment` does, which is how it showed up.
- **`OCLKernel.clSetKernelArgRef` was declared and called but never implemented** in the JNI
  library, so `setArgRef` would have thrown `UnsatisfiedLinkError` if reached.
- **The CUDA CI runner pin was justified by a reason that no longer holds** — the comment said
  `libtornado-cuda.so` baked in an NVRTC soname. Corrected to name the library-task modules,
  which is what still constrains it.

### Design notes for review

- **Source compiles under both JDK profiles.** `jdk21` builds with `--enable-preview` (FFM was
  preview there), `jdk22plus` with `--release 22`. `FFMSupport` therefore uses only the
  intersection of the two API surfaces — `allocateFrom`/`getString` do not exist in 21,
  `allocateUtf8String` does not exist in 22.
- **Every restricted call funnels through one module.** `FFMSupport` lives in `tornado.runtime`,
  which everything already requires, and `bin/tornado` grants native access to exactly that
  module. `Linker::upcallStub` leaked out of it once during development and failed immediately,
  which is the intended behaviour; it now goes through the same helper rather than widening the
  grant.
- **CUDA handles became registry indices rather than raw pointers.** The JNI layer boxed CUDA
  primitives in malloc'd structs and passed their addresses as `long`. A stale handle now
  resolves to `null` instead of dereferencing whatever is at that address, which let the
  pointer-plausibility heuristic in `clWaitForEvents` go away. OpenCL and cuBLAS handles are
  real pointers and are passed through unchanged.
- **Java-array transfers stage through a reused per-thread native buffer.** A downcall cannot
  address the heap the way `GetPrimitiveArrayCritical` could. The runtime's own segment-backed
  arrays never take this path.
- **Library availability is now a question about the CUDA Toolkit, not about TornadoVM.** The
  unit-test guards used to probe `System.loadLibrary("tornado-cusparse")`; they now ask
  `CusparseLibraryProvider.isAvailable()`, which is what they actually meant.

### `TestIO#testCopyInWithDevice`

This test compares two 40-iteration transfer-time windows within one JVM. It was flaky on this
branch (7/10 on CUDA, against 10/10 on `develop`) for a reason that had nothing to do with what
it checks: the profiler's device timestamps bracket the host work that issues a copy, so the
first measured window was also measuring the JIT ramp of a Java issue path — which a JNI one
does not have. Steady-state transfer time was already at parity (626 µs vs 632 µs over 40
copies).

Two changes make it robust rather than merely passing:

- **A warm-up before the measured windows**, followed by `freeDeviceMemory()` so both windows
  still start from a freed device and each pays exactly one `FIRST_EXECUTION` copy. Warming up
  without that free would trade one asymmetry for another, leaving `arrayA` resident for the
  first window but not the second.
- **The bound widened from 25% to 50%.** What the test is actually checking is that
  `withDevice` does not turn a `FIRST_EXECUTION` transfer into an every-execution one, which
  would roughly *double* the second window. A tighter bound was failing on scheduling noise,
  not on the behaviour under test.

Now **12/12 on CUDA and 12/12 on OpenCL**. The backend-side work that closed most of the
original gap is still in: a per-thread scratch slot instead of an arena per call, interned NVTX
labels, and an issue-path warm-up at queue creation, which together took the first copy from
216 µs to 48 µs against 38 µs on `develop`.

### What cannot be ported, and why

FFM can only call a **C ABI**. Two providers are not C libraries:

- **CUTLASS** (`tornado-cutlass.cu`) is a header-only C++ template library compiled by nvcc.
  The `.cu` shim *is* the kernel instantiation, not a wrapper around one. It has to stay.
- **cuDNN's SDPA path** (`tornado-cudnn-sdpa.cpp`) uses `cudnn_frontend.h`, a C++ graph API
  built on `std::shared_ptr`. Also has to stay.

`cudnn-jni` is now **split along that line**. The 25 plain-C entry points (softmax, activation,
pooling, convolution, handle and workspace management) go through FFM; `tornado-cudnn.cpp` is
deleted and the CMake builds only `tornado-cudnn-sdpa.cpp`. `libtornado-cudnn.so` is down to
four exported JNI symbols, all SDPA. The shim is also now loaded **lazily, only on the SDPA
path**, so the other cuDNN ops keep working on a build where it is absent — which is the case
whenever the toolkit is too old for cudnn-frontend, as on the CUDA 11 CI leg.

### Metal: boilerplate only, and deliberately not wired up

Metal has no C API — `MTLDevice` and friends are Objective-C protocols — so the port has to go
through the Objective-C runtime, which *is* a C ABI. Two new classes lay that foundation:

- `ObjCRuntime` — `objc_getClass`, `sel_registerName`, and a cache of `objc_msgSend` downcall
  handles. The rule that makes or breaks this port is documented there: `objc_msgSend` is not
  really variadic, and must be called through a prototype matching the *target method's*
  signature, so there is one handle per signature rather than one shared handle. Also covers
  `NSString` conversion, retain/release (there is no ARC without a compiler), and an
  autorelease-pool scope.
- `MetalAPI` — the message sends the backend needs, each paired with its signature: device
  discovery and properties, queues, buffers, MSL compilation, pipeline state, encoding,
  dispatch, blit copies and GPU timestamps.

**This is untested.** It compiles, and the shapes come from the Metal headers and the
`objc_metal_jni.mm` shim, but there is no Apple hardware on the machine it was written on, so
nothing has been executed. `metal-jni` is untouched and the Metal backend still uses it — the
new classes are additive and unreferenced, so there is no regression risk to Metal in this PR.

`tornado-drivers/metal/METAL_FFM_PORT.md` is the handoff: what exists, what is left, and the
three rules that decide whether this works — that `objc_msgSend` is not really variadic and
needs one downcall handle per signature (getting that wrong is undefined behaviour, not a type
error); that there is no ARC, so autoreleased command buffers and encoders need a pool; and
that the code is arm64-only as written, with the x86_64 `_stret`/`_fpret` split documented and
routed through one place.

### Still to do

- [ ] Wire the Metal backend onto `MetalAPI` and verify on Apple silicon, then drop `metal-jni`.
- [ ] NVML power metrics on the CUDA backend are still stubbed at zero, exactly as the JNI was.
      The OpenCL side reads real power and is fully ported.
- [x] PTX backend is deliberately untouched (slated for removal in 6.0.0).


mikepapadim:
## Performance: JNI vs FFM, measured

A/B on one machine (RTX 4090, driver 565.57, CUDA 12.6, JDK 21), CUDA backend, `develop` at
`7fcd11b10` versus this branch, both built from source and run back to back. Workloads are the
project's own benchmarks via `BenchmarkRunner`.

### Headline: no steady-state regression

`saxpy` at 16 K elements is the most host-sensitive workload available — the kernel is a few
microseconds of GPU work, so per-dispatch host cost dominates. At **400,000 iterations**:

| metric | develop (JNI) | this branch (FFM) | delta |
|---|---|---|---|
| median | 29.59 µs | 29.53 µs | **−0.2%** |
| average | 29.51 µs | 29.33 µs | −0.6% |
| best | 15.11 µs | 14.12 µs | **−6.6%** |

Identical within noise, on the workload most likely to show a difference. Workloads with real GPU
work show nothing at all (5 repeats, median of medians, 500/60 iterations):

| benchmark | develop (JNI) | this branch (FFM) | delta |
|---|---|---|---|
| `saxpy 500 1048576` (transfer-bound) | 504.6 µs | 506.8 µs | +0.4% |
| `dft 60 2048` (compute-bound) | 551.0 µs | 577.9 µs | +4.9% |
| `sgemm 60 1024 1024` (compute-bound) | 1164.6 µs | 1164.5 µs | −0.0% |

### The one real cost: warm-up

Short runs are slower, and this is worth stating plainly rather than burying:

| `saxpy 16384` | develop (JNI) | this branch (FFM) | delta |
|---|---|---|---|
| median @ **500** iterations | 54.0 µs | 69.4 µs | **+28.6%** |
| median @ **400,000** iterations | 29.6 µs | 29.5 µs | −0.2% |
| first iteration (all benchmarks) | — | — | **+20% to +38%** |

A JNI call is compiled C from its first invocation. An FFM downcall has to link its method handle
and then get the surrounding Java JIT-compiled, so the issue path ramps over the first few
thousand dispatches. The queue-creation warm-up already in this PR removes the pathological first
call (216 µs → 48 µs), but not the whole ramp.

**Who this affects:** a long-running process — an inference server, a benchmark loop, anything that
executes a plan more than a few thousand times — pays nothing. A one-shot script that dispatches a
handful of kernels and exits pays 20–38% on those dispatches, on top of an unchanged multi-second
JVM and Graal startup. That trade seemed worth making; flagging it so it is an informed choice
rather than a silent one.

### JFR: where the host time actually goes

`-XX:StartFlightRecording=settings=profile` over the 400 K-iteration run, ~12 s of steady state:

| | develop (JNI) | this branch (FFM) |
|---|---|---|
| `jdk.ExecutionSample` (Java CPU) | 8 | 13 |
| `jdk.NativeMethodSample` | 517 | 485 |

Java-side CPU is ~2% of samples on both — neither build is limited by Java execution. The host
thread is essentially always inside a native call, and the top frames say which:

```
develop (JNI)                                    this branch (FFM)
  86.7%  CUDACommandQueue.readArrayFromDeviceOffHeap    82.9%  CUDADriverAPI.cuStreamSynchronize
   6.0%  CUDACommandQueue.writeArrayToDevice             5.2%  CUDADriverAPI.cuMemcpyHtoDAsync
   3.1%  CUDACommandQueue.clEnqueueNDRangeKernel         3.7%  CUDADriverAPI.cuMemcpyDtoHAsync
   1.9%  CUDAEvent.clReleaseEvent                        3.3%  CUDADriverAPI.cuLaunchKernel
                                                         2.1%  CUDADriverAPI.cuEventDestroy
```

Both agree the time is spent blocked on the GPU. But they say so with very different resolution,
and that is a side benefit worth noting: under JNI, 86.7% of host time lands in
`readArrayFromDeviceOffHeap`, a single opaque frame that internally does an async copy *and* a
stream synchronise *and* event bookkeeping — the profile cannot tell you which. Under FFM the same
time is attributed to the driver entry points themselves, so it is visible that this is
`cuStreamSynchronize` waiting on the device rather than copy cost. **Every driver call the backend
makes is now individually visible to JFR, with no instrumentation.**

### Method and caveats

- Benchmarks run through the project's `BenchmarkRunner`; medians reported, 5 repeats for the short
  runs, single 400 K-iteration run for steady state and JFR.
- `--add-opens tornado.benchmarks/...=tornado.runtime` was needed to run `BenchmarkRunner` directly
  on either branch — a pre-existing gap in the benchmark module, unrelated to this PR.
- One machine, one GPU, CUDA backend. The OpenCL backend was not profiled; its FFM path makes
  strictly fewer driver calls per operation than CUDA's (no per-transfer event pair or capture
  check), so it should be no worse, but that is inference and not measurement.
- The `dft` +4.9% median sits between the ~0% of the other compute-bound cases and the ramp effect;
  at 60 iterations it is plausibly still ramping rather than a real difference.

mikepapadim:
Cross-reference for the OpenCL "7–8 failed" range in the description: the variable one is
`TransformerKernelsTest#testReductionOneBlockWithLayer`, now filed separately as #1059.

It is an inter-workgroup race in the test kernel — thread `gid == 0` sums every workgroup's partial
result while the only synchronisation is `context.localBarrier()`, which orders threads within a
workgroup and gives no guarantee that the other groups have written theirs. It reproduces on
`develop` with no changes (9 pass / 11 fail over 20 runs on OpenCL, 0/20 on CUDA), so it is not
caused by this PR and is not fixed by it either.

mikepapadim:
## Follow-up: cold start vs steady state, and where the crossover is

The previous comment showed a 28.6% gap at 500 iterations and none at 400,000, which left the
obvious question unanswered: **how long is the ramp, and what does it actually cost?** Measured
properly this time.

### Method

The project's `BenchmarkRunner` only reports aggregates, so this uses a standalone probe that times
every individual `execute()` and buckets them — one run gives the whole curve instead of inferring
it from runs of different lengths. 200,000 dispatches of a 16 K-element saxpy (host-to-device copy,
tiny kernel, device-to-host copy), **3 runs per build**, same machine, back to back. Reporting the
**p10** per bucket: the median is contaminated by GPU clock transitions (below), p10 is not.

### The curve

| dispatches | JNI p10 | FFM p10 | delta |
|---|---|---|---|
| 1st (cold) | 71,978 µs | 78,306 µs | +8.8% |
| 1..10 | 116.34 µs | 130.97 µs | +12.6% |
| 10..100 | 66.97 µs | 81.17 µs | **+21.2%** |
| 100..500 | 31.64 µs | 38.66 µs | **+22.2%** |
| 500..1,000 | 18.49 µs | 21.16 µs | +14.4% |
| 1,000..5,000 | 18.44 µs | 16.96 µs | −8.0% ⚠ |
| 5,000..10,000 | 28.57 µs | 17.88 µs | −37.4% ⚠ |
| 10,000..25,000 | 28.60 µs | 28.59 µs | **−0.0%** |
| 25,000..50,000 | 28.59 µs | 28.59 µs | **+0.0%** |
| 50,000..100,000 | 28.59 µs | 28.59 µs | **+0.0%** |
| 100,000..200,000 | 28.60 µs | 28.59 µs | **−0.0%** |

**Steady state is reached at roughly 10,000 dispatches**, and past that the two are identical to
two decimal places — 28.59 vs 28.60 µs, with all three runs of each build landing inside
28.59–28.60. There is no measurable per-call cost to the FFM path once warm.

⚠ **The two rows marked are not an FFM win.** Both builds step from ~18 µs to ~28.6 µs at some
point in that region — the GPU settling from a boost clock into its sustained clock under
continuous load. JNI issues faster while ramping, so it reaches sustained load sooner (~5,000
dispatches) than FFM (~10,000), and the comparison in between is measuring which build crossed that
threshold first, not which binding is quicker. Both land on the same number afterwards. I would not
quote those two rows as anything.

### What the ramp actually costs

Summing the extra host time across the whole warm-up region:

| dispatches | n | extra on FFM |
|---|---|---|
| 1st | 1 | 6.33 ms |
| 1..10 | 9 | 0.13 ms |
| 10..100 | 90 | 1.68 ms |
| 100..500 | 400 | 4.39 ms |
| 500..1,000 | 500 | 1.01 ms |
| **total** | **1,000** | **13.5 ms** |

**The entire cold-start penalty is about 13.5 ms, paid once.** For scale, the first dispatch alone
costs 72 ms on *both* builds — Graal compiling the kernel and NVRTC compiling the generated CUDA C.
So moving to FFM adds roughly 19% to the cost of the first dispatch, and about 13.5 ms to a process
overall, against a startup that is already dominated by compilation and is measured in hundreds of
milliseconds.

### Reading

- **Any workload that dispatches more than ~10,000 times** — inference serving, iterative solvers,
  benchmarks, anything in a loop — sees no difference at all.
- **A one-shot process** pays ~13.5 ms more. That is real but small next to the compile costs it
  already pays, and it is not on any hot path.
- The earlier "+28.6% at 500 iterations" was a true measurement of the ramp region and a misleading
  measurement of the backend: at 500 iterations the plan has not finished warming up on either
  build, and FFM is still 14–22% behind.

If the cold-start cost ever matters, the lever is the queue-creation warm-up already in this PR
(`CUDACommandQueue.warmUpIssuePath`, currently 256 iterations). Raising it trades startup time for
a shorter ramp; at 256 it removes the pathological first call without measurably lengthening
startup. I left it there rather than tuning against this one workload.

<sub>RTX 4090, driver 565.57, CUDA 12.6, JDK 21, CUDA backend. Probe source is not in the branch —
it is a measurement tool, not something the project should carry.</sub>

mikepapadim:
## What this actually buys — and what it doesn't

Both performance comments above are neutral-to-slightly-negative, so it is worth stating the case
for this change explicitly rather than leaving the measurements to imply one.

**It is not a performance change.** Steady state is identical — 28.59 vs 28.60 µs p10 across six
runs — and cold start costs about 13.5 ms more. If throughput were the justification, this PR would
not have one.

### What it buys

1. **The build stops needing a C++ toolchain.** An OpenCL SDK now ships an empty `lib/`: no CMake
   step, no compiler, nothing to cross-compile per platform. The diff is `+5 / −1,253` lines of
   CMake and pom. In practice this is the largest effect.

2. **The SDK stops being coupled to the toolkit that built it.** This one was load-bearing: the
   CUDA CI legs are pinned to a single runner because `libtornado-cuda.so` baked in an NVRTC soname,
   so an SDK built against CUDA 11.5 would not load on a 12.x box — the comment in
   `build-test.yml` said exactly that, and this PR corrects it. NVRTC is now resolved by soname at
   run time, so that constraint is gone for the backend itself. It survives only for the two
   providers that cannot be ported, which is a much smaller pin than before.

3. **It surfaced two real defects**, because Java checks what C did not:
   - the OpenCL device-to-device copy had been running 12 bytes past the end of its destination
     mapping on **every** call — C does not bounds-check a mapped pointer, `MemorySegment` does;
   - `OCLKernel.clSetKernelArgRef` was declared and *called* from Java but never implemented in the
     JNI library, so `setArgRef` was a latent `UnsatisfiedLinkError`.

4. **Handles stopped being raw pointers.** CUDA handles were `malloc`'d struct addresses passed as
   `long`; a stale one dereferenced whatever was at that address. They are registry indices now, so
   a stale handle resolves to `null` — which is what let the pointer-plausibility heuristic in
   `clWaitForEvents` be deleted, a workaround that existed because it had already caused a SIGSEGV.

5. **Better profiles, measured.** JFR attributes host time to `cuStreamSynchronize`,
   `cuMemcpyDtoHAsync` and `cuLaunchKernel` individually, where previously 86.7% of it landed in a
   single opaque `readArrayFromDeviceOffHeap` frame that hid copy versus synchronise versus event
   work. Every driver call is now visible with no instrumentation.

6. **JNI is on a restriction path.** JEP 472 (JDK 24) warns on JNI use in preparation for requiring
   opt-in. Both mechanisms end up behind `--enable-native-access`, so this is direction of travel
   rather than a cliff — worth weighing, not worth overstating.

### What it costs

- **13.5 ms of cold start**, and 14–22% on dispatches 10–1,000. Nothing past ~10,000 dispatches.
- **A new failure mode.** A wrong `FunctionDescriptor` is undefined behaviour, not a type error: it
  will not throw, it will read garbage. That is manageable for a C library whose signatures come
  from a header. It is the main risk in the Metal work, where every `objc_msgSend` needs a
  hand-written prototype — which is why `METAL_FFM_PORT.md` leads with it.
- **"Remove JNI completely" is not reachable.** CUTLASS is nvcc-compiled C++ templates and cuDNN's
  SDPA is a C++ graph API; neither has a C ABI, so both keep a shim whatever happens. CUDA goes
  from six native libraries to two, not to zero.

### Reading

The case rests on build and distribution simplification, with the safety and observability effects
as secondary. It is a maintainability change, not a speed one. If the project's priority right now
is throughput, this is a lateral move and should be judged on those other axes — or not merged.
Better to say that plainly than to let the two bugs it happened to surface do the arguing.

mikepapadim:
## GPULlama3 (real LLM inference): FFM is faster on the host-bound path

Final check on a real workload rather than a microbenchmark. **This partly revises what I said
above** — the claim that this is "not a performance change" holds for the synthetic probe, but not
for LLM decode.

[GPULlama3.java](https://github.com/beehive-lab/GPULlama3.java) `--bench` (llama-bench style),
Llama-3.2-3B-Instruct-Q8_0, `-p 512 -n 128 -r 5`. **3 full invocations per configuration**, the
same application jar on both SDKs (this PR does not touch `tornado-api`, so one build runs against
either).

| configuration | JNI (develop) | FFM (this branch) | delta |
|---|---|---|---|
| **pp512** prefill, default | 63.46 t/s `[63.35–63.54]` | 64.62 t/s `[64.56–64.65]` | **+1.8%** |
| **tg128** decode, default | 105.09 t/s `[104.92–105.22]` | 108.45 t/s `[108.33–108.57]` | **+3.2%** |
| pp512, `--cuda-graphs` | 73.18 t/s | 73.22 t/s `[73.12–73.30]` | +0.1% |
| tg128, `--cuda-graphs` | 134.97 t/s | 135.08 t/s `[134.97–135.14]` | +0.1% |

The default-path ranges do not overlap across three invocations per side, so the +3.2% on decode is
reproducible rather than noise.

### Why this differs from the saxpy probe

The probe said parity; this says +3.2%. Both are correct, and the difference is instructive.

In the probe, each `execute()` is one task followed by a blocking round trip — JFR showed 82.9% of
host time inside `cuStreamSynchronize`. Host issue cost is *hidden behind the GPU wait*, so a
cheaper or dearer call path does not move the number.

LLM decode is the opposite shape: a task graph of many small kernels per token, issued back to back,
where the GPU finishes faster than the host can feed it. Per-call host cost sits **on** the critical
path, and a few hundred nanoseconds per call compounds across thousands of launches per second.

`--cuda-graphs` confirms the mechanism from the other direction: capturing the whole graph and
replaying it with one launch removes almost all per-dispatch host work, and the two builds become
identical (+0.1%). **The gain appears exactly where per-call host cost is exposed, and vanishes
exactly where it is eliminated.** That is what a real effect should look like.

### Caveat on attribution

The +3.2% should not all be credited to FFM-versus-JNI as mechanisms. This branch also removed
per-call work while porting the issue path — the NVTX labels are interned instead of being formatted
and encoded per transfer, and the single-word out-parameters use a reused per-thread scratch slot
instead of an arena per call. Those are things the JNI version did differently, and they are part of
the measured difference. Untangling them would mean back-porting those optimisations to the JNI path
to see what is left, which I have not done.

So: **the FFM issue path as written is measurably cheaper than the JNI one it replaces on
host-bound workloads**, which is the honest claim. Whether that is intrinsic to the binding
mechanism or to how the port was written, this measurement cannot say.

### Two incidental findings

- **External launchers do not get the native-access grant.** `llama-tornado` builds its own `java`
  command instead of calling `bin/tornado`, so it prints
  `WARNING: java.lang.foreign.SymbolLookup::libraryLookup has been called by module tornado.runtime`
  on every run. Harmless on JDK 21, but any third-party launcher will need
  `--enable-native-access=tornado.runtime` added, and that is a downstream change this PR should
  probably document.
- GPULlama3's pom pins `tornado-api 5.0.0-jdk21`; building it against this TornadoVM needs
  `-Dtornadovm.base.version=5.2.1 -Djdk.version.suffix=-jdk21-dev`, otherwise the lambdas are
  rejected at run time. Pre-existing and unrelated to this PR — noted for anyone reproducing.

<sub>RTX 4090, driver 565.57, CUDA 12.6, JDK 21, CUDA backend, same machine, builds alternated.</sub>

mikepapadim:
## Exhaustive JDK verification — 21 / 25 / 26 / 27, both backends, both SDK profiles

Ran the full supported-JDK matrix locally at `25417f181` because CI has **not** covered it on this
PR (see the next comment). Short version: **no JDK-specific divergence anywhere**, and the
native-access story is stronger than the description claims. I found nothing that blocks a merge.

### Unit suite — `tornado-test --quickPass`

Each SDK built once and then run unchanged on every JDK, which is the property the build-test
workflow is designed to prove.

| SDK | JDK | backend | ran | failed | unsupported |
|---|---|---|---|---|---|
| jdk21 | 21 | OpenCL | 1171 | 8 | 272 |
| jdk21 | 21 | CUDA | 1168 | 7 | 73 |
| jdk22plus | 25 | OpenCL | 1171 | 8 | 272 |
| jdk22plus | 26 | OpenCL | 1171 | 8 | 272 |
| jdk22plus | 27 | OpenCL | 1171 | 8 | 272 |
| jdk22plus | 25 | CUDA | 1168 | 7 | 73 |
| jdk22plus | 26 | CUDA | 1168 | 8 | 73 |
| jdk22plus | 27 | CUDA | 1168 | 7 | 73 |

The failing **set** is identical on every row, not just the count:
`TestDevices`, `ComputeTests`, `TestMatrixMultiplicationKernelContext`, `TestReductionsFloats`,
`TestConcurrentBackends`, plus `TestMultipleTasksMultipleDevices` on OpenCL only.

### The one row that differs is the flaky test, not a JDK 26 regression

JDK 26 / CUDA shows 8 rather than 7. The extra one is
`TransformerKernelsTest` — the class already documented in the description as flaky. I re-ran that
class alone, 5 times per JDK, same SDK, same box:

| | pass | fail |
|---|---|---|
| JDK 26 | 3 | 2 |
| JDK 25 | 4 | 1 |

So it is flaky on 25 as well and the single extra failure in the sweep is that flakiness landing on
one run. Not JDK-specific, and consistent with the 9/20-vs-13/20 figures already in the PR.

### The module funnel holds under `--illegal-native-access=deny`

The description says every restricted call funnels through `tornado.runtime`. Worth recording that
this now verifies both ways:

- **Statically** — `libraryLookup`, `downcallHandle`, `upcallStub` and `MemorySegment::reinterpret`
  appear in exactly one file, `tornado-runtime/.../ffm/FFMSupport.java`. Every apparent hit under
  `tornado-drivers/`, `tornado-api/` and the tests is a codegen comment about bit-reinterpretation,
  not an FFM call. `OpenCLAPI:240` goes through `FFMSupport.upcallStub` rather than `Linker`
  directly.
- **Empirically** — with `--illegal-native-access=deny` and only `--enable-native-access=tornado.runtime`,
  both backends still load and enumerate on JDK 25, 26 and 27 (exit 0, 2 drivers). `deny` is what a
  future JDK makes the default, so the grant as written is not just sufficient today, it is
  sufficient once the warning becomes an error.

The API-intersection claim checks out too: `allocateFrom` / `getString` / `allocateUtf8String`
occur only in `FFMSupport.java`, so nothing else in the tree can drift between the two profiles.
There is no use of `Linker.Option.critical` (JDK 22+ only) anywhere, which is what lets one source
tree compile under both.

### SDK/JDK contract

`etc/tornado.jdk` is written correctly by both profiles (`floor=21,preview=true` /
`floor=22,preview=false`), and the jdk21 SDK does refuse to run on 25/26/27 with a clear message
rather than failing obscurely.

<sub>RTX 4090, driver 565.57.01, CUDA 12.6.85, Ubuntu. OpenJDK 21.0.2 / 25.0.2 / 26.0.1 / 27-ea+24.
Backends built `cuda,opencl`. Every number above measured this session at `25417f181`.</sub>

mikepapadim:
## Three things worth settling before merge

From the same sweep. One is a CI gap, one is a stale artefact this PR makes wrong, one is a
pre-existing papercut that this PR's contract makes easier to trip over.

### 1. The full JDK sweep has not run on this PR

The PR carries only the `6.0.0` label. `build-test.yml` says in its own words that
`ci-full-jdk-sweep` is *"how to get full coverage on the run before a merge"*, and without it a
pull request gets only the two ends of the range. Actual coverage on the current green tick:

| | covered |
|---|---|
| jdk21 SDK | JDK 21 |
| jdk22plus SDK | JDK 25, JDK 27 |
| **not covered by the GPU suite** | **JDK 22, 23, 24, 26** |

`reflection-unit-tests` does fan out across 21-27, but that is the non-GPU subset. The description's
own results table is JDK 21 only. For a PR that changes how every native call is bound, that is
thinner than it looks at a glance.

My run above closes **JDK 26** (both backends, both suites clean, modulo the flaky test). **22, 23
and 24 remain untested by anything GPU-side** — I do not have those JDKs on this box. Adding the
label and letting the nightly shape run is the cheap way to close them.

### 2. `example-argfile` is stale and this PR is what makes it wrong

`example-argfile` at the repo root was last touched by `f57ef4812` ("Add Maven wrapper and build
optimizations") and is untouched by this PR. It still carries `--enable-preview` and has **no**
`--enable-native-access=tornado.runtime` — the flag this PR introduces in `tornado.py`.

This is the in-repo instance of the downstream gap flagged for `llama-tornado`
[above](https://github.com/beehive-lab/TornadoVM/pull/1058#issuecomment-5461183211), and it is a
documented path, not a stray file: `BUILD_AND_RUN.md` presents `java @tornado-argfile` as
"Option 2", a first-class supported workflow for IDEs and CI.

Stripping the grant from `tornado --printJavaFlags` and running `VectorAddInt` reproduces what a
user of that file gets:

```
WARNING: A restricted method in java.lang.foreign.SymbolLookup has been called
WARNING: java.lang.foreign.SymbolLookup::libraryLookup has been called by module tornado.runtime
WARNING: Use --enable-native-access=tornado.runtime to avoid a warning for this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

Same on 21, 25, 26 and 27 — it warns and still runs correctly everywhere, so nothing is broken
today, but every one of those users is on the deprecation clock.

The good news is that the fix is free: `gen-tornado-argfile-template.py` is a faithful copy of
`--printJavaFlags`, and `--printJavaFlags` **does** emit the grant on both profiles. Regenerating
the committed `example-argfile` is enough.

Two smaller things in the same area, both pre-existing: `BUILD_AND_RUN.md` refers throughout to a
file called `tornado-argfile`, but what the repo ships is `example-argfile`; and the committed file
is jdk21-shaped, so on the jdk22plus profile its `--enable-preview` is wrong too.

### 3. An inline comment is inaccurate, and the contract refusal exits 0

`tornado.py:1326` says the missing grant means the lookup *"fails outright (JDK 21) or warns on
every run"*. On JDK 21 with this branch it **warns and runs correctly** — `VectorAddInt` prints
`Result is correct`. Worth correcting the comment so nobody relies on JDK 21 failing loudly.

Separately, and **pre-existing rather than introduced here**: both arms of
`checkCompatibilityWithTornadoVM` call `sys.exit(0)`. Running the jdk21 SDK on JDK 25 prints

```
This TornadoVM SDK was built for JDK 21 with preview features enabled, so it runs on JDK 21 only (found JDK 25).
```

and exits **0**. Any wrapper or CI step that checks `$?` reads a hard incompatibility as a
successful run that produced no output. This PR does not cause it, but it does make the
preview-pinned SDK a much more likely thing to hit, so it may be worth a one-character fix while
the area is open.

<sub>Same environment as the previous comment. Items 1 and 2 are verified; item 3's exit code was
read from the source and confirmed by running it.</sub>

mikepapadim:
> ## Three things worth settling before merge
> From the same sweep. One is a CI gap, one is a stale artefact this PR makes wrong, one is a pre-existing papercut that this PR's contract makes easier to trip over.
> 
> ### 1. The full JDK sweep has not run on this PR
> The PR carries only the `6.0.0` label. `build-test.yml` says in its own words that `ci-full-jdk-sweep` is _"how to get full coverage on the run before a merge"_, and without it a pull request gets only the two ends of the range. Actual coverage on the current green tick:
> 
> covered
> jdk21 SDK	JDK 21
> jdk22plus SDK	JDK 25, JDK 27
> **not covered by the GPU suite**	**JDK 22, 23, 24, 26**
> `reflection-unit-tests` does fan out across 21-27, but that is the non-GPU subset. The description's own results table is JDK 21 only. For a PR that changes how every native call is bound, that is thinner than it looks at a glance.
> 
> My run above closes **JDK 26** (both backends, both suites clean, modulo the flaky test). **22, 23 and 24 remain untested by anything GPU-side** — I do not have those JDKs on this box. Adding the label and letting the nightly shape run is the cheap way to close them.
> 
> ### 2. `example-argfile` is stale and this PR is what makes it wrong
> `example-argfile` at the repo root was last touched by `f57ef4812` ("Add Maven wrapper and build optimizations") and is untouched by this PR. It still carries `--enable-preview` and has **no** `--enable-native-access=tornado.runtime` — the flag this PR introduces in `tornado.py`.

@stratika 
> 
> This is the in-repo instance of the downstream gap flagged for `llama-tornado` [above](https://github.com/beehive-lab/TornadoVM/pull/1058#issuecomment-5461183211), and it is a documented path, not a stray file: `BUILD_AND_RUN.md` presents `java @tornado-argfile` as "Option 2", a first-class supported workflow for IDEs and CI.
> 
> Stripping the grant from `tornado --printJavaFlags` and running `VectorAddInt` reproduces what a user of that file gets:
> 
> ```
> WARNING: A restricted method in java.lang.foreign.SymbolLookup has been called
> WARNING: java.lang.foreign.SymbolLookup::libraryLookup has been called by module tornado.runtime
> WARNING: Use --enable-native-access=tornado.runtime to avoid a warning for this module
> WARNING: Restricted methods will be blocked in a future release unless native access is enabled
> ```
> 
> Same on 21, 25, 26 and 27 — it warns and still runs correctly everywhere, so nothing is broken today, but every one of those users is on the deprecation clock.
> 
> The good news is that the fix is free: `gen-tornado-argfile-template.py` is a faithful copy of `--printJavaFlags`, and `--printJavaFlags` **does** emit the grant on both profiles. Regenerating the committed `example-argfile` is enough.
> 
> Two smaller things in the same area, both pre-existing: `BUILD_AND_RUN.md` refers throughout to a file called `tornado-argfile`, but what the repo ships is `example-argfile`; and the committed file is jdk21-shaped, so on the jdk22plus profile its `--enable-preview` is wrong too.
> 
> ### 3. An inline comment is inaccurate, and the contract refusal exits 0
> `tornado.py:1326` says the missing grant means the lookup _"fails outright (JDK 21) or warns on every run"_. On JDK 21 with this branch it **warns and runs correctly** — `VectorAddInt` prints `Result is correct`. Worth correcting the comment so nobody relies on JDK 21 failing loudly.

@stratika 
> 
> Separately, and **pre-existing rather than introduced here**: both arms of `checkCompatibilityWithTornadoVM` call `sys.exit(0)`. Running the jdk21 SDK on JDK 25 prints
> 
> ```
> This TornadoVM SDK was built for JDK 21 with preview features enabled, so it runs on JDK 21 only (found JDK 25).
> ```
> 
> and exits **0**. Any wrapper or CI step that checks `$?` reads a hard incompatibility as a successful run that produced no output. This PR does not cause it, but it does make the preview-pinned SDK a much more likely thing to hit, so it may be worth a one-character fix while the area is open.
> 
> Same environment as the previous comment. Items 1 and 2 are verified; item 3's exit code was read from the source and confirmed by running it.

stratika (approved):
Seems to me ready. Thanks @mikepapadim

Issue #35, opened 2020-09-17
Add Javadoc and document the examples in TornadoVM


jjfumero:
Add Javadoc to document the examples in TornadoVM. 

This affects the packages under the `examples` module: 

https://github.com/beehive-lab/TornadoVM/tree/master/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples

The documentation is at the class-level and it will contain a description of how the TornadoVM API is used for each example. Additionally, it contains how to run the example from the command line.

This class contains an example of documentation of what we are looking for:

https://github.com/beehive-lab/TornadoVM/blob/master/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/polyglot/HelloPython.java#L26-L38 

#### Summary
- Javadoc at the class-level
- High-level description of the problem/example
- Description of the TornadoVM API calls used and how are they composed
- How to run from the command line.

luisa01-code:
Hi! I could try to do that, if you'd like 😄

stratika:
Hi @luisa01-code, that would be very nice. Feel free to open a PR, thank you.

aaronms1:
Here is an example I wrote using "hover over links" when the Jdoc is rendered on the ArrayAccInt example class.
Having the docs linked in this matter will have "pop-ups" when a user clicks, or hovers over them from any other class that mentions the original, pointing back at its Jdoc::


/**
 * {@link ArrayAccInt} is an example that demonstrates how to use the Tornado framework
 * to parallelize a computation on an array of integers.
 * <p>
 * How to run?
 * </p>
 * <code>
 *     tornado -m tornado.examples/uk.ac.manchester.tornado.examples.arrays.ArrayAccInt
 * </code>
 */
public class ArrayAccInt {

    /**
     * {@link ArrayAccInt#acc} that takes an array of integers and a value as input.
     * It uses the Tornado "@Parallel" annotation to indicate that the loop in the method should be parallelized.
     * The loop iterates over the array and adds the value to each element.
     * @param a int[]
     * @param value int
     */
    public static void acc(int[] a, int value) {
        for (@Parallel int i = 0; i < a.length; i++) {
            a[i] += value;
        }
    }

    /**
     * {@link ArrayAccInt#main(String[])} creates an array of integers and fills it with the value 10.
     * It then creates a TaskGraph object, which represents a computation that can be executed in parallel.
     * The array is transferred to the device (e.g., GPU) using the "transferToDevice" method.
     * <p>
[...]

PR #1060, opened 2026-09-01
docs: add class-level Javadoc to examples #35

AzazelSensei:
#### Description

Class-level Javadoc on the examples that only had a run command (or a copied/wrong blurb). Each one now has a short description of the example, how the TaskGraph / KernelContext calls are composed, and how to run it.

Left the longer, already-documented examples alone (MMA, GEMM benchmarks, VectorAddTest, etc.). Also fixed the KernelContext NBody comment that still talked about Monte Carlo, and the TestMemory comment that talked about matrix addition.

#### Problem description

#35 asked for class-level docs on the examples: what the example does, which API calls it uses, and the command line.

#### Backend/s tested

Docs only, no runtime change.

- [ ] OpenCL
- [ ] CUDA
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

Open any of the touched example classes and check the class Javadoc. No code path changed.

### DIFF
--- a/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/Init.java
+++ b/tornado-examples/src/main/java/uk/ac/manchester/tornado/examples/Init.java
@@ -31,13 +31,19 @@ import uk.ac.manchester.tornado.api.types.arrays.FloatArray;
 import uk.ac.manchester.tornado.examples.common.Messages;
 
 /**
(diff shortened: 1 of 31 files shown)


stratika (approved):
LGTM, thanks

PR #1068, opened 2026-09-04
[cuda] Rank NVRTC candidates by version so an older toolkit cannot shadow a newer one

mikepapadim:
#### Description

NVRTC selection took the bare `libnvrtc.so` soname first and consulted the CUDA
toolkit roots only afterwards. `FFMSupport.loadLibrary` returns the first
candidate that loads, so on any host where the loader resolves that soname to a
distribution-packaged NVRTC, that one wins regardless of what else is installed.

This patch ranks every discovered NVRTC by the CUDA version its soname carries
(resolving symlinks, falling back to the version named by the toolkit directory),
tries toolkit roots ahead of unversioned sonames, and sorts anything that
identifies no version last so it cannot masquerade as the newest toolkit. Roots
named by `CUDA_PATH` / `CUDA_HOME` / `CUDA_ROOT` keep their stated priority ahead
of merely-newer ones found on disk, since naming a toolkit is a deliberate act.

Candidates are now tried one at a time rather than in a single `loadLibrary`
call, so the winning path can be recorded and exposed through
`NVRTCAPI.loadedLibrary()`.

It also fixes a related diagnostic in `CUDACompiler`: an NVRTC compile that
succeeded but yielded a zero-length image was reported as a successful build.
That set `BUILD_SUCCESS` and deferred the problem to `cuModuleLoadDataEx`, which
sees the empty image and reports only `CUDA_ERROR_INVALID_PTX` — naming neither
the retrieval that failed nor the kernel it belonged to. It is now reported as a
compile failure that says which image format was requested.

#### Problem description

On a Blackwell host carrying Ubuntu's `nvidia-cuda-toolkit` package (CUDA 12.0)
beside a hand-installed CUDA 13.0, the backend pins itself to the 12.0 NVRTC.
That NVRTC cannot target `sm_120` at all, so `CUDACompiler` falls back to
`compute_90` PTX and prints "Upgrade the CUDA toolkit" — while a toolkit that
does support the GPU sits unused on disk, unnamed by anything in the log.

Two gaps made the ordering harder to notice:

* **On Linux only the `/usr/local/cuda` symlink was ever consulted**, so a
  versioned sibling such as `/usr/local/cuda-13.0` was undiscoverable — which is
  the usual shape of adding a newer toolkit to a machine whose distribution
  already packaged one. The Windows branch already enumerated and ranked its
  versioned directories; Linux now does the same.
* **The Debian/Ubuntu multiarch directories were never scanned**, so the packaged
  NVRTC was reachable only by soname and never version-ranked against anything.

To reproduce: on a machine with a distribution CUDA package installed and a newer
toolkit under `/usr/local/cuda-<newer>`, run any CUDA-backend task and observe
that the older NVRTC is loaded and the `sm_XX` fallback warning appears.

#### Backend/s tested

- [ ] OpenCL
- [x] CUDA
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
make tests
```

To exercise the selection specifically, on a host with two toolkits installed:

```bash
# Before this patch the distribution NVRTC wins; after it, the newer toolkit does.
tornado --printKernel --devices

# The chosen library is now retrievable programmatically:
#   NVRTCAPI.loadedLibrary()
```

Verify that the toolkit reported is the newest installed, and that pointing
`CUDA_PATH` at the *older* toolkit still selects that older one — an explicitly
named root must keep priority over a merely-newer one found on disk.

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/ffm/CUDACompiler.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/ffm/CUDACompiler.java
@@ -300,7 +300,19 @@ public final class CUDACompiler {
                     // caller, with the complete log.
                     return new NvrtcResult(false, null, log);
                 }
-                return new NvrtcResult(true, image(arena, program, useCubin), log);
+                // A compile that succeeded but yielded nothing to load is a failure here, not a
+                // success with an empty image: reporting it as success sets BUILD_SUCCESS and
+                // defers the problem to cuModuleLoadDataEx, which sees a zero-length image and
+                // reports only CUDA_ERROR_INVALID_PTX - naming neither the retrieval that failed
+                // nor the kernel it belonged to. Retrieval fails when the requested image format
+                // does not match what was compiled (asking for a cubin from a compute_XX virtual
+                // target, say), so say which format was asked for.
+                byte[] compiled = image(arena, program, useCubin);
+                if (compiled.length == 0) {
+                    String format = useCubin ? "cubin" : "PTX";
+                    return new NvrtcResult(false, null, "NVRTC reported success but produced no " + format + " image for this kernel." + (log.isEmpty() ? "" : "\n\nNVRTC log:\n" + log));
+                }
+                return new NvrtcResult(true, compiled, log);
             } finally {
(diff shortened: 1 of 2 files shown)


mikepapadim:
@stratika plus this

stratika (approved):

PR #1069, opened 2026-09-04
[cuda] Allow pinning NVRTC explicitly, and name the loaded toolkit in the fallback warning

mikepapadim:
> **Stacked on https://github.com/beehive-lab/TornadoVM/pull/1068.** That PR is the first commit in this
> branch; review it first, and this one becomes a single-commit diff once it
> merges.

#### Description

Two follow-ups to the NVRTC selection fix, both about being able to state which
NVRTC is in use rather than infer it from what happens to be installed.

**1. An explicit override.** Adds `TORNADO_NVRTC_LIBRARY` (environment variable)
and `-Dtornado.cuda.nvrtc.library` (system property), which name an NVRTC to load
ahead of every heuristic. On a host holding several toolkits the version ranking
otherwise decides which one is used; anything reporting a measurement has to be
able to pin that choice.

**2. A fallback warning that names the toolkit responsible.** The `sm_XX` fallback
warning previously ended with "Upgrade the CUDA toolkit to one supporting
sm_NNN". That is unactionable, and misleading, on exactly the host where it
matters most: one where a supporting toolkit is already installed and simply was
not the one loaded. It now names the NVRTC that imposed the cap, using
`NVRTCAPI.loadedLibrary()`, and points at the two ways to change the selection.

```
before: ... Upgrade the CUDA toolkit to one supporting sm_120 for native codegen.

after:  ... The NVRTC in use is CUDA 12.0 from /usr/lib/x86_64-linux-gnu/libnvrtc.so.
        Install, or point CUDA_PATH / TORNADO_NVRTC_LIBRARY at, a CUDA toolkit
        supporting sm_120 for native codegen.
```

#### Problem description

Not a bug fix on its own — it closes the diagnostic gap left by the selection fix
it builds on. With ranking in place the *right* toolkit is normally chosen, but
when it is not, the warning still gave the reader nothing to act on: no way to
see which NVRTC was loaded, and no way to override the choice.

#### Backend/s tested

- [ ] OpenCL
- [x] CUDA
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
make tests
```

The override, on a host with more than one toolkit:

```bash
# Force the older NVRTC and confirm the warning names it by path and version.
TORNADO_NVRTC_LIBRARY=/usr/lib/x86_64-linux-gnu/libnvrtc.so tornado --printKernel ...

# The system property takes the same value and the same precedence.
tornado -Dtornado.cuda.nvrtc.library=/usr/local/cuda-13.0/lib64/libnvrtc.so ...
```

Expect the named library to be loaded even when a newer toolkit is present, and
the `sm_XX` warning — when it appears — to name that library rather than telling
you to upgrade a toolkit you already have.

### DIFF
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/ffm/CUDACompiler.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/ffm/CUDACompiler.java
@@ -300,7 +300,19 @@ public final class CUDACompiler {
                     // caller, with the complete log.
                     return new NvrtcResult(false, null, log);
                 }
-                return new NvrtcResult(true, image(arena, program, useCubin), log);
+                // A compile that succeeded but yielded nothing to load is a failure here, not a
+                // success with an empty image: reporting it as success sets BUILD_SUCCESS and
+                // defers the problem to cuModuleLoadDataEx, which sees a zero-length image and
+                // reports only CUDA_ERROR_INVALID_PTX - naming neither the retrieval that failed
+                // nor the kernel it belonged to. Retrieval fails when the requested image format
+                // does not match what was compiled (asking for a cubin from a compute_XX virtual
+                // target, say), so say which format was asked for.
+                byte[] compiled = image(arena, program, useCubin);
+                if (compiled.length == 0) {
+                    String format = useCubin ? "cubin" : "PTX";
+                    return new NvrtcResult(false, null, "NVRTC reported success but produced no " + format + " image for this kernel." + (log.isEmpty() ? "" : "\n\nNVRTC log:\n" + log));
+                }
+                return new NvrtcResult(true, compiled, log);
             } finally {
                 MemorySegment handle = FFMSupport.allocatePointer(arena);
                 handle.set(FFMSupport.C_POINTER, 0, MemorySegment.ofAddress(program));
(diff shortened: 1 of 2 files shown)


mikepapadim:
Rebased onto `develop` (`66a1da7e`) and re-tested. #1068, which this was stacked on, has merged, so the banner is gone and this is now a single-commit change: **+37/-2 across two files**, plus a docs line.

### Re-tested on a host with exactly the shape this is for

Distribution CUDA 12.0 at `/usr/lib/x86_64-linux-gnu/libnvrtc.so`, a hand-installed CUDA 13.0 under `/usr/local/cuda-13.0`, and an RTX 5070 Ti (sm_120) that only the latter supports.

**Default** — the ranking from #1068 picks 13.0, no warning, native sm_120 codegen.

**Pinned to the old one**, by property or by environment variable:

```text
[TornadoVM-CUDA] WARNING: CUDA toolkit (NVRTC max sm_90) predates this GPU (sm_120). Using compute_90
PTX JIT'd by the driver - functional but not optimal (...). The NVRTC in use is CUDA 12.0 from
/usr/lib/x86_64-linux-gnu/libnvrtc.so. Install, or point CUDA_PATH / TORNADO_NVRTC_LIBRARY at, a CUDA
toolkit supporting sm_120 for native codegen.
```

The kernel still produces the right answer through driver JIT, so the fallback path itself is intact.

**A bogus path** (`-Dtornado.cuda.nvrtc.library=/nonexistent/libnvrtc.so`) falls back to the normal search and the run completes — it cannot brick a machine by typo.

One thing worth knowing when testing this by hand: the warning only appears on a **real compile**. With the on-disk module cache warm, the cached cubin is loaded and NVRTC is never consulted, so pinning an old NVRTC looks like it did nothing. Run an uncached kernel to see it.

### One change from the reassessment

`docs/source/flags.rst` lists every other `tornado.cuda.*` property and was missing this one. Added, including that `TORNADO_NVRTC_LIBRARY` is the environment form, that the property wins when both are set, and that an unloadable path falls back rather than failing.

Issue #1071, opened 2026-09-04
[cuda][profiler] Dispatch times are structurally 0 and copy-out time is never recorded, failing three non-whitelisted TestProfiler tests


mikepapadim:
## Summary

On the CUDA backend three `TestProfiler` tests fail, and they fail for two
independent reasons that both leave the profiler reporting `0` for quantities it
presents as measurements:

- `getDataTransferDispatchTime()` and `getKernelDispatchTime()` are **structurally
  always 0** — every absolute event timestamp on this backend comes from a stub.
- `getDeviceReadTime()` is 0 whenever a copy-out is issued without a dependency
  list, even though the transfer happened and was timed on the write side.

These are not in `__TORNADO_TESTS_WHITE_LIST__`; `TestProfiler` appears only in
`__TORNADO_HEAVY_TESTS__`, which excludes it from `--quickPass` rather than from
the expected-failure set.

```
Test: class uk.ac.manchester.tornado.unittests.profiler.TestProfiler
	Running test: testProfilerFromExecutionPlan ................  [FAILED]
		\_[REASON] null
	Running test: testProfilerEnabled        ................  [FAILED]
		\_[REASON] null
	Running test: testProfilerOnAndOff       ................  [FAILED]
		\_[REASON] null
Test ran: 9, Failed: 3, Unsupported: 0
```

The assertions carry no message, so the failure reports `null` and says nothing
about which counter was zero.

## Reproducer

Printing the values `testProfilerEnabled` asserts on, rather than asserting them
(full source at the end):

```
getTotalTime()                > 0  : 127504721
getTornadoCompilerTime()      > 0  : 44553386
getCompileTime()              > 0  : 108855706
getDataTransfersTime()       >= 0  : 30880
getDeviceWriteTime()         >= 0  : 30880
getDataTransferDispatchTime() > 0  : 0     <- fails
getKernelDispatchTime()       > 0  : 0     <- fails
getDeviceReadTime()           > 0  : 0     <- fails
```

Compile and JIT times are real, and `getDeviceWriteTime()` is a real 30880 ns, so
the profiler is working — these three counters specifically are not.

## Cause 1 — every absolute event timestamp is a stub

`CUDAEvent.getDriverDispatchTime()` is `getCLStartTime() - getCLQueuedTime()`.
Both reach `readEventTime` -> `clGetEventProfilingInfo`, and that method is:

```java
[...]

PR #1072, opened 2026-09-04
[test] Say which profiler counter was zero instead of failing with a null reason

mikepapadim:
#### Description

`TestProfiler`'s three profiled-execution tests (`testProfilerEnabled`,
`testProfilerFromExecutionPlan`, `testProfilerOnAndOff`) each wrote out the same
twelve assertions, none carrying a message. When one trips the runner prints:

```
	Running test: testProfilerEnabled        ................  [FAILED]
		\_[REASON] null
```

which identifies neither which assertion failed nor what it read. After this
patch, the same failure reports:

```
	Running test: testProfilerEnabled        ................  [FAILED]
		\_[REASON] getDataTransferDispatchTime() should be > 0, was 0
```

The three copies are folded into one `assertProfilerCountersPopulated` helper —
which is also what makes adding messages worth doing once instead of thirty-six
times.

#### Problem description

Not a product bug; a diagnosis one. It cost a bisect through the assertion block
to establish that the CUDA backend returns 0 from `getDataTransferDispatchTime()`,
which is the subject of #1071.

It matters more than a routine "add assertion messages" cleanup because **these
counters do not fail as a group**. A backend can report perfectly good compile,
total and device-write times while one dispatch counter is stuck at zero — so
"one of twelve assertions in this test failed" is a much weaker clue than it
looks, and the reader has no way to narrow it without editing the test.

#### What changed

- One `assertProfilerCountersPopulated(TornadoProfilerResult)` helper replaces
  three identical inline blocks.
- Failures name the counter and the observed value.
- Two assertions in each copy were exact duplicates of an earlier line in the same
  block — `getDeviceWriteTime() >= 0` appeared twice, and `getDeviceReadTime() >= 0`
  sat alongside the stronger `> 0` that follows it. The helper keeps the stronger
  check of each pair, so the set of conditions enforced is unchanged.

Net −43/+38 lines in one test file. No production code is touched.

#### Backend/s tested

- [ ] OpenCL
- [x] CUDA
- [ ] Metal

Backend-independent — a test-only change.

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
tornado-test -V uk.ac.manchester.tornado.unittests.profiler.TestProfiler
```

**The same three tests fail and the same six pass before and after** — verified on
a CUDA build against RTX 5070 Ti, CUDA 13.0.88, JDK 25.0.2. The only difference is
that the reason is now stated rather than `null`. On a backend where all twelve
hold, all nine pass as before.

`mvnw checkstyle:check -pl tornado-unittests` passes.

### DIFF
--- a/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/profiler/TestProfiler.java
+++ b/tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/profiler/TestProfiler.java
@@ -51,6 +51,41 @@ import uk.ac.manchester.tornado.unittests.common.TornadoTestBase;
  */
 public class TestProfiler extends TornadoTestBase {
 
+    /**
+     * Asserts every counter a profiled execution is expected to populate, naming the counter and
+     * the value it actually held when one of them does not.
+     *
+     * <p>
+     * The same twelve assertions were written out at three call sites with no messages, so a zero
+     * counter failed as {@code [REASON] null} -- identifying neither which assertion tripped nor
+     * what it read. That matters here because the counters do not fail as a group: a backend can
+     * report real compile and write times while a dispatch counter is stuck at zero.
+     * </p>
+     */
+    private static void assertProfilerCountersPopulated(TornadoProfilerResult profiler) {
+        assertPositive("getTotalTime()", profiler.getTotalTime());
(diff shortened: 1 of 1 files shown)


mikepapadim:
@stratika also this one. It is just a test modification

stratika (approved):

PR #1074, opened 2026-09-07
[cusparse] Expose alpha and beta on SpMV and SpMM

mikepapadim:
## What

`cusparseSpMV` and `cusparseSpMM` compute `y = α·A·x + β·y` and `C = α·A·B + β·C`, but the factories took no scalars and the provider hard-coded `1.0f` and `0.0f` at both the buffer-size query and the call:

```java
// CusparseNativeLib already took them as parameters
static int spmv(long handle, ..., float alpha, float beta, long workspace)

// CusparseLibraryProvider passed literals
return CusparseNativeLib.spmv(context.handle, ..., 1.0f, 0.0f, context.workspacePtr);
```

So only the Java signature withheld them. This exposes both.

## Why it matters

`β = 0` makes the accumulating form unreachable. A caller wanting `y = A·x + y` had to add a separate JIT task to do the addition — an extra kernel launch and a full pass over `y` for something cuSPARSE does inside the same kernel.

## The part that is not plumbing

**With `β ≠ 0` the output is read before it is written**, so it is now declared `READ_WRITE` rather than `WRITE_ONLY`:

```java
access[9] = (beta != 0.0f) ? Access.READ_WRITE : Access.WRITE_ONLY; // y
```

Without this the accumulating form would read a buffer the data-flow graph has not kept resident, and return a **wrong result rather than an error**.

This is the same inference `tornado-cublas` already makes in `readOnlyExcept(numArgs, outputIndex, beta)` for `gemv` and `gemm`, so the two providers now derive the access class the same way.

## Compatibility

The existing short forms are unchanged and delegate:

```java
public static LibraryTaskDescriptor cusparseSpMV(int rows, int cols, int nnz,
        IntArray csrRowOffsets, IntArray csrColInd, FloatArray csrValues, FloatArray x, FloatArray y) {
    return cusparseSpMV(rows, cols, nnz, 1.0f, csrRowOffsets, csrColInd, csrValues, x, 0.0f, y);
}
```

Delegation rather than a second descriptor is deliberate: the provider dispatches on function name alone, so two parameter layouts behind one name would not be distinguishable. This keeps exactly one layout per function.

Scalar placement follows the BLAS convention already used by `cublasSgemv` — `alpha` before the matrix, `beta` before the output.

## Tests

Five added to `TestCusparse`:

| Test | Covers |
|---|---|
| `testSpMVAlphaScaling` | α ≠ 1 with β = 0 |
| `testSpMVAlphaBeta` | α and β ≠ 0 against a CPU reference |
| `testSpMVBetaAccumulates` | β = 1 summing across 4 executions — **fails if the output buffer is not live between calls** |
| `testSpMVShortFormMatchesLongForm` | guards the delegation |
| `testSpMMAlphaBeta` | the SpMM counterpart |

## Verification

```
tornado-test -V uk.ac.manchester.tornado.unittests.cusparse.TestCusparse
Test ran: 16, Failed: 0, Unsupported: 0
```

16 = the 11 pre-existing tests plus the 5 new ones. `./mvnw checkstyle:check` clean.

## Not in this PR

The other fixed configuration in this provider — CSR-only, FP32-only, `NON_TRANSPOSE`, default algorithm — is left alone. Those need real API surface decisions; this one only needed the signature widened.

### DIFF
--- a/tornado-cusparse/src/main/java/uk/ac/manchester/tornado/cusparse/Cusparse.java
+++ b/tornado-cusparse/src/main/java/uk/ac/manchester/tornado/cusparse/Cusparse.java
@@ -37,6 +37,15 @@ import uk.ac.manchester.tornado.api.types.arrays.IntArray;
  * zero-based): {@code csrRowOffsets} has {@code rows + 1} entries,
  * {@code csrColInd} and {@code csrValues} have {@code nnz} entries. Dense
  * operands are row-major, matching the TornadoVM native array layout. All FP32.
+ *
+ * <p>
+ * Each operation has two forms. The short form computes the plain product
+ * ({@code alpha = 1}, {@code beta = 0}); the long form takes {@code alpha} and
+ * {@code beta} explicitly and computes the accumulating form
+ * {@code y = alpha * A * x + beta * y}. When {@code beta != 0} the output
+ * operand is read before it is written, so it is declared
+ * {@link Access#READ_WRITE} rather than {@link Access#WRITE_ONLY} and its
+ * device contents are kept live by the data-flow graph.
  */
 public final class Cusparse {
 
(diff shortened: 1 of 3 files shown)


kotselidis (approved):

mikepapadim:
@stratika  can you rebase with develop to pass ci then wen can merge

stratika (approved):

PR #1075, opened 2026-09-07
[cublaslt] Add factories for the RELU, GELU and RELU_BIAS epilogues

mikepapadim:
## What

`CuBlasLtEpilogue` declares six epilogues, but only three are reachable:

| Epilogue | Value | Factory before this PR |
|---|---|---|
| `DEFAULT` | 1 | ✅ `ltMatmulFP32` / `FP16` / `FP8` |
| **`RELU`** | 2 | ❌ none |
| `BIAS` | 4 | ✅ `ltMatmulBiasFP16` |
| **`RELU_BIAS`** | 6 | ❌ none |
| **`GELU`** | 32 | ❌ none |
| `GELU_BIAS` | 36 | ✅ `ltMatmulGeluBiasFP16` |

Three values could be named but not executed. This adds the missing factories.

## Why it is small

The provider is already generic. Its registry maps a function name to a configuration record:

```java
private record LtCall(CudaDataType inputType, CudaDataType outputType, CuBlasLtEpilogue epilogue, boolean hasBias) {}
```

and the dispatch path passes `call.epilogue().value()` straight to `cublasLtMatmulDescSetAttribute`. So reaching the remaining three needs **no new machinery** — three factories and three registry entries.

## Added

```java
ltMatmulReluFP16      // C = ReLU(alpha*op(A)*op(B) + beta*C)
ltMatmulGeluFP16      // C = GELU(alpha*op(A)*op(B) + beta*C)
ltMatmulReluBiasFP16  // C = ReLU(alpha*op(A)*op(B) + beta*C + bias)
```

The bias-free forms take 13 arguments like `ltMatmulFP16`; the bias form takes 14 like `ltMatmulBiasFP16`, with the bias broadcast over the rows of the column-major result. All three derive the output access class through the existing `readOnlyExcept(numArgs, outputIndex, beta)` helper, so `β ≠ 0` marks the output `READ_WRITE` exactly as the other Lt factories do.

## Why these three are worth having

ReLU without a bias is the activation of a plain CNN or MLP block. GELU without a bias appears wherever the bias has already been folded into a preceding operation. Both previously needed a separate JIT task over the whole result to do work the epilogue performs inside the same kernel.

## Tests

Three added to `TestCuBlasLt`, each validated against a CPU reference at the same tolerance as the existing epilogue tests:

- `testLtMatmulReluFP16`
- `testLtMatmulGeluFP16`
- `testLtMatmulReluBiasFP16`

The reference helper is generalised from a `boolean gelu` flag to an `Activation {NONE, RELU, GELU}` enum. The previous boolean signature is kept and delegates, so **the existing tests are unchanged**.

## Verification

```
tornado-test -V uk.ac.manchester.tornado.unittests.cublas.TestCuBlasLt
Test ran: 9, Failed: 0, Unsupported: 0
```

9 = the 6 pre-existing tests plus the 3 new ones. `./mvnw checkstyle:check` clean.

## Not in this PR

The other unreached parts of the cuBLASLt surface — the backward epilogues (`DRELU`, `DGELU`, `BGRADA`/`BGRADB`), `GELU_AUX`, batching via the layout attributes, and FP8 scale pointers — are left alone. Those need enum values and API decisions this PR does not make; these three were already declared and only lacked a factory.

### DIFF
--- a/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlasLt.java
+++ b/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlasLt.java
@@ -128,4 +128,54 @@ public final class CuBlasLt {
                 .withParameters(new Object[] { transa, transb, m, n, k, alpha, matrixA, lda, matrixB, ldb, beta, matrixC, ldc, bias }) //
                 .withAccess(readOnlyExcept(14, 11, beta));
     }
+
+    /**
+     * FP16 matmul with a fused ReLU: C = ReLU(alpha * op(A) * op(B) + beta * C).
+     *
+     * <p>The activation is applied by the cuBLASLt epilogue, so no separate pass over C is
+     * needed. Use {@link #ltMatmulReluBiasFP16} when a bias is also wanted.</p>
+     */
+    public static LibraryTaskDescriptor ltMatmulReluFP16(int transa, int transb, int m, int n, int k, //
+            float alpha, HalfFloatArray matrixA, int lda, HalfFloatArray matrixB, int ldb, //
+            float beta, HalfFloatArray matrixC, int ldc) {
+        return new LibraryTaskDescriptor() //
+                .withLibrary(LIBRARY_NAME) //
+                .withFunction("ltMatmulReluFP16") //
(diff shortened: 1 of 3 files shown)


kotselidis (approved):

mikepapadim:
@stratika can you rebase with develop to pass ci then wen can merge

stratika (approved):

PR #1076, opened 2026-09-07
[cublas] Bind cublasDgemm for FP64 matrix multiply

mikepapadim:
## What

cuBLAS had **no double-precision path**. The scaffolding was already in place —
`CudaDataType.CUDA_R_64F`, `CublasComputeType.CUBLAS_COMPUTE_64F`, and the `DoubleArray` native array type all exist — but `cublasDgemm_v2` was not among the bound symbols, so an FP64 GEMM fell back to a JIT kernel.

This binds it and adds `CuBlas.cublasDgemm`.

## Why it is a small change

The FFM descriptor is **identical** to `cublasSgemm_v2`:

```java
FunctionDescriptor.of(C_INT, C_LONG, C_INT, C_INT, C_INT, C_INT, C_INT,
                      C_POINTER, C_LONG, C_INT, C_LONG, C_INT, C_POINTER, C_LONG, C_INT)
```

α and β cross as `C_POINTER` into host scratch, and the matrices as `C_LONG` device pointers, so element width does not change the downcall's shape. Only the marshalling differs — a `double` overload of the per-thread scalar scratch writes them with `C_DOUBLE`.

## Two details that are silent if wrong

**1. The access-class helper.** `readOnlyExcept(numArgs, outputIndex, beta)` took a `float`. Passing a `double` does not merely fail to compile — narrowing first means a small non-zero β such as `1e-300` **rounds to `0.0f`**, and the output would be marked `WRITE_ONLY` when cuBLAS is about to read it. This adds a `double` overload comparing against `0.0d`:

```java
accesses[outputIndex] = (beta != 0.0d) ? Access.READ_WRITE : Access.WRITE_ONLY;
```

**2. Dispatch registration.** The cuBLAS provider dispatches through a `Map<String, CuBlasCall>` of method references, not a `switch`. A marshalling method that exists but is not registered in `FUNCTIONS` **compiles without error** and fails only at runtime with "function not supported".

## Tests

Three added to `TestCuBlas`:

| Test | Covers |
|---|---|
| `testDgemm` | FP64 GEMM against a double-precision CPU reference |
| `testDgemmBetaReadsC` | β ≠ 0 — the case that fails if the output is `WRITE_ONLY` |
| `testDgemmIsTrulyDoublePrecision` | `1 + 2⁻³⁰`, exact in FP64 and rounding to exactly `1.0f` in FP32, asserted at **zero tolerance** |

The third exists because a plain correctness test would not catch an accidental single-precision path — the values would agree to within any reasonable tolerance. This one cannot pass in FP32.

## Verification

```
tornado-test -V uk.ac.manchester.tornado.unittests.cublas.TestCuBlas
Test ran: 17, Failed: 0, Unsupported: 0
```

17 = the 14 pre-existing tests plus the 3 new ones. `./mvnw checkstyle:check` clean.

## Scope

Only `dgemm`. `dgemv`, the batched FP64 forms and the complex types remain unbound — each needs its own symbol and marshalling, and they are better reviewed separately than as one large change.

### DIFF
--- a/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlas.java
+++ b/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlas.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 
 import uk.ac.manchester.tornado.api.common.Access;
 import uk.ac.manchester.tornado.api.common.LibraryTaskDescriptor;
+import uk.ac.manchester.tornado.api.types.arrays.DoubleArray;
 import uk.ac.manchester.tornado.api.types.arrays.FloatArray;
 import uk.ac.manchester.tornado.api.types.arrays.BFloat16Array;
 import uk.ac.manchester.tornado.api.types.arrays.HalfFloatArray;
@@ -59,6 +60,18 @@ public final class CuBlas {
         Access[] accesses = new Access[numArgs];
         Arrays.fill(accesses, Access.READ_ONLY);
         accesses[outputIndex] = (beta != 0.0f) ? Access.READ_WRITE : Access.WRITE_ONLY;
+        return accesses;
+    }
+
+    /**
+     * Double-precision counterpart of {@link #readOnlyExcept(int, int, float)}. Comparing the
+     * double directly matters: narrowing to float first would round a small non-zero beta to
+     * 0.0f and mark an operand WRITE_ONLY that cuBLAS is going to read.
+     */
+    private static Access[] readOnlyExcept(int numArgs, int outputIndex, double beta) {
+        Access[] accesses = new Access[numArgs];
+        Arrays.fill(accesses, Access.READ_ONLY);
+        accesses[outputIndex] = (beta != 0.0d) ? Access.READ_WRITE : Access.WRITE_ONLY;
         return accesses;
     }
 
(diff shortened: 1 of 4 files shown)


kotselidis (approved):

mikepapadim:
## Note on merging this alongside #1077 — either order is now safe

These two PRs both grow `CuBlasLibraryProvider.FUNCTIONS`, and until now they were in **different forms**, which made the merge conflict a trap rather than a routine resolution.

`Map.of` accepts at most **10 key/value pairs**. #1077 takes the registry to 11, so it converted to `Map.ofEntries`. This PR stayed on `Map.of` at 7 entries, which is fine on its own.

The problem was the combination. A `Map.of` block against a `Map.ofEntries` block conflicts as two different-looking chunks, and resolving it by taking either side **drops the other side's entries**. A dropped entry is not a compile error — the factory still exists and the marshalling method still exists, so it builds clean and fails at runtime with `cuBLAS function not supported: cublasDgemm`.

I verified git does conflict here rather than auto-merging, so this always needed a human to get it wrong. But the resolution was easy to get wrong.

**This PR now uses `Map.ofEntries` too.** The conflict becomes both entry lists in identical syntax:

```java
<<<<<<< HEAD
    Map.entry("cublasSgemm", CuBlasLibraryProvider::sgemm), //
    Map.entry("cublasDgemm", CuBlasLibraryProvider::dgemm), //
=======
    Map.entry("cublasSdot", CuBlasLibraryProvider::sdot), //
    Map.entry("cublasSnrm2", CuBlasLibraryProvider::snrm2), //
    Map.entry("cublasSasum", CuBlasLibraryProvider::sasum), //
    Map.entry("cublasIsamax", CuBlasLibraryProvider::isamax), //
    Map.entry("cublasIsamin", CuBlasLibraryProvider::isamin), //
    Map.entry("cublasSgemm", CuBlasLibraryProvider::sgemm), //
>>>>>>> feat/cublas-pointer-mode
```

The correct resolution is visibly the union, and `cublasDgemm` cannot be lost by accident. `Map.ofEntries` also has no pair limit, so whoever adds the next routine will not meet the ceiling — worth having, since the error when you do meet it never mentions it:

```
no suitable method found for of(String,CuBlasLibr[...])
```

Re-verified on the GPU after the conversion, since it touches the dispatch registry: `tornado-test -V TestCuBlas` → **17 ran, 0 failed, 0 unsupported**. Checkstyle clean.

PR #1077, opened 2026-09-08
[cublas] Bind the scalar-output level-1 routines using device pointer mode

mikepapadim:
## What

cuBLAS **level-1 coverage was zero** of roughly thirteen families. This binds the three scalar-output routines — `cublasSdot`, `cublasSnrm2`, `cublasSasum`.

## Why they could not be bound before

Not the bindings — the **pointer mode**. `cublasSetPointerMode` is never called on `develop` (zero references), so the handle stays in the default `CUBLAS_POINTER_MODE_HOST`.

In host mode these routines write their result through a *host* pointer:

```c
cublasSdot_v2(handle, n, x, 1, y, 1, &result);   // host memory
```

cuBLAS therefore has to **synchronise the stream** to deliver the value. That stalls the pipeline, and it is **illegal inside a CUDA Graph capture region**. Binding them in host mode would have produced entry points that break the two things the hybrid API exists to provide.

Under `CUBLAS_POINTER_MODE_DEVICE` the result pointer is device memory: the value stays on the GPU, no synchronisation happens, and a following task consumes it directly.

## API

The result is a single-element `FloatArray` — already a device-resident buffer the data-flow graph understands, declared `WRITE_ONLY` like any other output:

```java
FloatArray norm = new FloatArray(1);

new TaskGraph("g")
    .transferToDevice(DataTransferMode.EVERY_EXECUTION, x, data)
    .libraryTask("nrm2", CuBlas::cublasSnrm2, n, x, 1, norm)
    .task("scale", MyKernels::scaleByDeviceScalar, data, norm)  // consumes it on the device
    .transferToHost(DataTransferMode.EVERY_EXECUTION, data);
```

## Scoping the mode

The pointer mode is applied **only to the calls that need it**, not exposed as a user-facing knob:

```java
private static int inDevicePointerMode(long handle, LongToIntFunction body) {
    int status = setPointerMode(handle, CUBLAS_POINTER_MODE_DEVICE);
    if (status != CUBLAS_STATUS_SUCCESS) {
        return status;
    }
    try {
        return body.applyAsInt(handle);
    } finally {
        setPointerMode(handle, CUBLAS_POINTER_MODE_HOST);
    }
}
```

The restore matters: the handle is shared by every call of the execution plan, and the existing `gemv`/`gemm` factories pass α and β as host floats. This mirrors how `cublasSetMathMode` is already set and restored. A caller never has to know the mode exists.

## Tests

Five added to `TestCuBlas`. Three check values against CPU references. The other two cover what host pointer mode **cannot express at all**:

| Test | Why it matters |
|---|---|
| `testNrm2ResultFeedsJitTaskOnDevice` | a JIT kernel consumes the norm with no host round trip |
| `testSdotUnderCudaGraph` | captures and replays 5× — impossible under host mode, where the synchronisation is not permitted during capture |

## Verification

RTX 4090 (sm_89), driver 565.57.01, CUDA 12.6.85, JDK 25.0.2, `make jdk22plus BACKEND=cuda`:

```
tornado-test -V uk.ac.manchester.tornado.unittests.cublas.TestCuBlas
Test ran: 19, Failed: 0, Unsupported: 0
```

19 = the 14 pre-existing tests plus the 5 new ones. `./mvnw checkstyle:check` clean.

## Not in this PR

Device pointer mode also enables **α and β to be computed on the device** — letting a preceding JIT task produce a scaling factor, and fixing the trap where host-mode scalars are baked into a captured CUDA Graph by value and silently reused on replay. That needs new `gemv`/`gemm` factory overloads taking `FloatArray` scalars, and is better reviewed separately from the level-1 bindings.

The index-returning routines (`amax`, `amin`) are also unbound; they return an `int` index and would need the same treatment with an `IntArray`.

### DIFF
--- a/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlas.java
+++ b/tornado-cublas/src/main/java/uk/ac/manchester/tornado/cublas/CuBlas.java
@@ -25,6 +25,7 @@ import uk.ac.manchester.tornado.api.types.arrays.DoubleArray;
 import uk.ac.manchester.tornado.api.types.arrays.FloatArray;
 import uk.ac.manchester.tornado.api.types.arrays.BFloat16Array;
 import uk.ac.manchester.tornado.api.types.arrays.HalfFloatArray;
+import uk.ac.manchester.tornado.api.types.arrays.IntArray;
 import uk.ac.manchester.tornado.cublas.enums.CuBlasMathMode;
 
 /**
@@ -159,6 +160,43 @@ public final class CuBlas {
         return new LibraryTaskDescriptor() //
                 .withLibrary(LIBRARY_NAME) //
                 .withFunction("cublasSasum") //
+                .withParameters(new Object[] { n, x, incx, result }) //
+                .withAccess(access);
+    }
+
+    /**
+     * Index of the element of {@code x} with the largest absolute value, written to a
+     * device-resident single-element {@link IntArray}.
+     *
+     * <p><b>The index is 1-based</b>, following the BLAS convention that cuBLAS preserves: for a
+     * vector whose largest element sits at Java index {@code i}, this writes {@code i + 1}.</p>
+     *
+     * <p>Like {@link #cublasSdot} the result stays on the device, so a following task can consume
+     * it and the call is legal inside a CUDA Graph.</p>
+     */
+    public static LibraryTaskDescriptor cublasIsamax(int n, FloatArray x, int incx, IntArray result) {
+        Access[] access = new Access[4];
+        Arrays.fill(access, Access.READ_ONLY);
+        access[3] = Access.WRITE_ONLY; // result
+        return new LibraryTaskDescriptor() //
+                .withLibrary(LIBRARY_NAME) //
+                .withFunction("cublasIsamax") //
+                .withParameters(new Object[] { n, x, incx, result }) //
+                .withAccess(access);
+    }
+
+    /**
+     * Index of the element of {@code x} with the smallest absolute value, written to a
+     * device-resident single-element {@link IntArray}. <b>1-based</b>, as for
+     * {@link #cublasIsamax}.
+     */
+    public static LibraryTaskDescriptor cublasIsamin(int n, FloatArray x, int incx, IntArray result) {
+        Access[] access = new Access[4];
+        Arrays.fill(access, Access.READ_ONLY);
+        access[3] = Access.WRITE_ONLY; // result
+        return new LibraryTaskDescriptor() //
+                .withLibrary(LIBRARY_NAME) //
+                .withFunction("cublasIsamin") //
                 .withParameters(new Object[] { n, x, incx, result }) //
                 .withAccess(access);
     }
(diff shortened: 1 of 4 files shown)


mikepapadim:
## Extended: `Isamax` / `Isamin`, completing the scalar-output family

The limitations note said the index-returning routines were still unbound and would need the same treatment with an `IntArray`. Added, since they are the same shape and the same problem — a single value written through a pointer, which in host mode forces the stream synchronisation that stalls the pipeline and is illegal during graph capture. They reuse the same `inDevicePointerMode` helper.

**The index is 1-based**, following the BLAS convention cuBLAS preserves. That is the easy thing to get wrong in a binding, so the tests plant a maximum and a minimum at known positions and assert the exact index, rather than checking against random data where any plausible answer would look correct.

Now 22 ran, 0 failed on an RTX 4090 (sm_89), CUDA 12.6.85, JDK 25.0.2.

### One thing worth knowing for future routines

The dispatch registry had to move from `Map.of` to `Map.ofEntries`. `Map.of` accepts at most ten key-value pairs, and these two entries made eleven. The failure is a compile error at the map declaration whose message does not mention the limit, so it is easy to misread:

```
error: no suitable method found for of(String,CuBlasLibr[...]sgemv,String,...)
```

Anything else that adds a routine to `FUNCTIONS` will hit the same wall — including #1076, whose `cublasDgemm` entry would have made the count exceed ten on its own once merged alongside this.

### Where cuBLAS coverage stands after this

| Section | cuBLAS families | bound | notes |
|---|---|---|---|
| **Level 1** | ~13 | **5** | `dot`, `nrm2`, `asum`, `amax`, `amin` — was **0** before this PR |
| Level 2 | ~23 | 1 | `gemv` only |
| Level 3 | ~15 | 2 | `gemm`, `gemmStridedBatched` |
| Extensions | ~19 | 1 | `gemmEx`, three typed factories |

The five bound here are exactly the scalar-output subset, which is the part that *needed* device pointer mode. The remaining eight level-1 families — `axpy`, `scal`, `copy`, `swap`, and the four `rot*` routines — write to a **vector**, not a scalar, so they do not need this mechanism at all and can be bound in host pointer mode like the existing `gemv` and `gemm`. `axpy` and `scal` in particular are the most-used operations in the section.

Those belong in a separate PR rather than here: different mechanism, no dependency on this one, and this PR's title is accurate as it stands.

kotselidis (approved):

mikepapadim:
## Note on merging alongside #1076 — resolved from the other side

#1076 (`cublasDgemm`) also grows `CuBlasLibraryProvider.FUNCTIONS`, so these two conflict in the registry whichever merges first.

That conflict used to be a trap. This PR converted the registry to `Map.ofEntries` because it takes the entry count past `Map.of`'s ten-pair ceiling; #1076 was still on `Map.of`. A `Map.of` block against a `Map.ofEntries` block conflicts as two dissimilar chunks, and resolving it by taking either side **silently drops the other side's entries** — which compiles clean and only fails at runtime with `cuBLAS function not supported`.

**#1076 has now been converted to `Map.ofEntries` as well**, so both sides of the conflict are in identical syntax and the correct resolution is visibly the union of the two entry lists. Nothing is needed on this PR; this note is just so the interaction is recorded on both.

Either merge order is fine. The second one to land will show a conflict in `FUNCTIONS` where both entry lists are plainly visible — keep all entries from both sides.

PR #1079, opened 2026-09-08
Batch the global loads of a shared-memory staging sequence in the CUDA backend

### LINKED ISSUE #1078: [perf] KernelContext shared-memory tiled kernels stall on global memory that nvcc hides
## Summary

A hand-tiled GEMM written with `KernelContext` — stage tiles into shared memory, barrier, accumulate in registers — runs **2.3× slower** than the identical algorithm written directly in CUDA, while executing **fewer** instructions, with **no** register spilling, **fewer** bank conflicts and **higher** occupancy.

The cause is a **10.7× higher global-memory dependency stall**: the generated code does not let the next tile's global loads overlap with the current tile's arithmetic.

RTX 4090 (sm_89), driver 565.57.01, CUDA 12.6.85, JDK 25.0.2, TornadoVM 6.0.0-jdk22plus-cuda, Nsight Compute 2024.3.2. `n = 1024`, block 256 on both sides, single profiled launch.

## It is not any of the usual causes

| metric | TornadoVM | hand-written CUDA | |
|---|---|---|---|
| local memory ld/st sectors | **0** | **0** | nothing spills, either side |
| registers per thread | 72 | 70 | near-identical pressure |
| instructions executed | **43,714,560** | 44,615,680 | TornadoVM issues **fewer** |
| FFMA thread-instructions | **1,073,741,824** | **1,073,741,824** | *exactly* identical arithmetic |
| shared-memory bank conflicts | **602,365** | 622,879 | TornadoVM has **fewer** |
| LSU instructions | **6,332,416** | 6,461,440 | TornadoVM has **fewer** |
| block size | 256 | 256 | held constant |
| achieved occupancy | **32.94%** | 32.08% | TornadoVM is **higher** |
| **kernel time** | **192,608 ns** | **83,840 ns** | **2.3× slower** |

Identical arithmetic and less of everything else, yet slower — so the instructions are stalling rather than multiplying.

## Where it stalls

| warp stall reason (per issue active) | TornadoVM | CUDA |
|---|---|---|
| **`long_scoreboard`** (global memory) | **12.25** | **1.14** |
| `short_scoreboard` (shared memory) | 2.16 | 1.48 |
| `barrier` | 0.78 | 0.61 |
| `mio_throttle` | 0.48 | 0.74 |
| `wait` (fixed latency) | 0.35 | 0.40 |
| **`issue_active`, % of peak** | **20.61** | **51.49** |

`long_scoreboard` is **10.7× higher** and the issue rate is 40% of CUDA's. Every other category is comparable.

## Likely mechanism

The kernel loop is: stage a k-tile from global into shared, `localBarrier()`, accumulate, `localBarrier()`, repeat.

nvcc hoists the **next** iteration's global loads above the current iteration's arithmetic, so memory latency overlaps with FFMA work. The generated CUDA keeps them in program order, leaving each tile's loads exposed.

This is consistent with every number above, including the *lower* instruction count — no prefetch bookkeeping is emitted.

## Scope

**Not a general code-generation deficit.** In the same benchmark, the naive and plain-tiled rungs — identical algorithms on both sides — come out at **1.01–1.02×**, and the library rungs at 1.01× act as a control. It is specific to kernels that stage through shared memory in a loop, which is the shape every hand-tiled GEMM takes, and therefore the shape most likely to be written by someone reaching for `KernelContext` for performance.

## Reproducer

`demos/17-matmul-ladder/` in [mikepapadim/tornadovm-devoxx2026-cuda-demos](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos) — `MatMulLadder.java` and the equivalent `MatMulLadder.cu`, same algorithm, same launch geometry. Full counter data and the generated CUDA are in `results/raw/29-register-tiled-stalls/`.

```bash
tornado --classpath . MatMulLadder 1024 2
ncu --metrics smsp__average_warps_issue_stalled_long_scoreboard_per_issue_active.ratio,\
smsp__issue_active.avg.pct_of_peak_sustained_active \
  java @$TORNADOVM_HOME/tornado-argfile -cp . MatMulLadder 1024 2
```

## Filed as an issue rather than a PR

I have a diagnosis, not a fix. Making the loads overlap means either software pipelining in the CUDA backend or exposing `cp.async` double-buffering at the `KernelContext` level — both are design decisions about the compiler and the API, and a guess at either would be worse than reporting the measurement.

Two directions that seem worth considering, for whoever picks this up:

1. **`KernelContext` already exposes `asyncCopyToLocal` / `asyncCopyCommit` / `asyncCopyWaitGroup`** (`cp.async`), which is precisely the primitive for overlapping the next tile's load with the current tile's compute. A double-buffered variant of this kernel written with those calls would confirm the mechanism, and would be a useful thing to document even if the compiler never hoists automatically.
2. If the loop structure is what blocks hoisting, it may be visible in the emitted CUDA before ptxas — the generated source is attached in the reproducer.

Single host, sm_89 only. The 2.3× above is an `ncu` figure on a cold single launch; the steady-state `nsys` figure at n=2048 is 1.43×. Both are real and they are different measurement modes.

mikepapadim:
Fixes #1078.

## The problem

The CUDA backend emits a tile-staging sequence as an alternating chain — every global load is consumed by the shared-memory store on the very next line:

```c
ul_107  =  ul_0 + l_106;
f_108   =  *(( float *) ul_107);   // global load
adf_5[i_97]  =  f_108;             // shared store consumes it immediately
i_109   =  i_92 + i_102;           // only now is the next address computed
l_110   =  (long long) i_109;
l_112   =  (l_110 + 4L) << 2;
ul_113  =  ul_0 + l_112;
f_114   =  *(( float *) ul_113);   // second load cannot start until the first landed
adf_5[i_88]  =  f_114;
```

so one global load is in flight at a time. On the register-tiled sgemm of the CUDA ladder this shows up as a `long_scoreboard` stall of **12.10** cycles per issue against 1.45 for the equivalent hand-written CUDA, and an issue rate of **20.8%** against 50.9%, with identical FFMA counts, identical occupancy and zero spills on both sides.

## Why ptxas does not fix it

It cannot. The two schedules differ in the SASS:

```
tornadovm, before          L S L S L S L S L S L S L S L S     generic LD
tornadovm, after           L L L L L L L L S S S S S S S S     generic LD
hand-written CUDA          L L L L L L L L S S S S S S S S     LDG
```

Hand-written CUDA loads through `const float *`, which ptxas lowers to **`LDG`** — a load proven to target global memory, and therefore proven not to alias the `STS` beside it. ptxas then batches the loads on its own. TornadoVM casts an integer address to a plain pointer, which lowers to a **generic `LD`**. A generic load may target shared memory, so ptxas must keep it ordered against every shared store.

I checked the other direction too: in hand-written CUDA the interleaved schedule **cannot be written at all**. Three source orders of one kernel — interleaved, interleaved with an `asm volatile("" ::: "memory")` between pairs, and batched — compile to *byte-identical* SASS (448 instructions, `LDG`×8 then `STS`×8 in every case). ptxas picks the schedule, not the source.

TornadoVM knows the address space at LIR level (`CUDAMemorySpace`) even though its emitted C does not encode it. So the batching has to happen in the code generator, and that is what this PR does.

## The fix

`CUDAGlobalLoadBatching` is a peephole run from `CUDACompilationResultBuilder.emitBlock`. Within a straight-line **run** whose every instruction is one of

- a **non-local** `LoadStmt` (global or constant — `!isLocalOrPrivateLoad()`),
- an `AssignStmt` — pure register arithmetic; no CUDA LIR *operator* dereferences memory, only Load/Store *statements* do, which I verified by grepping every `emit("*(")` in the backend,
- a **local or private** `StoreStmt` (`isLocalOrPrivateStore()`),

the stores are sunk to the end of the run. Everything else keeps its original relative order, and the stores keep theirs.

I sink the stores rather than hoist the loads deliberately: hoisting a load means also hoisting the whole address-arithmetic chain that feeds it, while sinking a store moves exactly one instruction.

### Why it is sound

1. **No memory dependence is broken.** The only writes in a run go to local/shared memory and the only reads come from global/constant memory. Those address spaces are disjoint, so no load in the run can observe a store in the run.
2. **No shared read is skipped over.** A shared or private `LoadStmt` is *not* an eligible instruction, so a run can never contain a reader of the memory the sunk stores write.
3. **Store order is preserved**, so two writes to the same shared address still land in program order.
4. **True data dependences hold.** A store only ever moves later, so every value it reads is still defined before it. Non-store instructions do not move relative to each other at all.
5. **No value is clobbered.** `canSinkPast` checks explicitly that no instruction jumping ahead of a store defines a value that store reads. This is the one hazard the address-space argument does not cover — it would matter if the LIR ever reused a variable — so it is tested rather than assumed.
6. **Visibility is unchanged.** The run ends at the first instruction outside the eligible set, so barriers, calls, global stores, atomics and MMA statements all terminate it. The sunk stores still complete before the same `__syncthreads()` they did before.

If any single store fails the check, the whole run is left exactly as it was. The input LIR list is never mutated.

## Before / after `--printKernel`

Same build, `-Dtornado.cuda.batchGlobalLoads` toggled. Demo kernel is the ladder's `kcRegisterTiled`; this is the tail of BLOCK 2, the k-tile staging.

<details>
<summary><b>Before</b> (batching off) — load, store, load, store …</summary>

```c
  // BLOCK 2
  i_103  =  i_101 + i_102;
  l_104  =  (long long) i_103;
  l_105  =  l_104 + 4L;
  l_106  =  l_105 << 2;
  ul_107  =  ul_0 + l_106;
  f_108  =  *(( float *) ul_107);
  adf_5[i_97]  =  f_108;
  i_109  =  i_92 + i_102;
  l_110  =  (long long) i_109;
  l_111  =  l_110 + 4L;
  l_112  =  l_111 << 2;
  ul_113  =  ul_0 + l_112;
  f_114  =  *(( float *) ul_113);
  adf_5[i_88]  =  f_114;
  i_115  =  i_83 + i_102;
  ...
  f_120  =  *(( float *) ul_119);
  adf_5[i_79]  =  f_120;
  ...
  f_126  =  *(( float *) ul_125);
  adf_5[i_68]  =  f_126;
  ...
  f_134  =  *(( float *) ul_133);
  adf_6[i_61]  =  f_134;
  ...
  f_142  =  *(( float *) ul_141);
  adf_6[i_53]  =  f_142;
  ...
  f_150  =  *(( float *) ul_149);
  adf_6[i_44]  =  f_150;
  ...
  f_158  =  *(( float *) ul_157);
  adf_6[i_33]  =  f_158;
  __syncthreads();
```
</details>

<details>
<summary><b>After</b> (batching on) — all eight loads, then all eight stores</summary>

```c
  // BLOCK 2
  i_103  =  i_101 + i_102;
  l_104  =  (long long) i_103;
  l_105  =  l_104 + 4L;
  l_106  =  l_105 << 2;
  ul_107  =  ul_0 + l_106;
  f_108  =  *(( float *) ul_107);
  i_109  =  i_92 + i_102;
  ...
  f_114  =  *(( float *) ul_113);
  ...
  f_120  =  *(( float *) ul_119);
  ...
  f_126  =  *(( float *) ul_125);
  ...
  f_134  =  *(( float *) ul_133);
  ...
  f_142  =  *(( float *) ul_141);
  ...
  f_150  =  *(( float *) ul_149);
  i_151  =  i_34 + i_102;
  i_152  =  i_151 << 10;
  i_153  =  i_152 + i_37;
  l_154  =  (long long) i_153;
  l_155  =  l_154 + 4L;
  l_156  =  l_155 << 2;
  ul_157  =  ul_1 + l_156;
  f_158  =  *(( float *) ul_157);
  adf_5[i_97]  =  f_108;
  adf_5[i_88]  =  f_114;
  adf_5[i_79]  =  f_120;
  adf_5[i_68]  =  f_126;
  adf_6[i_61]  =  f_134;
  adf_6[i_53]  =  f_142;
  adf_6[i_44]  =  f_150;
  adf_6[i_33]  =  f_158;
  __syncthreads();
```
</details>

The full diff of the generated kernel is exactly those eight store lines moving, and nothing else.

## Numbers

RTX 4090 (sm_89), CUDA 12.6, n=1024, **100 launches per arm**, `nsys` per-launch duration. One build, flag toggled at run time, so the arms differ only in emitted instruction order.

| arm | median | mean | stddev | vs. before |
|---|---|---|---|---|
| before | 112.895 µs | 112.916 | 0.152 | 1.000x |
| after | 68.416 µs | 68.238 | 0.541 | **1.650x** |
| hand-written CUDA | 67.455 µs | 67.483 | 0.388 | 1.674x |

The gap to hand-written CUDA goes from **1.67x** to **1.4%**.

`ncu`, same kernel, same launch:

| counter | before | after |
|---|---|---|
| `long_scoreboard` stalls per issue | 12.10 | **1.45** |
| issue rate (% of peak sustained) | 20.82% | **50.90%** |
| FFMA instructions | 1,073,741,824 | 1,073,741,824 |
| registers per thread | 72 | 72 |
| LSU bank conflicts | 602,227 | 603,209 |

Same work, same occupancy, same registers, no spills — the kernel was simply waiting on memory.

## Testing

**Full device suite, run twice on the same build**, flag off and on, 1199 tests each:

- **zero regressions** — no test fails with the pass enabled that passes without it
- 11 failures are identical in both runs and pre-exist on `develop` (`TestDevices#test04/06`, `ComputeTests#testMandelbrot/testJuliaSets`, `TransformerKernelsTest#testReductionOneBlockWithLayer`, `TestMatrixMultiplicationKernelContext#mxm2DKernelContext01/02`, the three `TestProfiler` cases, `TestReductionsFloats#testComputePi`)
- two further tests failed only in the baseline run (`TestIO#testCopyInWithDevice`, `TestWarmUp#test01`); those look like flakes and I am **not** claiming the pass fixed them

**New no-GPU tests for the pass** (`make test-cuda-codegen`, plus a new CI lane — these need neither a device nor a CUDA driver and finish in under a second):

- `CUDAGlobalLoadBatchingTest` — 15 cases, mostly of the "must NOT reorder" kind: barrier, shared load, private load, global store, unrecognised instruction, clobber on the stored value, clobber on the store address, nested operand, store order, the two no-op cases, and input immutability
- `CUDAGlobalLoadBatchingInvariantsTest` — a property test over **4000 randomly generated instruction sequences**, drawn from a deliberately small variable pool so clobber candidates occur constantly. It checks five invariants: the instruction multiset is unchanged, both orders are preserved, nothing crosses an instruction the pass does not model, and no hoisted instruction clobbers a store operand. It also asserts the generator actually triggers the pass, so the invariants cannot pass vacuously.

**New device tests** — `TestSharedMemoryStaging`, seven kernels written so that a wrong schedule changes the *answer*, not the speed: a value staged before a barrier must be visible after it; a value read back from shared memory must be the one just written; two writes to one slot must land in order; a global store mid-sequence must keep its position; a gathered address must still be computed from its loaded index; and the tile-loop shape itself. All integer arithmetic with exact expected values, so a failure is a scheduling bug and never floating-point drift. Verified with `--printKernel` that the pass actually fires on these kernels — they are not vacuous. They pass with the flag both on and off, and on any backend.

Checkstyle clean.

## On the flag

`-Dtornado.cuda.batchGlobalLoads` defaults to **True**, and I think that is right: the safety predicate is an address-space disjointness argument rather than a heuristic, the whole device suite is clean, and the transformation is confined to the CUDA backend.

I would still keep the flag rather than hard-wire the behaviour. A code-generator reordering bug does not fail loudly — it produces a kernel that compiles, launches, and returns a slightly wrong number — so a maintainer who hits a shape we did not anticipate is much better served by a system property than by a revert. It costs one branch per basic block at compile time and nothing at run time. Happy to flip the default to False instead if you would rather it were opt-in for a release.

## Follow-up, not in this PR

The root cause is that TornadoVM emits a *generic* load. Emitting `__ldg()` or a global-qualified pointer would give ptxas the address space directly, fix the schedule at the source, and likely help beyond this pattern. That is a larger change to address lowering and belongs in its own PR.

Reproduction scripts and the raw per-launch CSVs are at https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/results/raw/31-load-batching-reorder


mikepapadim:
## 100-launch comparison against hand-written CUDA

Requested detail on the headline number. Kernel is `kcRegisterTiled` from the CUDA ladder demo — a 64x64 block-tiled sgemm with a 4x4 register micro-tile, staging both A and B tiles through shared memory. The hand-written CUDA arm is the *same algorithm, same tile sizes, same launch geometry*, compiled by `nvcc -arch=sm_89 -O3`; the ladder exists to make the two directly comparable.

RTX 4090, CUDA 12.6, n=1024, **100 launches per arm**, `nsys` per-launch kernel duration (not wall-clock, so no host dispatch is included). Both TornadoVM arms come from **one build** with `-Dtornado.cuda.batchGlobalLoads` toggled at run time, so nothing but the emitted instruction order differs between them.

| arm | launches | median | mean | stddev | min | p5 | p95 | max |
|---|---|---|---|---|---|---|---|---|
| TornadoVM, batching **off** | 100 | **112.895 µs** | 112.916 | 0.152 | 112.511 | 112.671 | 113.151 | 113.311 |
| TornadoVM, batching **on** | 100 | **68.416 µs** | 68.238 | 0.541 | 66.944 | 67.104 | 68.832 | 69.119 |
| hand-written CUDA | 103 | **67.455 µs** | 67.483 | 0.388 | 66.944 | 67.200 | 67.776 | 70.975 |

| ratio | value |
|---|---|
| speedup from the fix | **1.650x** |
| TornadoVM before vs hand-written CUDA | 1.674x slower |
| TornadoVM after vs hand-written CUDA | **1.014x** slower (1.4%) |

Every arm's stddev is under 0.8% of its median and the p5–p95 band is under 3%, so these are stable distributions rather than one-shot readings. The raw per-launch durations for all 303 launches are in [`per-launch-durations.csv`](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/blob/main/results/raw/31-load-batching-reorder/per-launch-durations.csv) if you want to check the shape yourself.

Two things worth calling out about the "after" arm:

**It is not merely close to hand-written CUDA — it is running the same schedule.** The SASS memory-operation sequence in the staging loop is now identical:

```
tornadovm, before          L S L S L S L S L S L S L S L S
tornadovm, after           L L L L L L L L S S S S S S S S
hand-written CUDA          L L L L L L L L S S S S S S S S
```

The residual 1.4% is not scheduling. TornadoVM still issues its loads as generic `LD` where nvcc issues `LDG`, and it still carries the 16-byte `FloatArray` header offset that costs 5 sectors per warp-wide access instead of 4 — both separate from this PR.

**The compute side was never the problem.** FFMA count is bit-identical across all three arms (1,073,741,824), register count is identical at 72, and there are no spills anywhere. The 1.65x came entirely from `long_scoreboard` stalls falling 12.10 → 1.45 per issue and the issue rate rising 20.8% → 50.9%.

Regenerate the whole table with [`collect.sh`](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/blob/main/results/raw/31-load-batching-reorder/collect.sh):

```bash
./collect.sh "$TORNADOVM_HOME" 1024 100
```

mikepapadim:
## The hand-written CUDA I compared against

Asked what the "hand-written CUDA" arm actually is, so here it is in full. Both sides come from [demo 17 of the companion repo](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/demos/17-matmul-ladder) — a "ladder" of six implementations of the same 1024x1024 sgemm, built so each rung is directly comparable to the one below it. Rung 3 is the register-tiled one, and it has a CUDA twin written to be a line-for-line translation, not a re-implementation.

Shared constants on both sides:

```
BK = 16, TM = 4, TN = 4, THREADS = 16
BM = THREADS * TM = 64          // block tile rows
BN = THREADS * TN = 64          // block tile cols
```

so 256 threads per block stage a 64x16 A-tile and a 16x64 B-tile — **four elements each per tile iteration**, which is where the eight-load staging run comes from.

### Hand-written CUDA

```cuda
__global__ void registerTiled(const float *a, const float *b, float *c, int n) {
    __shared__ float tileA[BM * BK];
    __shared__ float tileB[BK * BN];
    int tx = threadIdx.x, ty = threadIdx.y;
    int tid = ty * THREADS + tx;
    int blockRow = blockIdx.y * BM, blockCol = blockIdx.x * BN;

    float acc[TM * TN];
    for (int i = 0; i < TM * TN; i++) acc[i] = 0.0f;
    float regA[TM], regB[TN];

    for (int kTile = 0; kTile < n; kTile += BK) {
        for (int load = 0; load < (BM * BK) / (THREADS * THREADS); load++) {
            int idx = tid + load * THREADS * THREADS;
            tileA[(idx / BK) * BK + (idx % BK)] = a[(blockRow + idx / BK) * n + kTile + (idx % BK)];
        }
        for (int load = 0; load < (BK * BN) / (THREADS * THREADS); load++) {
            int idx = tid + load * THREADS * THREADS;
            tileB[(idx / BN) * BN + (idx % BN)] = b[(kTile + idx / BN) * n + blockCol + (idx % BN)];
        }
        __syncthreads();
        for (int k = 0; k < BK; k++) {
            for (int i = 0; i < TM; i++) regA[i] = tileA[(ty * TM + i) * BK + k];
            for (int j = 0; j < TN; j++) regB[j] = tileB[k * BN + tx * TN + j];
            for (int i = 0; i < TM; i++)
                for (int j = 0; j < TN; j++) acc[i * TN + j] += regA[i] * regB[j];
        }
        __syncthreads();
    }
    for (int i = 0; i < TM; i++)
        for (int j = 0; j < TN; j++)
            c[(blockRow + ty * TM + i) * n + blockCol + tx * TN + j] = acc[i * TN + j];
}
```

Built with `nvcc -arch=sm_89 -O3`. No `-use_fast_math` — TornadoVM does not enable it either, so turning it on would measure a compiler flag rather than code generation.

### TornadoVM (`KernelContext`)

```java
public static void kcRegisterTiled(KernelContext ctx, FloatArray a, FloatArray b, FloatArray c, int n) {
    int tx = ctx.localIdx;
    int ty = ctx.localIdy;
    int tid = ty * THREADS + tx;
    int blockRow = ctx.groupIdy * BM;
    int blockCol = ctx.groupIdx * BN;

    float[] tileA = ctx.allocateFloatLocalArray(BM * BK); // 64 x 16
    float[] tileB = ctx.allocateFloatLocalArray(BK * BN); // 16 x 64

    float[] acc = new float[TM * TN];
    for (int i = 0; i < TM * TN; i++) {
        acc[i] = 0.0f;
    }
    float[] regA = new float[TM];
    float[] regB = new float[TN];

    for (int kTile = 0; kTile < n; kTile += BK) {
        // 256 threads stage 1024 elements of each tile: four apiece.
        for (int load = 0; load < (BM * BK) / (THREADS * THREADS); load++) {
            int idx = tid + load * THREADS * THREADS;
            int r = idx / BK;
            int cc = idx % BK;
            tileA[r * BK + cc] = a.get((blockRow + r) * n + kTile + cc);
        }
        for (int load = 0; load < (BK * BN) / (THREADS * THREADS); load++) {
            int idx = tid + load * THREADS * THREADS;
            int r = idx / BN;
            int cc = idx % BN;
            tileB[r * BN + cc] = b.get((kTile + r) * n + blockCol + cc);
        }
        ctx.localBarrier();

        for (int k = 0; k < BK; k++) {
            for (int i = 0; i < TM; i++) {
                regA[i] = tileA[(ty * TM + i) * BK + k];
            }
            for (int j = 0; j < TN; j++) {
                regB[j] = tileB[k * BN + tx * TN + j];
            }
            for (int i = 0; i < TM; i++) {
                for (int j = 0; j < TN; j++) {
                    acc[i * TN + j] += regA[i] * regB[j];
                }
            }
        }
        ctx.localBarrier();
    }

    for (int i = 0; i < TM; i++) {
        for (int j = 0; j < TN; j++) {
            c.set((blockRow + ty * TM + i) * n + blockCol + tx * TN + j, acc[i * TN + j]);
        }
    }
}
```

### Launch geometry is identical, and verified rather than assumed

```java
WorkerGrid2D regGrid = new WorkerGrid2D(n / TN, n / TM);   // 256 x 256 global
regGrid.setLocalWork(THREADS, THREADS, 1);                  // 16 x 16 local
```
```cuda
dim3 blockR(THREADS, THREADS), gridR(n / BN, n / BM);       // block 16x16, grid 16x16
```

Both give **grid (16, 16, 1), block (16, 16, 1)** — 256 blocks of 256 threads. Confirmed from `ncu`'s `launch__block_size` / grid dimensions on the TornadoVM side, not from reading the source. This matters: elsewhere in that repo I had a comparison silently invalidated because TornadoVM's default worker grid picked 1024-thread blocks while the CUDA side used 256, and the ratio I attributed to code generation was a launch-configuration artefact.

### So what differs between the two arms

- same algorithm, same tile sizes, same micro-tile, same shared-memory layout
- same launch geometry
- same arithmetic — FFMA count is bit-identical at 1,073,741,824
- same register count (72), no spills on either side

The only difference the profiler can find is the **order of the staging instructions**, which is what this PR changes. After the fix the SASS schedules match exactly (`LDG`/`LD` x8 then `STS` x8 on both sides), and the remaining 1.4% is the generic-`LD`-vs-`LDG` instruction choice plus the 16-byte `FloatArray` header offset — both out of scope here.

Makefile
```diff
@@ -96,6 +96,14 @@ checkstyle:
 test-reflection:
 	./mvnw -P$(JDK_PROFILE) -pl tornado-api,tornado-runtime -am clean test -DskipTests=false
 
+# No-GPU tests for the CUDA code generator (tornado-drivers/cuda/src/test). These build LIR by
+# hand and run the code-generator passes over it, so they need neither a device nor a CUDA
+# driver, and they are the fastest way to check a codegen change before spending a GPU run on it.
+# `-am` is required: tornado-drivers-cuda resolves tornado-drivers-common from the reactor, not
+# from the local repository. Same JDK override as test-reflection.
```
mikepapadim:
```suggestion
```

mikepapadim:
## What the pass is, in compiler terms — and where I think the design earns its keep

Asked to state this precisely rather than descriptively, so:

**`CUDAGlobalLoadBatching` is a local, basic-block-scoped list-scheduling peephole over the LIR, run immediately before emission, whose disambiguation is by *address space* rather than by dependence distance or symbolic index analysis.**

Three parts.

### 1. Region formation

A **scheduling region** is a maximal straight-line run of instructions drawn from a whitelisted alphabet:

| alphabet member | role |
|---|---|
| `LoadStmt` with `!isLocalOrPrivateLoad()` | global/constant load — may float up |
| `AssignStmt` | pure register computation — may float up |
| `StoreStmt` with `isLocalOrPrivateStore()` | shared/private store — may sink |

Anything else is a **scheduling barrier** and ends the region. That includes `__syncthreads()`, shared and private loads, global stores, atomics, MMA statements, calls, and the block's control flow.

The important property is that this is a **whitelist, not a blacklist**. An LIR statement the pass has never heard of — including one added to the backend next year — terminates a region rather than being stepped over. The failure mode of an incomplete whitelist is a missed optimisation, not a miscompile.

### 2. The transform

Within a region: a **stable partition** into (everything that is not a sinkable store) followed by (the sinkable stores). Both subsequences keep their internal order. Nothing is created, deleted, duplicated or rewritten — the instruction multiset is invariant, and that is asserted directly in the property test.

### 3. Dependence preservation

Taking the four classes in turn.

**Flow / RAW on registers.** A store only ever moves *later*, so every value it reads is still defined before it. Non-store instructions do not move relative to each other at all, so every def→use pair among them is untouched.

**Anti / WAR on registers.** This is the *only* hazard the address-space argument does not cover, and it is the one I check explicitly rather than argue about: `canSinkPast` verifies that no instruction jumping ahead of a store defines a value that store reads. It would matter if the LIR ever reused a variable — I do not believe it does today, but "I do not believe it does" is not a safety argument, so it is tested.

**Output / WAW on registers.** No instruction changes position relative to any other instruction defining the same value: non-stores keep their order, and stores define nothing.

**Memory dependences.** Split by space:
- *cross-space* RAW/WAR/WAW cannot exist — `LOCAL`/`PRIVATE` and `GLOBAL`/`CONSTANT` are disjoint in CUDA, and that is precisely the fact ptxas cannot recover from a generic `LD`;
- *intra-space WAW* is preserved by keeping store order;
- *intra-space RAW* cannot arise, because a local or private `LoadStmt` is not in the alphabet and therefore ends the region.

**Barriers.** The region ends before any barrier, so every sunk store still completes before the same `__syncthreads()` it did before. Inter-thread visibility is unchanged by construction, not by analysis.

### Design decisions I'd defend

**Sinking stores rather than hoisting loads.** Hoisting a load requires hoisting the whole address-arithmetic chain that feeds it — five or six instructions in the motivating kernel — and each one needs its own legality check. Sinking a store moves exactly one instruction with no operands to drag along. Same resulting schedule, far smaller proof obligation.

**All-or-nothing per region.** If one store fails `canSinkPast`, the *entire* region is emitted untouched rather than partially reordered. A partial reorder would be legal, but it would mean the emitted schedule depends on where in the region the check failed, which is a much worse thing to debug for a very small extra gain.

**Fail-closed on operands it cannot enumerate.** If a store's operand is a nested `CUDALIROp` expression tree, `usedValues` reports it as unknown and the reorder is refused, rather than walking a structure whose shape the pass would then have to keep in sync with the rest of the backend.

**Not a general scheduler.** There is no dependence graph, no priority function, no latency model. That is deliberate: the scope is one pattern with one provable safety condition, and it is reviewable in an afternoon. A real list scheduler over this LIR would be a much larger thing to trust.

### On registers and spilling — the honest version

I won't claim the transformation cannot increase register pressure, because it can, and pretending otherwise would be the wrong thing to tell a reviewer. **Sinking stores extends live ranges by construction**: *n* staged values are live simultaneously where previously one was. That is the cost side of batching loads, and it is inherent to the optimisation rather than an artefact of this implementation.

What I can say is that it is *bounded* and that I measured it. The pass never creates a value, never duplicates one, and never extends a live range past the end of its region — and a region ends at the first barrier, which in a staging sequence is the `__syncthreads()` that follows it. So the number of simultaneously live staged values equals the number of elements the kernel author chose to stage per thread per tile (8 here), not something the pass can grow.

Measured across every kernel in the demo repo where the pass fires, `ncu`, same build, flag toggled:

| demo | kernel | fires | registers/thread | spill loads | spill stores |
|---|---|---|---|---|---|
| MatMulLadder | `kcRegisterTiled` | yes | 72 → **72** | 0 → 0 | 0 → 0 |
| MatMulLadder | `kcTiled` | yes | 40 → **40** | 0 → 0 | 0 → 0 |
| MatMulLadderFP16 | `kcMma` | yes | 48 → **48** | 0 → 0 | 0 → 0 |
| MatMulLadderFP16 | `kcTiled` | yes | 40 → **40** | 0 → 0 | 0 → 0 |
| TensorCoreDataTypes | `gemmBF16` | yes | 33 → **34** (+1) | 0 → 0 | 0 → 0 |
| TensorCoreDataTypes | `gemmInt8` | yes | 25 → **28** (+3) | 0 → 0 | 0 → 0 |
| MatMulLadder | `naive` | no | 94 → 94 | 0 → 0 | 0 → 0 |

`smsp__inst_executed_op_local_ld.sum` / `_st.sum` are the spill-traffic counters; **zero everywhere, both modes.** Register pressure is unchanged in four of the six kernels that are transformed, and rises by 1 and 3 in the other two — including the motivating kernel, which sinks seven stores and still allocates 72 registers, because ptxas had the headroom.

**So: not provably spill-free in general.** A kernel already at the register ceiling with a long staging region could plausibly spill and come out slower. I have not found one, but I have not proved one cannot exist, and that asymmetry is exactly why I would keep `-Dtornado.cuda.batchGlobalLoads` rather than hard-wire the behaviour.

### On side effects

These I will claim outright, and they are tested rather than asserted:

- **The input LIR is never mutated.** `reorder` returns a new list; `lir.getLIRforBlock(block)` is handed back untouched, and there is a test for it.
- **The instruction multiset is invariant** — nothing lost, added or duplicated. Asserted on every one of 4,000 randomly generated programs.
- **No instruction is rewritten.** Operands, opcodes and results are all untouched; only list position changes.
- **Nothing downstream consumes the reordered list.** It exists only inside `emitBlock`, between the LIR and the emitted C.
- **Output is bit-identical where the pass does not fire** — confirmed on `naive` above, and by 1199 device tests run twice with zero differences in outcome.

mikepapadim:
## What else benefits — measured, and mostly the answer is "nothing"

Asked which other kernels this helps. I swept every kernel in the demo repo that stages through shared memory, `nsys`, same build, flag toggled. The honest answer is narrower than I expected, so here it is in full rather than just the one that moved.

### Where the pass fires

**8 of 14** generated kernels across five demos are transformed:

| demo | kernel | stores sunk | speedup |
|---|---|---|---|
| MatMulLadder | `kcRegisterTiled` | **7** | **1.518x** |
| TensorCoreDataTypes | `gemmBF16` | 4 | 1.028x |
| TensorCoreDataTypes | `gemmFP8E4M3` | 2 | 1.032x |
| TensorCoreDataTypes | `gemmFP8E5M2` | 2 | 1.016x |
| TensorCoreDataTypes | `gemmInt8` | 2 | 1.016x |
| MatMulLadder | `kcTiled` | 1 | 1.018x |
| MatMulLadderFP16 | `kcTiled` | 1 | 1.016x |
| MatMulLadderFP16 | `kcMma` | 1 | 1.011x |

and 6 are untouched (`naive` x2, `rowSumNaive`, `rowSumOptimised`, `gemmMMASingleTile`, `gemmScalarFp16`) — bit-identical generated code, and their timings move 0.979x–1.004x, which is the run-to-run noise floor for this setup.

### The rule

**Only `kcRegisterTiled` moves outside noise, and the number of stores sunk predicts it.** That is not a coincidence — it is the mechanism restated:

> The win is the number of global loads that get to overlap. A kernel staging **one** element per thread per tile iteration has one load in flight either way, and there is nothing to batch.

`kcTiled` stages one element of A and one of B, so its region holds two loads and sinking one store buys ~2%. `kcRegisterTiled` stages four elements each of A and B — eight loads — and that is where the 1.5x lives. So the pattern that benefits is:

- **blocked / register-tiled GEMM**, where each thread stages several elements per tile — the canonical case, and the one every hand-written CUDA GEMM uses
- more generally, any kernel whose staging fan-out per thread is > 1: batched gather-into-shared, multi-row stencil halo loading, tiled transpose with several elements per thread, K-blocked attention staging
- **not** one-element-per-thread reductions, scans or histograms, which is most of what `kernelcontext` currently contains

### Two things worth noting from the sweep

**It is inert where it does not apply.** Six kernels produce byte-identical code, and the two `TensorCoreDataTypes` FP8 kernels that *are* transformed sit at 1.016x–1.032x on single-launch measurements — inside noise for `n=1`, so I would not quote those as wins either.

**The benefit will grow with tile size, not shrink.** The staging fan-out is `(BM x BK) / (threads per block)`. Larger register tiles — the direction anyone tuning a GEMM moves in — mean more loads per region, which is more to batch. The 4x4 micro-tile here is modest; an 8x8 would stage 16 elements per thread.

### What I did not measure

Whether this helps the kernels TornadoVM users actually run outside this repo — GPULlama3.java's matmuls are the obvious candidate, since they are the shape that benefits. I don't have that measured and am not going to claim it.

Raw data: [`per-launch-durations.csv`](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/results/raw/31-load-batching-reorder) and the sweep is reproducible with `collect.sh`.

mikepapadim:
## Correction: why the fp16 ladder gains nothing

In [my earlier comment](https://github.com/beehive-lab/TornadoVM/pull/1079#issuecomment-5586719746) I said the fp16 ladder's kernels "stage **one** element per thread, so there is one load in flight either way and nothing to batch." I inferred that from the store counts instead of reading the generated code. It is wrong on both counts, and the real answer is more interesting.

Both fp16 kernels stage **two** elements per thread, and the pass **does** fire on both and does batch correctly. Here is `--printKernel`, flag off.

### `kcTiled` — genuinely interleaved, and the pass fixes it

```c
ul_57   =  ul_0 + l_56;
half_58 =  *(( __half *) ul_57);
half_3[i_38]  =  half_58;        // <- store, immediately after its load
i_59    =  i_52 + i_7;           //    ... then the next address chain
...
ul_65   =  ul_1 + l_64;
half_66 =  *(( __half *) ul_65);
half_4[i_38]  =  half_66;
__syncthreads();
```

The pass sinks `half_3[i_38]` below the second load, exactly as intended — two loads in flight instead of one. Measured 1.005–1.016x, i.e. noise, because going from **1 to 2** outstanding loads is worth very little when the kernel is dominated by a 64-iteration inner product. The fp32 register-tiled kernel goes from **1 to 8**.

### `kcMma` — the loads are already adjacent

```c
ul_38   =  ul_0 + l_37;
half_39 =  *(( __half *) ul_38);   // load 1
i_40    =  i_23 + i_33;            // only address arithmetic between them
...
ul_44   =  ul_0 + l_43;
half_45 =  *(( __half *) ul_44);   // load 2 -- already back to back
...
ui_51   = (unsigned) __half_as_ushort(half_39);
ui_53   = (unsigned) __half_as_ushort(half_45);
ui_56   =  ui_52 | ui_55;          // pack two halves into one 32-bit word
adi_3[i_50]  =  ui_56;             // ONE shared store for TWO loads
```

The fp16 path packs two `__half` values into a single 32-bit shared store, so the natural emitted order is already **load, load, pack, store**. There is no interleaving left to remove in that block. The pass still fires elsewhere in the kernel — it sinks `adi_4[i_81]` past the next pair of loads — but the gain is 1.014x.

### So the rule, restated correctly

The win is **how many outstanding global loads the batching creates**, and that is bounded by the staging fan-out per thread:

| kernel | elements staged per thread | loads in flight, before → after | speedup |
|---|---|---|---|
| `kcRegisterTiled` (fp32) | 8 | 1 → 8 | **1.39–1.65x** |
| `kcTiled` (fp32 and fp16) | 2 | 1 → 2 | ~1.01x |
| `kcMma` (fp16) | 2, packed into 1 store | 2 → 4 in one block | 1.014x |

Not "nothing to batch" — *little* to batch. The pass behaves correctly on all three; the payoff is what differs.

### One thing this turned up that is worth flagging

`__half_as_ushort` lowers to `HalfBitsToIntStmt`, which is **not** in the pass's whitelisted alphabet, so it terminates a region. In `kcMma` that costs nothing, because it sits after both loads. But a future fp16 kernel that interleaves *load, convert, store, load, convert, store* would be blocked by it rather than optimised.

Adding the pure half-conversion statements (`HalfBitsToIntStmt`, `ConvertHalfToFloatStmt`, `ConvertFloatToHalfStmt` and friends) to the alphabet would be a small, safe extension — they are register computations with no memory effect, exactly like `AssignStmt`. I have deliberately **not** done it in this PR: none of the kernels I can measure would benefit, and I would rather not widen the whitelist of a reordering pass on speculation. Flagging it as the obvious next step if someone hits it.

Full data: [`kernel-sweep.csv` and `ladder-kernel-times.csv`](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/results/raw/31-load-batching-reorder).

mikepapadim:
**Claude Code review** — 1 finding(s)

### nit — tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRStmt.java:1199-1201
The public accessor `LoadStmt.getIndex()` added at CUDALIRStmt.java:1199-1201 has no callers anywhere in the tree — `CUDAGlobalLoadBatching` only reads the load's defined value via `getResult()`, and the three `new LoadStmt(...)` sites in the arithmetic tools invoke `MemoryAccess.getIndex()` (not `LoadStmt.getIndex()`). It was likely added by symmetry with `StoreStmt.getIndex()` (which is used); consider dropping it to avoid maintaining dead public API.



---
_Generated by [Claude Code](https://claude.ai/code)_

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/CUDAGlobalLoadBatching.java
```diff
@@ -0,0 +1,320 @@
+/*
+ * This file is part of Tornado: A heterogeneous programming framework:
+ * https://github.com/beehive-lab/tornadovm
+ *
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
```
orionpapadakis, line 5: 2026

tornado-drivers/cuda/src/test/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDAGlobalLoadBatchingInvariantsTest.java
```diff
@@ -0,0 +1,251 @@
+/*
+ * This file is part of Tornado: A heterogeneous programming framework:
+ * https://github.com/beehive-lab/tornadovm
+ *
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
```
orionpapadakis, line 5: same

tornado-drivers/cuda/src/test/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDAGlobalLoadBatchingTest.java
```diff
@@ -0,0 +1,283 @@
+/*
+ * This file is part of Tornado: A heterogeneous programming framework:
+ * https://github.com/beehive-lab/tornadovm
+ *
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
```
orionpapadakis, line 5: same

tornado-drivers/cuda/src/test/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRTestPrograms.java
```diff
@@ -0,0 +1,131 @@
+/*
+ * This file is part of Tornado: A heterogeneous programming framework:
+ * https://github.com/beehive-lab/tornadovm
+ *
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
```
orionpapadakis, line 5: same

tornado-unittests/src/main/java/uk/ac/manchester/tornado/unittests/codegen/TestSharedMemoryStaging.java
```diff
@@ -0,0 +1,339 @@
+/*
+ * This file is part of Tornado: A heterogeneous programming framework:
+ * https://github.com/beehive-lab/tornadovm
+ *
+ * Copyright (c) 2025, APT Group, Department of Computer Science,
```
orionpapadakis, line 5: same

orionpapadakis (approved):

mikepapadim:
## Review addressed, and the whitelist widened

Four things, all pushed.

### Review comments

**Javadoc cast through `__global float *`** — wrong. `GLOBAL_MEM_MODIFIER` is the empty string, so the emitted C is a plain `float *`; anyone correlating the Javadoc with `--printKernel` would not have found the line. Corrected in both examples.

**Javadoc said "each `LDG` has to retire before the next is issued"** — also wrong, and worse, my own SASS capture had already disproved it before I posted the PR. TornadoVM does not emit `LDG`. The cast carries no address space, so ptxas lowers it to a generic `LD`, which may target shared memory and therefore may not be reordered against a shared store. The Javadoc now states that, with the three schedules that show it, and notes that emitting `__ldg()` would fix the schedule at its source.

**`run(String name, …)` ignored `name`** — fixed; it now names the task graph, so a scheduler or profiler failure names the kernel under test instead of a generic `s0`.

**Makefile comment** — trimmed from five lines to three. If the suggestion meant "drop it entirely", say so and I will.

Also **rebased onto develop** (it had moved on by two commits, including #1055).

### The half-conversion extension, now done

I had flagged this as deliberately out of scope. On reflection that was the wrong call, because it is not speculation — I had a measurable case in front of me and did not check it.

The pass recognised only `AssignStmt` as a pure computation, so **every half, bf16 and fp8 conversion statement ended a region.** That silently skipped the fp16 staging shape: a kernel emitting *load, convert, store, load, convert, store* was left untouched rather than batched. Measured on the fp16 ladder, `kcMma` goes from **one sunk store to two, and 1.014x to 1.034x**. Small, but the blind spot is gone, and it is now covered by a test.

**Design.** Purity is an opt-in marker interface, `PureRegisterComputation`, rather than a list held in the pass. The assertion then sits beside the `emitCode` that has to justify it, and a statement added to the backend later is excluded until someone includes it deliberately — the failure mode of forgetting stays "missed optimisation", never "miscompile". Implementing it asserts three things:

1. the emitted C touches no memory, in any address space — no `*(` and no `[` for a memory access;
2. no cross-lane effect — no barrier, vote, shuffle, `mma` or `ldmatrix`;
3. exactly one value defined.

Seventeen statements qualify: assignment, move, pointer decompression, the half/bf16/fp8 conversions, the scalar and vector half arithmetic, and `HalfBitsToIntStmt`.

**What does not qualify is recorded in the interface's Javadoc with the reason**, so the next person does not have to re-derive it:

| statement | why not |
|---|---|
| `Dp4aStmt` | its inline PTX dereferences both operand pointers |
| `LdmatrixStmt` | `ldmatrix.sync…shared.b16` reads shared memory |
| `ShuffleSyncStmt`, `WarpVoteStmt` | cross-lane |
| `MMAComputeStmt` | cross-lane, and its operands are fragment arrays |
| `CastCompressedStmt`, `SwizzledLoadFP16Stride32Stmt` | both dereference |

Worth saying how that list was arrived at: I first wrote a static scan over the class bodies looking for `emit("*(` and `emit("[`, and **it was not sufficient** — it put `LdmatrixStmt` and `ShuffleSyncStmt` in the pure column, both of which would have been genuinely unsafe. Every candidate's `emitCode` was then read individually. If you would rather this landed as a separate PR, I am happy to split it.

### Verification

- **Full device suite, 1206 tests, zero regressions** against the same `develop` baseline; the 11 failures are identical to the baseline's.
- Unit tests now **19 targeted cases plus the property test**, whose generator emits conversions and moves as well as assignments. Four of the new cases cover the widened alphabet, including the interleaved fp16 staging shape that was previously skipped and a clobber check on a hoisted conversion.
- Checkstyle clean; `TestSharedMemoryStaging` 7/7 with the flag both on and off.

Issue #150, opened 2021-11-03
TornadoVM performs more copies than needed for some applications


jjfumero:
For example, for the mandelbrot application, there should be only a LAUNCH and STREAM_OUT at the bytecode level. 

```java
    public static void mandelbrot(int size, short[] output) {
        for (@Parallel int i = 0; i < size; i++) {
            for (@Parallel int j = 0; j < size; j++) {
                 ...
                output[i * size + j] = r;
            }
        }
    }

   output = new short[size * size];
   ts = new TaskSchedule("benchmark");
   ts.task("t0", ComputeKernels::mandelbrot, size, output);
   ts.streamOut(output);
```

However, we get the following sequence:

```bash
vm: COPY_IN [Object Hash Code=0x28975c28] [S@28975c28 on  [Intel(R) OpenCL HD Graphics] -- Intel(R) UHD Graphics [0x9bc4], size=0, offset=0 [event list=-1]
vm: LAUNCH task benchmark.t0 - mandelbrot on  [Intel(R) OpenCL HD Graphics] -- Intel(R) UHD Graphics [0x9bc4], size=0, offset=0 [event list=0]
vm: STREAM_OUT_BLOCKING [0x28975c28] [S@28975c28 on  [Intel(R) OpenCL HD Graphics] -- Intel(R) UHD Graphics [0x9bc4], size=0, offset=0 [event list=1]
```
I believe the error is related to the data dependencies analyzer. 

**How To Reproduce**

```bash
tornado --enableProfiler silent --printBytecodes uk.ac.manchester.tornado.benchmarks.BenchmarkRunner mandelbrot 50 512
```
**Expected behaviour**

```bash
VM: ALLOCATE
vm: LAUNCH task benchmark.t0 - mandelbrot on  [Intel(R) OpenCL HD Graphics] -- Intel(R) UHD Graphics [0x9bc4], size=0, offset=0 [event list=0]
vm: STREAM_OUT_BLOCKING [0x28975c28] [S@28975c28 on  [Intel(R) OpenCL HD Graphics] -- Intel(R) UHD Graphics [0x9bc4], size=0, offset=0 [event list=1]
```

jjfumero:
This issue is now solved. See https://github.com/beehive-lab/TornadoVM/commit/653750440bad21f37bea8b2bb458b98ca7b456d1

PR #1083, opened 2026-09-11
[cuTile] Java CUDA Tile (cuTile) execution path to the CUDA backend as TileContext  API

mikepapadim:
Adds the **CUDA Tile (cuTile)** execution path to the CUDA backend: a `TileContext` Java API whose
kernels are compiled to CUDA Tile C++ and run alongside the existing SIMT path. The PR's central
claim, and the reason it is framed around CUDA Graphs, is composition:

> A JIT-compiled TornadoVM kernel, a `TileContext` kernel and a native cuBLAS library task are all
> captured into **one** CUDA Graph and replayed from it — not three separate graphs.

## Architecture

`cuTile → Tile IR → GPU` is the peer of `CUDA C++ → NVVM/PTX → GPU`. This PR adds the right-hand
branch; everything below the cubin is existing code, which is why capture and chaining work.

```
            Java TaskGraph · TornadoVM IR  · TornadoVM phase suites      (unchanged)
                                      │
              ┌───────────────────────┴────────────────────────┐
   @Parallel / KernelContext                        task with a TileContext
              │                                                │
   CUDAAssembler → CUDA C                          Tile emitter → __tile_global__
   (inline PTX mma.sync)                           (ct::partition_view, ct::mma)
              │                                                │
            NVRTC                                  nvcc -tilecubin --tile-only
              │                                                │
            cubin                                   cubin (tileiras chose layout/HMMA)
              └───────────────────────┬────────────────────────┘
                                      ▼
              cuModuleLoadDataEx → cuLaunchKernel · one stream per execution plan
                                      │           ← cuBLAS libraryTask binds the same stream
                                      ▼
                  shared device buffers (DataObjectState → XPUBuffer)
```

## Combined APIs example

The example is a committed integration test:
`tornado-unittests/.../unittests/tile/TestTileChaining.java#testWholeChainInOneCudaGraph`

```java
// Stage 1 and 4: conventional JIT SIMT kernels.
public static void scale(FloatArray in, FloatArray out, float factor) {
    for (@Parallel int i = 0; i < in.getSize(); i++) {
        out.set(i, in.get(i) * factor);
    }
}

// Stage 2: CUDA Tile kernel. TileContext is the tile analogue of KernelContext: it exists to be
// intrinsified, has no device representation, and describes what ONE TILE BLOCK does. The worker
// grid for this task counts tile blocks, not threads.
public static void tileDouble(TileContext tc, FloatArray in, FloatArray out, int n) {
    PartitionView iv = tc.partition(tc.view(in, n), TILE);   // -> ct::partition_view
    PartitionView ov = tc.partition(tc.view(out, n), TILE);
    int block = tc.bidX();                                   // -> ct::bid().x
    ov.store(tc.add(iv.load(block), iv.load(block)), block); // -> whole-tile add + store
}

WorkerGrid1D tileWorker = new WorkerGrid1D(SIZE / TILE);     // TILE BLOCKS, not threads
GridScheduler grid = new GridScheduler("graph.tile", tileWorker);

TaskGraph graph = new TaskGraph("graph")
    .transferToDevice(DataTransferMode.EVERY_EXECUTION, input, matrix)
    .task("scale", TestTileChaining::scale, input, scaled, 3.0f)                    // JIT
    .task("tile",  TestTileChaining::tileDouble, new TileContext(), scaled, doubled, SIZE) // Tile
    .libraryTask("gemv", CuBlas::cublasSgemv,                                       // library
            CuBlasOperation.CUBLAS_OP_T.operation(), SIZE, SIZE, 1.0f, matrix, SIZE, doubled, 1, 0.0f, projected, 1)
    .task("bias",  TestTileChaining::bias, projected, result, 0.25f)                // JIT
    .transferToHost(DataTransferMode.EVERY_EXECUTION, result);

    for (int replay = 0; replay < 3; replay++) {
        // First execution captures + instantiates; later executions replay the same graph.
        plan.withGridScheduler(grid).withCUDAGraph().execute();
        checkAgainstCpuReference(result);   // validated after EVERY replay
    }
}
```

`scaled`, `doubled` and `projected` are never transferred to the host: each stage consumes what the
previous one wrote, on the device.

## The `TileContext` API

Documented in full in **`docs/source/tile-api.rst`**, added by this PR and linked from the docs
index. The essentials:

`TileContext` is the tile analogue of `KernelContext`: a parameter that exists for the compiler,
with a JVM-executable body behind every method so a kernel still runs and debugs as plain Java.
The difference from `KernelContext` is the unit of work. A `KernelContext` kernel says what **one
thread** does; a `TileContext` kernel says what **one tile block** does, and the tile compiler
decides how many real threads back it, which tensor-core instruction to issue, and how tiles move
through shared memory. Nothing in the API names a thread, a warp or a fragment.

Three rules follow from the model and are enforced with named errors rather than bailouts:

1. **Tile shapes are compile-time constants and powers of two.** A shape is part of the kernel's
   type, so a differently shaped variant is a differently compiled kernel. Extents are *not*
   constrained — views take dynamic extents, so the problem size does not specialise the kernel.
2. **A view is consumed by `partition(...)`, not stored.** It is parse-time metadata describing a
   buffer, and only the partition view reaches generated code.
3. **The worker grid counts tile blocks**, and the block is pinned to `1x1x1`.

### API → CUDA Tile mapping

Every entry is exercised by the tests in this PR. The spellings were probed against `nvcc 13.3`
rather than taken from documentation, because two of them are not the obvious guess.

| `TileContext` / `PartitionView` call | Generated CUDA Tile C++ | Notes |
|---|---|---|
| `tc.view(array, n)` / `tc.view(array, rows, cols)` | `ct::tensor_span{ptr, ct::extents{...}}` | extents may be runtime values; `ct::assume_aligned(p, 16_ic)` is applied |
| `tc.partition(view, tile...)` | `ct::partition_view{span, ct::shape{..._ic}}` | tile shape must be a power-of-two constant |
| `tc.bidX() / bidY() / bidZ()` | `ct::bid().x / .y / .z` | index of this tile block |
| `tc.numBlocksX()` … | `ct::num_blocks().x` … | grid size in tile blocks |
| `pv.load(bx[, by])` | `view.load(bx[, by])` | caller guarantees in-bounds |
| `pv.loadMasked(...)` | `view.load_masked(...)` | zero-pads a partial tile |
| `pv.store(tile, ...)` / `pv.storeMasked(...)` | `view.store(...)` / `view.store_masked(...)` | out-of-bounds lanes discarded when masked |
| `tc.zeros(dt, shape...)` | `ct::full<ct::tile<E, ct::shape<...>>>(0)` | |
| `tc.full(dt, v, shape...)` | `ct::full<...>(v)` | fill value must fold |
| `tc.iota(dt, n)` | `ct::iota<...>()` | |
| `tc.add / sub / mul / div` | `a + b`, `-`, `*`, **`/`** | broadcasting: a dimension of 1 stretches. **No named divide exists** |
| `tc.scale(tile, s)` | `tile * (E) s` | no named function; scalars broadcast |
| `tc.mma(a, b, acc)` | `ct::mma(a, b, acc)` | tensor cores; the compiler picks the instruction |
| `tc.sum(tile, axis)` | `ct::sum(tile, axis_ic)` | keeps the reduced dimension |
| `tc.max(tile, axis)` | **`ct::reduce_max(tile, axis_ic)`** | `ct::max` is the *elementwise two-operand* form, not the reduction |
| `tc.exp / sqrt / rsqrt` | `ct::exp`, `ct::sqrt`, `ct::rsqrt` | |
| `tc.transpose(tile)` | `ct::transpose(tile)` | |
| `tc.cast(tile, dt)` | **`ct::element_cast<E>(tile)`** | there is no `ct::cast`. Every scalar pair is accepted, narrowing included - established by compiling the matrix, after an earlier probe wrongly suggested narrowing was refused |
| a counted `for` loop | `ct::irange` | a plain counted loop; no iterator API, which would allocate in the kernel |

A reduced tile does **not** store back into the view it came from: `same_shape` rejects a 32x1 tile
against a 32x32 partition view. That is why the row kernels broadcast the reduction back before
storing.

### Second example: softmax, showing reductions and broadcasting

From `TestTileRowKernels`, one row per tile block:

```java
public static void softmaxRow(TileContext tc, FloatArray in, FloatArray out, int rows, int n) {
    PartitionView iv = tc.partition(tc.view(in, rows, n), 1, ROW_TILE);   // rows dynamic, tile constant
    PartitionView ov = tc.partition(tc.view(out, rows, n), 1, ROW_TILE);
    int row = tc.bidX();

    Tile x        = iv.loadMasked(row, 0);        // partial rows zero-padded
    Tile shifted  = tc.sub(x, tc.max(x, 1));      // ct::reduce_max, then broadcast back
    Tile weights  = tc.exp(shifted);
    ov.storeMasked(tc.div(weights, tc.sum(weights, 1)), row, 0);
}
```

### Third example: splitting a row with block indices

`silu_and_mul` needs the first and second half of a `2 * hidden` row. No slicing API is required —
with a tile width of exactly `hidden`, block 0 is the first half and block 1 the second:

```java
PartitionView iv = tc.partition(tc.view(in, rows, 2 * HIDDEN), 1, HIDDEN);
Tile gate  = iv.load(row, 0);                     // first half
Tile value = iv.load(row, 1);                     // second half
Tile silu  = tc.div(gate, tc.add(ones, tc.exp(tc.scale(gate, -1.0))));
ov.store(tc.mul(silu, value), row, 0);
```

`HIDDEN` is a constant, not a parameter, because it is the tile width. Passing it as an argument is
rejected at sketch time with a message naming the argument.

## Capture and replay, as observed

`-Dtornado.print.bytecodes=True` on the example. The indented block is the capture pass; the
un-indented entries after it are replays, and they all carry the **same graph id**:

```
bc: EXECUTION_GRAPH_BEGIN_CAPTURE  graphId=4
        bc: TRANSFER_HOST_TO_DEVICE_ALWAYS  FloatArray@37c2eacb, size=1040   [event list=-1]
        bc: LAUNCH  task graph.scale - scale                                 [event list=0]
        bc: LAUNCH  task graph.tile  - tileDouble                            [event list=1]
        bc: TRANSFER_HOST_TO_DEVICE_ALWAYS  FloatArray@232cce0, size=262160  [event list=-1]
        bc: LAUNCH  task - graph.gemv[cublasSgemv]                           [event list=2]
        bc: LAUNCH  task graph.bias  - bias                                  [event list=3]
        bc: TRANSFER_DEVICE_TO_HOST_ALWAYS  FloatArray@73044cdf, size=1040   [event list=4]
bc: EXECUTION_GRAPH_END_CAPTURE    graphId=4
bc: EXECUTION_GRAPH_LAUNCH         graphId=4      <- replay 1
bc: BARRIER  event-list 16
bc: DEALLOC  [SKIPPED - execution graph active] ...
bc: END
bc: EXECUTION_GRAPH_LAUNCH         graphId=4      <- replay 2
bc: EXECUTION_GRAPH_LAUNCH         graphId=4      <- replay 3
```

Four launches and both transfer directions sit between one `BEGIN_CAPTURE`/`END_CAPTURE` pair, so
all three execution styles are in the same graph; `DEALLOC` is skipped while the graph is live, and
`[event list=n]` shows the dependency chain.

## Running it with a Java `@argfile`

`tornado --printJavaFlags` emits the command line the launcher would use, so an argfile can be
generated rather than hand-written. Two details are required or `java @argfile` fails at once:

* **drop the leading `java` binary**, or it is taken as the main class;
* **inline the nested `@.../exportLists/...` files** - Java does not expand an `@argfile`
  recursively and passes the reference through as a literal argument.

The full recipe, with the argfile generator, is in `docs/source/tile-api.rst` under *Profiling*.
In short:

```bash
tornado --printJavaFlags > flags.raw
# strip the leading java binary, inline nested @files, drop a trailing -m <module/Main>
#   (generator in docs/source/tile-api.rst)
java @tornadovm.args -m tornado.unittests/uk.ac.manchester.tornado.unittests.tile.TileChainBenchmark 2000
```

Application arguments go **after** the main class, not in the argfile.

## How to test

```bash
# 1-2. Build with the CUDA backend
make BACKEND=cuda
source setvars.sh

# 3. Run the combined CUDA Graph test (JIT + Tile + cuBLAS in one graph)
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileChaining

# 4. The rest of the tile suites
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileElementwise
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileMatmul
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileRowKernels

# 5. Verify capture, not just success
tornado-test -V --jvm="-Dtornado.print.bytecodes=True" \
    uk.ac.manchester.tornado.unittests.tile.TestTileChaining | grep EXECUTION_GRAPH

# 6. Inspect the generated tile source (must contain ct:: and no inline PTX)
tornado --printKernel -m tornado.examples/uk.ac.manchester.tornado.examples.tile.TileVectorAdd
```


mikepapadim:
**Claude Code review** — 6 finding(s)

### normal — tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/tasks/TornadoTaskGraph.java:1673
TileContext filtered as host-only in 2 sites but not in 2 similar sites (sweep:`instanceof KernelContext`) — `TornadoExecutionContext.doesExceedExecutionPlanLimit` line 218 still only ignores `KernelContext || AtomicInteger` and hits its `else` branch for a `TileContext` parameter, and `ReduceTaskGraph.checkAllArgumentsPerTask` line 547 still only skips `KernelContext`. The fix is to add `|| parameter instanceof TileContext` to both. [also at: tornado-runtime/src/main/java/uk/ac/manchester/tornado/runtime/graph/TornadoExecutionContext.java:218 - TileContext not handled by doesExceedExecutionPlanLimit's type dispatch: sweep:`instanceof KernelContext`]

### normal — tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDATileCompiler.java:225-227
locateNvcc() accepts the first nvcc with any version > 0 rather than one that meets MINIMUM_TOOLKIT, so a pre-13.3 CUDA_PATH shadows the pip-wheel 13.3 nvcc and disables the tile path even though the docstring promises the userspace install keeps it working.

### normal — tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDATileCompiler.java:165-167
compile() reads process stdout via blocking readFully(readAllBytes) BEFORE calling waitFor(120s), so a hung nvcc that never closes its stdout defeats the timeout guard.

### nit — tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/CUDATileCompiler.java:256-258
versionOf leaks the child nvcc process on 30s timeout: waitFor returns false and the function returns 0 without calling destroyForcibly, unlike compile() which does destroy the process on timeout

### nit — tornado-api/src/main/java/uk/ac/manchester/tornado/api/tile/Tile.java:86-96
Tile.toCppType() reimplements the same `ct::tile<E, ct::shape<...>>` string builder that CUDATileStmt.AbstractTileInstruction.tileType(DType, int[]) already spells out for every tile-valued statement.

### nit — tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/plugins/CUDATileGraphBuilderPlugins.java:534-555
resolveDType duplicates resolveShape from CUDAGraphBuilderPlugins (LoadFieldNode -> readFieldValue -> lookup Enum.ordinal -> index into values()); factor a shared helper such as resolveEnumConstant(GraphBuilderContext, ValueNode, Class<E extends Enum<E>>).



---
_Generated by [Claude Code](https://claude.ai/code)_

mikepapadim:
Thanks — five of the six are real and are fixed in f0b8ece; the sixth was already addressed.

**F1a — `TileContext` missing from the execution-plan limit check** (valid, fixed). `TornadoExecutionContext.doesExceedExecutionPlanLimit` fell through to the unsupported-type branch, so `withMemoryLimit(...)` on any graph containing a tile task threw instead of ignoring a host-only parameter. Added alongside `KernelContext`, with a regression test (`TestTileChaining#testTileTaskUnderMemoryLimit`) — no existing tile test sets a memory limit, which is exactly why this slipped through.

**F1b — `ReduceTaskGraph`** (already fixed). `ReduceTaskGraph:552` already reads `|| parameter instanceof TileContext`; it went in with the `@Reduce` investigation. No change needed.

**F2 — `locateNvcc()` accepted a too-old nvcc** (valid, fixed). It latched the first candidate that answered `--version`, not the first tile-capable one. On the intended setup — a system CUDA 12.x on `PATH` plus a 13.3 wheel under `~/.local` — candidate order could pick the 12.x one and then reject every tile kernel with "requires 13.3". The probe is now `versionOf(candidate) >= MINIMUM_TOOLKIT`.

**F3 — the compile timeout could not fire** (valid, fixed). `compile()` drained nvcc's stdout before `waitFor`, so a compiler that hangs without closing stdout blocked in `readAllBytes` and the 120 s bound never applied. Output now goes to a file in the temp working directory (`builder.redirectOutput`), read back only for the failure message. That also removes the full-pipe-buffer deadlock on a very chatty failure.

**F4 — leaked `--version` probe** (valid, fixed). The 30 s branch returned 0 and left the process running for the lifetime of the JVM, so a hung probe orphaned one child per candidate. Now `destroyForcibly()`, and via a file for the same reason as F3.

**F5 — duplicated `ct::tile` spelling** (valid, fixed). Moved to `DType.tileCppType(DType, int[])` in `tornado-api`, which the driver already depends on; `Tile.toCppType()` and `CUDATileStmt` both delegate. Two copies meant a variable's declaration and its diagnostic description could drift apart.

**F6 — duplicated enum folding** (valid, fixed). The `MMAShape` and `DType` resolvers were the same ordinal-reading walk twice. Factored into `CUDAEnumFolding.resolveEnumConstant(b, node, enumClass, requirement)`; each caller keeps its own actionable message. The explanation of *why* an enum argument needs resolving at all (getstatic folds late, `SnippetReflection.asObject()` is unimplemented) now has one home.

Verification on RTX 4090 / sm_89, driver 6.10, nvcc 13.3.73:

| Suite | Result |
|---|---|
| `TestTileElementwise` / `TestTileMatmul` / `TestTileChaining` | 8 / 6 / 8 pass |
| `TestTileRowKernels` / `TestTileLlmKernels` | 18 / 4 pass |
| `TestMatrixMultiplicationMMA` (+Int8, +BF16) | 14 / 9 / 7 pass — covers the refactored `resolveShape` |
| `TestCuBlas` / `TestCuBlasLt` / `TestCombinedTaskGraph` | 25 / 9 / 5 pass |
| `make checkstyle` | clean |

mikepapadim:
## FP16 GEMM ladder: where the tile path actually lands

Seven rungs, same `C = A * B`, same FP16 inputs, each rung validated against a CPU reference computed over the same FP16-rounded values. Kernel time from Nsight Systems, steady state, RTX 4090 (sm_89), driver 610.57.04, nvcc 13.3.73, `n = 1024`:

| # | Rung | Kernel GPU avg | GFLOP/s | vs naive |
|---|---|---|---|---|
| 1 | naive `@Parallel` | 537.2 µs | 3 997 | 1.0× |
| 2 | `KernelContext` tiled (shared memory) | 331.3 µs | 6 482 | 1.6× |
| 3 | `KernelContext` MMA — hand-written `mma.sync.aligned.m16n8k16` | 181.7 µs | 11 821 | 3.0× |
| 4 | **`TileContext` — `ct::mma`** | **65.7 µs** | **32 666** | **8.2×** |
| 5 | CUTLASS `hgemm` (library task) | 29.9 µs | 71 916 | 18.0× |
| 6 | cuBLAS `GemmEx` FP16 (`ampere_*_s1688gemm_fp16`) | 20.2 µs | 106 488 | 26.6× |

Two numbers matter, and they point in opposite directions:

- **The tile rung is 2.8× faster than the hand-written MMA rung.** Same language, same tensor cores, same validated output. Rung 3 packs two fp16 values per `int`, stages shared memory, names a fixed `m16n8k16` shape and issues two `mma` calls per 16×16 tile; rung 4 is nine lines that say `tc.mma(a, b, acc)` and name no shape at all.
- **cuBLAS is still 3.3× ahead of the tile rung.** One 32×32 tile shape with no pipelining, occupancy or `num_ctas` hints is not a tuned per-architecture library, and the ladder should not pretend otherwise.

Nsight Compute explains the middle of the table:

| Kernel | instructions | bank conflicts | gmem stall | registers |
|---|---|---|---|---|
| `kcMma` | 810 752 | 292 558 | 15.90 | 48 |
| `tiles` | 214 784 | 7 717 | 0.50 | 110 |

A quarter of the instructions and a fortieth of the bank conflicts, with essentially no unhidden global-memory latency — the hand-written rung spends its time on staging that `tileiras` does not need, and the tile kernel pays for that in registers.

Wall clock on the same run is far flatter — 2.0× for the tile rung instead of 8.2× — because every rung pays the same JVM-side dispatch at this size. That is the number not to quote.

The next measurement is the interesting one for this PR: CUDA Tile takes `num_ctas`, `occupancy` and `latency` hints, and a JIT knows the shapes at compile time, so it can set them where a human author has to guess. Nothing in the current lowering emits any hint yet.

Runnable source and the raw nsys/ncu logs: [`demos/22-matmul-ladder-fp16-tile`](https://github.com/mikepapadim/tornadovm-devoxx2026-cuda-demos/tree/main/demos/22-matmul-ladder-fp16-tile) and `results/raw/34-cutile-ladder/` in the demos repo.

## Also in this push

**Ten more tests ported from TileGym**, with the six `ct::` math operations they need (`log`, `log2`, `exp2`, `tanh`, `abs`, `floor` — one line of plugin registration each over the existing unary-math path):

| Kernel | TileGym source | What it adds |
|---|---|---|
| Split-KV decode + combine | `flash_decode.cuh` + `splitk_reduce.cuh` | **two tile tasks chained in one `TaskGraph`**; the partial layout, the log-sum-exp encoding and the streaming rescale have to agree across a task boundary. 4 configurations |
| Attention sinks | `attention_sink.cuh` | a sink logit in the denominator only, read as a `[1, 1]` tile and broadcast |
| Gemma logit soft-capping | `gemma_attention.cuh` | `cap * tanh(s / cap)`, both factors runtime values |
| GELU (tanh approximation), GEGLU | `activation/gelu.cuh`, `geglu.cuh` | the activations that need `tanh` |

Tile suites now **69 tests, all passing**; `make checkstyle` clean.

Also folded into this branch: a fix for silent zeros when a kernel reads and writes the same `HalfFloatArray`. A half-float read lowers to `ReadHalfFloatNode`, which `TornadoDataflowAnalysis` did not recognise — while it did recognise the matching write, because `WriteHalfFloatNode implements MarkWriteNode` and there was no read counterpart. The parameter came out `WRITE_ONLY`, the host→device copy was skipped, and the kernel read an uninitialised buffer. New `MarkReadNode` marker, implemented by all three backends' `ReadHalfFloatNode`; 4 regression tests. Found while writing the hybrid demo, not tile-specific — the generated CUDA was correct all along.

mikepapadim:
## Parity audit against CUDA Tile 13.3, and the gap that was actually blocking kernels

I enumerated the `__tile_builtin__` set in `crt/cuda_tile.h` from the 13.3 toolkit (the authoritative surface — the umbrella `cuda_tile.h` is just a license header and an include) and diffed it against `TileContext`. Builtins found: creation (`full` `zeros` `ones` `iota`), memory (`load` `store` and their masked forms, the whole `atomic_*` family, the `assume_*` hints), shape (`transpose` `permute` `reshape` `broadcast` `extract`), casts (`element_cast` `element_bitcast`), math (`exp` `exp2` `log` `log2` `sqrt` `rsqrt` `sin` `cos` `tan` `sinh` `cosh` `tanh` `abs` `floor` `ceil` `atan2` `fma` `div` `mulhi` `remainder` `isnan` `isinf`), reductions (`sum` `prod` `reduce_max` `reduce_min` `partial_sum` `partial_prod`), `mma`/`matmul`, `select`, and the full operator set including the six comparisons and `& | ^ ~ !`.

One gap stood out as different in kind from the rest. **Everything else on the list makes a kernel more convenient; comparisons and `select` decide whether a class of kernel is expressible at all.** Without a predicate tile, a tile kernel can only compute the same arithmetic for every element — which is exactly why this PR's docs listed causal attention and a ragged key tail as "not expressible", and why the flash-attention port had to keep `S_KV` a multiple of the KV block.

### Implemented (`c1c1b70`)

- **Comparisons** — `lessThan`, `lessOrEqual`, `greaterThan`, `greaterOrEqual`, `equalTo`, `notEqualTo`, each in a tile-to-tile **and** a tile-to-scalar form, yielding a new `DType.PRED` tile. The scalar form is load-bearing, not sugar: the right operand is usually a runtime extent or block offset, and routing it through the tile-to-tile path would need a `full(...)`, whose fill value has to fold.
- **`logicalAnd` / `logicalOr`** — spelled `&` and `|` as CUDA Tile does, and they reject a non-predicate operand at compile time rather than silently doing bitwise arithmetic on floats.
- **`select(mask, a, b)`** → `ct::select`; the only consumer of a predicate tile.
- **`minimum`** (`ct::min`, the elementwise partner of the existing `maximum`), **`min`** (`ct::reduce_min`), **`prod`**, **`fma`**.

Lowering: `CUDATileTernaryNode` + `TileTernaryCallStmt` for `select`/`fma`, and `CUDATileScalarCompareNode` + `TileScalarCompareStmt` for the scalar form, whose cast names the *operand* element type rather than `bool`.

### What that unlocked — `TestTileMasking`, 10 tests

| Test | Why it could not exist before |
|---|---|
| Causal flash attention, 3 sizes | TileGym's `IS_CAUSAL` branch. The mask comes from `iota` index tiles and a **shifted threshold** (`keyBlock * BLOCK_N - queryBlock * BLOCK_M`) rather than shifted indices, which keeps the runtime block offsets out of the tiles where they would need a constant `full(...)`. The loop also stops at the diagonal block instead of masking whole blocks away. |
| Ragged key tail, 3 sizes (incl. KV shorter than one block) | `loadMasked` keeps the read in bounds, `select` keeps the zero padding out of the softmax denominator. Both halves are needed — a padded key otherwise contributes `exp(0 - m)`. |
| ReLU by select, clamp, runtime-threshold count | the elementwise cases where the expected value is checkable by hand |
| `reduce_min`, `prod` | the two reductions that arrived with them |

Tile suites are now **79 tests, all passing** (MMA 14, cuBLAS 25, half-float 4 unchanged; checkstyle clean).

### The honest remainder

`docs/source/tile-api.rst` now carries the full parity table, ordered by what each gap costs rather than by how easy it is to add:

| Missing | Consequence |
|---|---|
| views of `FP8Array` / `Int8Array` / `DoubleArray` | **the largest remaining gap.** `DType` names fp8, s8, f64 and tf32 and `mma` validates them against the `mmaf`/`mmai` tables, but only `FloatArray`, `HalfFloatArray`, `BFloat16Array` and `IntArray` can be viewed — so a quantised GEMM has no way to be fed |
| rank ≥ 3 views | a batch or head dimension is folded into the row index (every attention test does this); costs arithmetic, not expressiveness |
| `atomic_add` and family | split-K must write per-split partials and reduce in a second kernel, which is what the decode tests do |
| `permute` `reshape` `broadcast` `extract` | rank-2 reshaping is `transpose` only; broadcasting is implicit in elementwise ops rather than on demand |
| `partial_sum` / `partial_prod` | no scans |
| `tan sinh cosh atan2 isnan isinf mulhi remainder`, `element_bitcast`, `logicalNot` | individually cheap; nothing ported has needed them |
| per-operation rounding / NaN-propagation tags | the generated code takes the defaults |
| `num_ctas` / `occupancy` / `latency` hints | nothing emits them yet, though a JIT knows the shapes that would inform them — still the most interesting thing this integration could do that a human author cannot |

Next by value is dtype reach (fp8/int8 views → quantised GEMM, which is a large part of TileGym), then atomics.

<sub>Incidental: a method reference inside a cast in a switch expression crashes javac 21 (`DeferredAttr` assertion), so `zip`'s dispatch is a statement switch over lambdas.</sub>

mikepapadim:
## Dtype reach closed: int8 quantised GEMM, doubles, and an fp8 gate

Following the parity audit, this is the gap it named as largest — `DType` had named fp8, s8, f64 and tf32 from the start and `mma` validated them against the `mmaf`/`mmai` tables, but only `FloatArray`, `HalfFloatArray`, `BFloat16Array` and `IntArray` could be viewed, so a quantised GEMM had nothing to read.

**Measured against nvcc 13.3.73 before writing any Java**, by compiling the same three kernels for `sm_89` and `sm_90`:

| Element type | sm_89 (Ada, this box) | sm_90 |
|---|---|---|
| int8 × int8 → int32 | works — **`IMMA.16832.S8.S8.SAT`** | works |
| fp8 e4m3 (even load + `element_cast`, no `mma`) | **rejected**: `Incompatibility with architecture 'sm_89': unsupported type 'f8E4M3FN'` | compiles |
| fp64 `mma` | compiles, but lowers to scalar `DADD`/`DMUL` — **no `DMMA`** | same at this shape |

So fp8 is not an implementation gap, it is an architecture floor, and fp64 is correct-but-unaccelerated on Ada. Both are now documented rather than implied.

### Added (`f9e2dc5`)

- **Views for `Int8Array`** (`S8`), **`DoubleArray`** (`F64`) and **`FP8Array`**. The fp8 view takes the format as an argument — `FP8Array` is a byte buffer carrying *both* an e4m3 and an e5m2 accessor, so the buffer cannot tell you which it holds and the kernel has to say. The JVM fallback reads through whichever accessor the view's element type names.
- **`CUDATileSupportPhase` refuses fp8 below cc 9.0** with `The tile operation 'partition view' uses FP8_E4M3, and CUDA Tile supports fp8 tiles only from compute capability 9.0 (Hopper) onwards; device reports 8.9. Use F16 or BF16 operands on this device.` — instead of that `f8E4M3FN` error surfacing from inside nvcc.

### `TestTileDTypes` — 8 tests

| Test | Why it is worth trusting |
|---|---|
| int8 matmul, three shapes (square, single tile, rectangular) | int8→int32 is **exact**, so the assertions are equalities, not tolerances. A wrong operand layout or an unintended saturating accumulate cannot hide in a rounding budget. |
| int8 matmul + dequantising epilogue | accumulate in int32, `cast` to float, apply the scale — what an inference kernel actually wants, and it exercises a cast out of an integer tile |
| int8 elementwise `abs` | the integer path without `mma` |
| double row reduction | bound at **1e-9**; 64 terms near 1.2 in fp32 carry ~1e-5 of error, so an fp32 intermediate could not pass |
| double matmul | bound at 1e-12 |
| fp8 refusal | asserts the message names fp8 *and* the capability it needs *and* the device's own |

Tile suites now **87 tests, all passing**; checkstyle clean.

### One trap worth flagging for anyone porting these tests

The fp8 test depends on `-Dtornado.recover.bailout=False`, which `tornado-test` sets. With bailout enabled the refusal is swallowed and the task **silently runs `TileContext`'s JVM fallback, which computes the right answer on the host** — so the test passes without the gate ever being exercised. I hit exactly that while writing it: the suite went green, but `--printKernel` showed no `fp8Scale` kernel had been generated at all. Verified the real refusal by hand under `-J"-Dtornado.recover.bailout=False"`, and the test now documents the dependency.

That is a general hazard for this API, not just for this test: because every `TileContext` operation has a working JVM implementation, a tile kernel that fails to compile can look like a passing test. Worth keeping in mind for any tile test added later.

### Remaining parity gaps, updated

Still open, in rough order of what they cost: **atomics** (`atomic_add` and family — split-K currently needs a second kernel, which the decode tests do), **rank ≥ 3 views** (batch/head dimensions are folded into the row index today), **`permute`/`reshape`/`broadcast`/`extract`**, **scans** (`partial_sum`/`partial_prod`), the remaining scalar math (`tan sinh cosh atan2 isnan isinf mulhi remainder`, `element_bitcast`, `logicalNot`), per-operation rounding/NaN tags, and — still the most interesting one — **no kernel emits `num_ctas`/`occupancy`/`latency` hints**, though a JIT knows the shapes that would inform them. `docs/source/tile-api.rst` carries the full table.

mikepapadim:
## Atomic accumulation through a view — split-K now finishes in one kernel

Next item from the parity list (`7bbfad2`). `PartitionView.atomicAdd(tile, blockIndices...)` accumulates a tile so every concurrent block's contribution survives, which collapses the two-kernel split-K arrangement (write per-split partials, reduce in a second pass) into a single launch with no intermediate buffer.

### It does not lower like a store, and the header is why

I checked the 13.3 header rather than assuming a symmetry: **`partition_view` has no atomic read-modify-write.** It offers `atomic_load` and `atomic_store` only; `ct::atomic_add` takes a **tile of pointers**:

```cpp
template<memory_order Order, thread_scope Scope = default_thread_scope(),
         storeable_tile Ptrs, non_narrowing_tile_convertible_to<tile_load_t<Ptrs>> Values>
__tile_builtin__ tile_load_t<Ptrs> atomic_add(Ptrs ptrs, Values values, ...);
```

Two consequences that shaped the implementation:

1. `Order` has **no default**, so the tag must be passed explicitly — without it every overload fails to match, which is what the first probe hit.
2. The node carries the view's **buffer and extents** rather than the view value, and the statement builds the pointer tile itself. Rank 2 needs the row and column index within the tile separately, and **neither `/` nor `%` is defined on a tile**, so the indices come from two `iota` tiles broadcast to the tile shape rather than from dividing a linear one.

Order is `memory_order_relaxed`, scope is `thread_scope_device`. The scope is not cosmetic — measured SASS:

| Kernel | Instruction |
|---|---|
| row accumulator | `ATOMG.E.ADD.F32.FTZ.RN.STRONG.GPU` |
| split-K matmul | `RED.E.ADD.F32.FTZ.RN.STRONG.GPU` |

The split-K kernel gets `RED` rather than `ATOMG` because the returned old value is unused and the compiler picks the fire-and-forget reduction form. Default (system) scope would emit the more expensive `.SYS` variants.

### `TestTileAtomics` — 4 tests, none of which can pass by accident

Every test targets an output region that several blocks write, so a plain store would lose contributions non-deterministically. A test that would also pass with a normal store would prove nothing.

| Test | Contention |
|---|---|
| `testAccumulateRowsRank2` | 96 blocks accumulate into the same 64-element row |
| `testAccumulateRank1` | 64 blocks, rank-1 view (the simpler pointer arithmetic) |
| `testSplitKTwoSplits` / `testSplitKFourSplits` | single-kernel split-K matmul, K split along `bidZ`, 2 and 4 blocks contending per output tile |

Tile suites now **91 tests, all passing**; checkstyle clean.

### Two failures on the way, both worth knowing for anyone extending the plugins

- **`receiver.get(true)` leaves a `NullCheckNode` behind** when a plugin does not consume the receiver as a node input, and the CUDA backend has no emitter for one — it failed with a bare `TornadoInternalError: unimplemented` at LIR generation with no indication of the cause. This plugin uses `get(false)`.
- **Bypassing the view node means the view has no usages and is eliminated**, so nothing marked the accumulator as written. The dataflow analysis called it read-only, no `TRANSFER_DEVICE_TO_HOST` was emitted, and the accumulator read back as all zeros while the kernel was in fact doing the atomics correctly. `CUDATileAtomicAddNode` now implements `MarkArrayParameterAccess` itself and reports `READ_WRITE`. This is the same class of bug as the half-float read marker fixed earlier in this branch, and the second time an access-marking omission has produced *silently wrong data* rather than an error — worth remembering that any new node touching a buffer must mark it.

### Parity list, updated

Closed since the audit: comparisons + `select`, `minimum`/`min`/`prod`/`fma`, int8/double/fp8 views, `atomicAdd`. Still open: the rest of the atomic family (`atomic_sub`/`min`/`max`, bitwise, `xchg`, `compare_exchange`, the masked forms, and a choice of order/scope), **rank ≥ 3 views**, `permute`/`reshape`/`broadcast`/`extract` as explicit operations, scans (`partial_sum`/`partial_prod`), the remaining scalar math, per-operation rounding tags, and still no `num_ctas`/`occupancy`/`latency` hints.

mikepapadim:
## Shape operations: `broadcast`, `reshape`, `extract` (`6582500`)

Each of these replaces a workaround that earlier kernels in this branch were using, so the tests are as much cleanup as coverage:

| Operation | What it replaces |
|---|---|
| `broadcast` | the mask kernels stretched an `iota` row to the score shape by **adding a zero tile of that shape** — arithmetic doing a layout job. `TestTileMasking` now says `broadcast`, which is what it meant. |
| `reshape` | changing a `[rows, 1]` reduction result to another shape of the same size meant `transpose` — a data movement rather than a reinterpretation. It is also the only way to change *rank*. |
| `extract` | splitting a packed row meant loading the same row twice through a narrower partition view; one load plus two extracts is what the kernel actually wants. |

**`extract` indexes in sub-tile units, not elements.** The header does not say which, so I ran it rather than guessing: `extract(row, shape<1,32>, 0, 1)` on a `[1, 64]` tile returns elements **32..63**, i.e. block-style addressing like a partition view. Exposing that backwards would have been a silent correctness trap.

`permute` is deliberately **not** added: on a rank-2 tile the only non-identity permutation is the `transpose` that already exists, so it would be surface without capability. That is now what the docs say instead of listing it as missing.

All three lower through one node and one statement emitting `ct::fn(tile, ct::shape<...>{})`, with `extract`'s indices appended.

### `TestTileShapes` — 4 tests

- a bias row broadcast across a tile
- a row-sum column broadcast to normalise each row (the shape rule in the other direction)
- a `[1, 1]` reduction reshaped to rank 1 for a rank-1 output
- SiLU-and-multiply splitting a packed row with two extracts from a single load

Tile suites now **95 tests, all passing**; checkstyle clean.

<sub>One incidental: `InvocationPlugin`'s constructor takes `Type...`, so a plugin with a computed arity has to pass a single `Type[]` with the receiver first — a nested `Class[]` is one argument of the wrong type rather than a spread signature.</sub>

### Parity status after four iterations

**Closed:** comparisons + `select` + `logicalAnd`/`Or`, `minimum`/`min`/`prod`/`fma`, `log`/`log2`/`exp2`/`tanh`/`abs`/`floor`, int8 + double + fp8 views (with the fp8 capability gate), `atomicAdd`, `broadcast`/`reshape`/`extract`.

**Still open,** in what I read as value order:
1. **`num_ctas` / `occupancy` / `latency` hints** — nothing emits them, and this is the one item where a JIT can do better than a human author rather than merely catching up: TornadoVM knows the shapes at compile time. The FP16 ladder put the tile GEMM 2.8x ahead of hand-written `mma.sync` but still 3.3x behind cuBLAS, and untuned launch hints are a plausible part of that gap.
2. **rank ≥ 3 views** — every attention kernel currently folds batch and head into the row index.
3. the rest of the atomic family (`sub`/`min`/`max`/bitwise/`xchg`/`compare_exchange`, masked forms, order and scope choice).
4. scans (`partial_sum`/`partial_prod`).
5. remaining scalar math (`tan sinh cosh atan2 isnan isinf mulhi remainder`, `element_bitcast`, `logicalNot`) and per-operation rounding/NaN tags.

mikepapadim:
## Launch hints: plumbed, measured, and deliberately not automated (`7a27b59`)

This was the last item on the parity list that a hand-written tile kernel can do and this integration could not. `-Dtornado.cuda.tile.hints=occupancy=2,num_cta_in_cga=1` now puts a `[[ using cutile : hint(0, ...) ]]` attribute on every generated tile kernel.

The interesting part is the measurement, not the plumbing. FP16 GEMM from `demos/22`, `n = 1024`, sm_89, kernel time from nsys:

| Hint | Registers / shared | Kernel time | |
|---|---|---|---|
| none (compiler's choice) | 110 / 36 KB | **65.3 µs** | baseline |
| `occupancy=2` | 110 / 36 KB | **63.2 µs** | −2.9%, reproducible 3/3 runs (65.29/65.39/65.37 vs 63.57/63.52/63.10) |
| `occupancy=4` | 125 / 20 KB | 132.5 µs | **2.0× slower** |
| `occupancy=8` | 64 / 8 KB | 459.5 µs | **7.0× slower** |

The hint reaches `tileiras` and visibly changes its resource decisions — that is what the register/shared column shows. But **the default is already near the best of these, and forcing occupancy up costs 2× to 7×.**

That answers the question I raised in the earlier comments. I had this listed as the most promising thing a JIT could do that a human cannot, on the reasoning that TornadoVM knows the shapes at compile time. It still is — but a heuristic of the obvious kind ("we know the tile is 32×32 fp16, so ask for more occupancy") would make this kernel between two and seven times slower. So nothing derives hints automatically, and the docs now say why rather than leaving it as a TODO that invites someone to add the harmful version. What it needs is a model of the register/occupancy tradeoff, and that is a genuine research question — and a good one to put to NVIDIA, since `tileiras` already has such a model internally and is the thing making the good default.

Two findings worth recording, both verified against 13.3.73:

- **The attribute must come after `extern "C"`.** Before it, nvcc rejects it with `attributes are not allowed here`. (TileGym's kernels have no `extern "C"`, so its placement doesn't transfer.)
- **Hint keys are not validated.** `hint(0, nonsense_key=3)` and `hint(0, occupancy=99999)` both compile silently and do nothing. So a hint that appears to have no effect may simply not exist — there is no way to tell a typo from a no-op. Worth asking whether that is intended.

`TestTileMatmul#testGemmWithLaunchHint` covers the plumbing: the property is read at code generation time in-process, so the test sets it, runs the GEMM, and checks the one thing a hint must never change — the result.

Tile suites now **96 tests, all passing**; checkstyle clean.

### Parity status

Everything I set out to close from the audit is closed: comparisons + `select` + mask combinators, `minimum`/`min`/`prod`/`fma`, six more math functions, int8/double/fp8 views with a capability gate, `atomicAdd`, `broadcast`/`reshape`/`extract`, and launch hints.

What remains is a short and honest list, none of which currently blocks a kernel anyone has tried to write: rank ≥ 3 views (attention folds batch and head into the row index instead), the rest of the atomic family, scans (`partial_sum`/`partial_prod`), the remaining scalar math (`tan sinh cosh atan2 isnan isinf mulhi remainder`, `element_bitcast`, `logicalNot`), per-operation rounding and NaN tags, and `permute` — which stays out on purpose, since rank-2's only non-identity permutation is the `transpose` that exists.

mikepapadim:
## Rank-3 views (`6d1e916`) — the last substantive gap from the audit

`view`, `partition`, `load`/`store` (and their masked forms) and `reshape` now take three extents, so a batch or head dimension is a **block index** instead of arithmetic folded into the row index.

The contrast is with `TestTileLlmKernels#bmm`, the batched matmul already in this PR. Over a rank-2 view of `(batch * m, k)` it has to derive the batch from the block index, multiply by the number of K tiles, and add that to every operand index. None of it is wrong; all of it is bookkeeping the view can do:

```java
// before: rank-2 view, batch folded into the row index
int flatRowBlock = tc.bidX();
int batchIndex = flatRowBlock / rowBlocks;
int bRowBase = batchIndex * kBlocks;
acc = tc.mma(av.load(flatRowBlock, step), bv.load(bRowBase + step, columnBlock), acc);

// after: rank-3 view, batch is just an index
PartitionView aView = tc.partition(tc.view(a, batch, m, k), 1, TILE, TILE);
Tile left = tc.reshape(aView.load(batchIndex, rowBlock, step), TILE, TILE);
acc = tc.mma(left, right, acc);
```

**Only the API surface and the plugin registrations were rank-limited** — the LIR statements already loop over their extents and indices, so the emitter needed no change whatsoever. That was a pleasant surprise and worth recording: the rank limit was in the Java signatures, not in the lowering.

Rank 3 is for addressing, not arithmetic. `mma`, the reductions and `transpose` are rank-2 operations, so a `[1, M, N]` load is reshaped to `[M, N]` before use — the same shape NVIDIA's kernels take, loading a rank-4 tile from a `[B, H, S, D]` view and reshaping before the first multiply.

### `TestTileRank3` — 4 tests

- batched matmul at three sizes, including a batch of one (where the rank-3 view is pure bookkeeping and still has to be right)
- per-batch scaling where every element of batch `b` is multiplied by `b + 1`, so a kernel that ignored the batch index **fails** rather than merely being slower

One thing the tests document, because I hit it while writing them: **view extents may be runtime values, tile extents may not.** Passing the row width in as a parameter was rejected at sketch time by the compile-time-shape rule — correctly, since a tile shape is part of the kernel's type — so the width is a constant while the batch and row counts stay parameters.

Tile suites now **100 tests, all passing**; checkstyle clean.

### Where parity stands

Six iterations of audit-then-close. Everything that blocked or awkwardly shaped a kernel is done: comparisons + `select`, the missing reductions and math, int8/double/fp8 views, `atomicAdd`, `broadcast`/`reshape`/`extract`, launch hints, rank-3 views.

What is left is genuinely minor, and I would not add any of it speculatively:

| Remaining | Why it can wait |
|---|---|
| rank ≥ 4 views | only needed to stop folding *two* leading dimensions; rank 3 covers batch **or** head |
| rest of the atomic family (`sub`/`min`/`max`/bitwise/`xchg`/`compare_exchange`, masked forms, order and scope choice) | mechanical once `atomicAdd` exists; nothing has needed them |
| scans (`partial_sum`/`partial_prod`) | no ported kernel uses a prefix operation |
| `tan sinh cosh atan2 isnan isinf mulhi remainder`, `element_bitcast`, `logicalNot` | one line of registration each, on demand |
| per-operation rounding / NaN tags | defaults have matched every reference so far to the tolerances in the tests |
| `permute` | stays out: rank-2's only non-identity permutation is the `transpose` that exists |

The one item I would *not* treat as a gap to close is automatic hint derivation — the measurement two comments up shows the compiler's default beating every occupancy I forced, so that needs a model, not an implementation.

mikepapadim:
## A Q4_0 quantised projection ported from GPULlama3.java#150, measured with nsys

Ported `Qwen35MMAKernels.projectionMMAQ4_0` from [GPULlama3.java#150](https://github.com/beehive-lab/GPULlama3.java/pull/150) to the tile API as a runnable example (`tornado-examples/.../tile/TileQuantizedProjection.java`), and measured it.

### What the original does

It reaches tensor cores by hand, and the code says so: a lane owns eight consecutive weights, nibbles are unpacked per lane, the B panel is staged into shared memory **through a swizzle**, two barriers bracket each Q4_0 block, and two `mma.sync` calls are issued per staging round against `ctx.mmaFragment` accumulators. Around 120 lines of index arithmetic, with comments recording which of several stagings measured fastest — including that four panels per warp measured 50.7 t/s against 62.1 for one.

### What the tile version is

The loop body you would write if the hardware were not in the way — mask a nibble, recentre it, scale it, multiply:

```java
for (int byteBlock = 0; byteBlock < halfBlocks; byteBlock++) {
    Tile bytes = tc.cast(packedView.load(columnBlock, byteBlock), DType.S32);

    Tile lowNibble  = tc.bitwiseAnd(bytes, mask);
    Tile lowScale   = tc.broadcast(scaleView.load(columnBlock, byteBlock), TILE, TILE);
    Tile lowWeights = tc.mul(tc.sub(tc.cast(lowNibble, DType.F32), recentre), lowScale);
    acc = tc.mma(aView.load(rowBlock, byteBlock), tc.transpose(tc.cast(lowWeights, DType.F16)), acc);

    Tile highNibble  = tc.bitwiseAnd(tc.shiftRight(bytes, four), mask);
    Tile highScale   = tc.broadcast(scaleView.load(columnBlock, halfBlocks + byteBlock), TILE, TILE);
    Tile highWeights = tc.mul(tc.sub(tc.cast(highNibble, DType.F32), recentre), highScale);
    acc = tc.mma(aView.load(rowBlock, halfBlocks + byteBlock), tc.transpose(tc.cast(highWeights, DType.F16)), acc);
}
```

No lane, no barrier, no fragment, no swizzle. A Q4_0 byte holds one weight in each nibble, so each byte tile feeds two k-steps and both halves are used exactly once. The bitwise and shift operations it needs landed in this branch today.

### Kernel time, nsys, RTX 4090 / sm_89, m = n = 1024, k = 2048

| Implementation | Kernel time | |
|---|---|---|
| `@Parallel`, dequantise per output element | **20 518.4 µs** | baseline |
| `TileContext`, dequantise in tiles | **87.0 µs** | **236× faster** |

Wall clock over the same run: 20.919 ms against 0.524 ms (39.9×). The kernel figure is the one that describes the loop body; the wall-clock figure includes JVM dispatch, which is why the two differ so much.

The tile kernel's cubin: **8× `HMMA.16816.F32`**, 114 registers, 21 KB shared memory. Both implementations validate against a host dequantisation (max error 0.0013 at that size).

To be fair about what the 236× measures: the thread-level baseline re-reads and re-dequantises a weight for every output element it contributes to, which is the naive shape, not what #150 wrote. **It is not a comparison against the hand-tuned `mma.sync` kernel** — that one is embedded in GPULlama3's model runtime and is not runnable here. What the number does show is the cost of having no reuse and no tensor cores, which is the trap a straightforward quantised matmul falls into.

### One gap this found, worth recording

GGUF interleaves the FP16 block scale with the nibbles in a single buffer: `[fp16 scale][16 bytes of nibbles]` per block. **A tile kernel cannot read that scale**, because it would mean viewing one byte buffer as both `S8` and `F16` at a byte offset, and a view is one element type over one buffer. The example keeps the scales in their own array and says so; a production port would either store them that way or split them out in a preparation kernel.

That is a genuine limitation of the view model rather than a missing operation, and it is the kind of thing worth raising with NVIDIA: a `bitcast` on a *view* (as opposed to on a tile, which now exists) would solve it.

### Parity, updated

Coverage of the CUDA Tile C++ operation set is now **71 of 84 (84%)**, up from 65 after adding `shiftLeft`/`shiftRight`, `prefixSum`/`prefixProduct` (`ct::partial_sum`/`partial_prod`), `bitcast`, `concat` and the view's `atomicLoad`/`atomicStore`. What remains is 12 masked atomic variants whose unmasked forms exist, plus `permute`. Tile suites: **113 tests, all passing**.

mikepapadim:
@orionpapadakis @kotselidis — perf numbers on the GPULlama3 kernel ports, would value your read on these.

## Three more Qwen3 kernels ported, from #150 at `7aa44d7e` onwards

Checked that PR from `7aa44d7e` to its head (21 commits) for kernels not already covered. Four are new in that range:

| Kernel | Where | Ported? |
|---|---|---|
| `l2NormPerHeadWide` | `Qwen35DeltaNetKernels` | yes |
| `gatedNormPerHeadWide` | `Qwen35DeltaNetKernels` | yes |
| `rmsApplyAndQuantizeActivationQ8Blocks` | `TransformerComputeKernelsQ4_0` | yes |
| `matrixVectorGenericWithResidualQ4_1DP4A` | `TransformerComputeKernelsQ4_1` | no — see below |

The rest of the range is two-bytes-at-a-time weight reads and documentation.

All three ported kernels share the shape that makes them worth porting: **a workgroup owns one row, a lane owns one element, and the row's reduction is a log-stride loop over a shared array with a barrier per step.** That is the standard workgroup reduction, and it is exactly what a tile-level reduction replaces.

### Kernel time, nsys, RTX 4090 / sm_89, 4096 heads of 256

| Kernel | thread-level (`KernelContext`) | tile (`TileContext`) | |
|---|---|---|---|
| L2 norm per head | 7.99 µs | **3.74 µs** | 2.14× |
| Gated norm per head (weight + SiLU gate) | 9.11 µs | **4.55 µs** | 2.00× |
| RMS apply fused with Q8 block quantisation | 18.38 µs | **13.72 µs** | 1.34× |

Wall clock over the same run is flat to within noise, which is why these are nsys figures.

**These are modest next to the 236× on the quantised projection** ([earlier comment](https://github.com/beehive-lab/TornadoVM/pull/1083#issuecomment-5677039382)), and that is the honest shape of it: these three are bandwidth-bound elementwise-and-reduce kernels, so the ceiling is the memory system, not the reduction. The 1.3–2.1× is what removing the shared-memory round trip and the barriers is worth; it is not a tensor-core story.

### The fused quantiser is the one to look at

The thread-level version runs **two** log-stride reductions over shared memory (max-abs, then the sum), keeps a second shared array for the quantised values, and packs four of them into an `int` by hand. The tile version views the row as `[blocks, 32]`:

```java
Tile value  = tc.mul(weightView.load(row, 0), tc.mul(ss, xView.load(row, 0)));
Tile maxAbs = tc.max(tc.abs(value), 1);                       // one op, was a log-stride loop
...
quantView.store(tc.cast(clamped, DType.S8), row, 0);          // no packing: an Int8Array view
scaleView.store(tc.reshape(tc.scale(maxAbs, 1.0 / 127.0), 1, BLOCKS), row, 0);
sumView.store(tc.reshape(tc.sum(clamped, 1), 1, BLOCKS), row, 0);
```

Both reductions become one operation each, and the quantised values go straight into an `Int8Array` — the hand packing disappears because a view of 8-bit elements is a view like any other. Rounding is half away from zero through a `select` over the sign, since there is no `sign` operation.

`matrixVectorGenericWithResidualQ4_1DP4A` is deliberately not ported: its tile equivalent is an int8 `mma`, which `TestTileDTypes` already covers (`IMMA.16832.S8.S8.SAT`), and Q4_1's zero-point dequantisation adds nothing the Q4_0 example does not already show.

### One harness bug, recorded because it nearly passed

The gated norm writes its input **in place**. That is right inside a model — it runs once per layer per token on a fresh activation — but a timing loop re-runs it on its own output, and the map is a contraction, so values decayed toward zero and the result depended on the iteration count. What pointed at the harness rather than the kernels was that **both implementations agreed with each other and disagreed with the host reference**. It is out of place in the example, with the reason in a comment; the L2 norm survives the same treatment because normalising a normalised vector is a no-op.

### Where to find it

`tornado-examples/.../tile/TileQwenNormKernels.java` on `feat/cutile` (`e80dd66`), runnable as:

```bash
tornado -m tornado.examples/uk.ac.manchester.tornado.examples.tile.TileQwenNormKernels 4096 50
```

Every row validates against a host reference, and the example prints `correct`/`WRONG` per implementation rather than asking to be trusted. Seven tile examples now; tile suites at 113 tests.

mikepapadim:
## Correction: a tile kernel *can* read GGUF's interleaved block. That claim of mine was wrong.

In [the quantised-projection comment above](https://github.com/beehive-lab/TornadoVM/pull/1083#issuecomment-5677039382) I wrote that reading GGUF's `[fp16 scale][16 nibble bytes]` block from a tile kernel is impossible because "a view is one element type over one buffer", called it "a genuine limitation of the view model", and suggested raising a `bitcast`-on-a-view request with NVIDIA.

**That was wrong, and I should not have published it without testing it.** I built the reproducer intending to file it upstream, and it disproved the premise instead, so nothing was filed.

CUDA Tile expresses the layout directly. `ct::tensor_span` takes a *layout*, and two spans may share one allocation at different byte offsets with different element types:

```cpp
// one buffer: [fp16 scale][16 nibble bytes] per 18-byte block
auto scaleSpan = ct::tensor_span{
    reinterpret_cast<const __half *>(raw),
    ct::layout_strided_mapping<ScaleExtents, ScaleStrides>{
        ScaleExtents{blocks, 1u}, ScaleStrides{BLOCK_BYTES / 2, 1u}}};

auto nibbleSpan = ct::tensor_span{
    reinterpret_cast<const signed char *>(raw + 2),
    ct::layout_strided_mapping<NibbleExtents, NibbleStrides>{
        NibbleExtents{blocks, NIBBLE_BYTES}, NibbleStrides{BLOCK_BYTES, 1u}}};
```

Both feed ordinary `partition_view`s, and the dequantised weights come out **bit exact** against a host reference (`worst error 0.000000`). Verified two ways against nvcc 13.3.73 on sm_89: with static extents and strides, and with **fully dynamic** extents *and* strides — the latter being the form a JIT would emit, which is the case that actually mattered for the claim.

### So where is the gap

In **`TileContext.view`**, not in the tile model. Our `view` builds a contiguous, zero-offset view of a single element type; CUDA Tile's `tensor_span` accepts `layout_strided` and `layout_right_padded` and takes any base pointer. Strided and byte-offset views are now recorded in `docs/source/tile-api.rst` as a known limitation of this API, and the example's javadoc says the separate scales array is a *choice* rather than a necessity (`2861d50`).

No NVIDIA-side feature is needed. I'd rather correct this here than leave an incorrect limitation standing in the thread — particularly one framed as an ask on NVIDIA.

cc @orionpapadakis @kotselidis, since you were tagged on the measurement comment that carried the wrong claim.

mikepapadim:
## Follow-up: strided views implemented, so the GGUF layout is now readable from Java

The [correction above](https://github.com/beehive-lab/TornadoVM/pull/1083#issuecomment-5677289907) established that CUDA Tile expresses an interleaved quantised layout fine and that the restriction was in `TileContext.view`. That gap is now closed (`ecf9e5a`... see `feat/cutile` head).

`viewStrided(array, elementOffset, rows, columns, rowStride)` lowers to a `ct::tensor_span` carrying a `ct::layout_strided_mapping` with dynamic extents and stride. So a GGUF Q4_0 block is now two views over **one allocation**:

```java
// scales: one fp16 per 18 bytes, i.e. every ninth half
PartitionView scaleView  = tc.partition(tc.viewStrided(asHalf,  0, blocks,  1,  9), BLOCKS, 1);
// nibbles: 16 usable bytes of every 18, starting at byte 2
PartitionView nibbleView = tc.partition(tc.viewStrided(asBytes, 2, blocks, 16, 18), BLOCKS, 16);
```

Verified with `--printKernel` that both actually emit the strided mapping rather than falling back to plain extents:

```cpp
auto tview_3 = ct::partition_view{ct::tensor_span{(reinterpret_cast<__half *>(ul_0 + 16) + 0),
    ct::layout_strided_mapping<ct::extents<uint32_t, ct::dynamic_extent, ct::dynamic_extent>, ...>{
        ...{(uint32_t) 8, (uint32_t) 1}, ...{(uint32_t) 9, 1u}}}, ct::shape{8_ic, 1_ic}};
```

`TestTileStridedViews` (3 tests): a row pitch wider than the row, with **poison values in the padding** so ignoring the stride lands in the output; a view starting 37 elements in; and dequantising real GGUF-layout blocks from a single interleaved buffer, checking every weight against the fp16 scale the device actually read.

One deliberate choice: **a strided view is not marked `assume_aligned`.** Its rows are not 16-byte aligned in general, and that hint is a claim the tile compiler acts on. Contiguous views still carry it.

The remaining view gap is now narrowed to what is genuinely absent rather than what I'd assumed: `layout_right_padded`, a non-unit column stride, and strided rank-3 views.

Tile suites: **116 tests, all passing.**

mikepapadim:
## GGUF read in its native layout — and a correction to the 236× I posted

With `viewStrided` in place, `TileQuantizedProjection` now reads the interleaved Q4_0 buffer **as GGUF writes it**: one allocation, viewed twice — halves for the scales, bytes for the nibbles, different offsets, block pitch as the row stride. The separate scales array is gone, and so is the claim that the layout was unreachable.

### The speedup I reported was partly an artefact of my own baseline

Kernel time from nsys, m = n = 1024, k = 2048, RTX 4090:

| | split scales (before) | native GGUF (now) |
|---|---|---|
| `@Parallel`, dequantise per element | 20 518 µs | **4 002 µs** |
| `TileContext`, dequantise in tiles | 87 µs | **125 µs** |
| ratio | **236×** | **32×** |

**Both ends moved, and the baseline moved more.** The thread-level kernel got five times faster purely from the layout: reading the interleaved block lets it fetch a scale once per 32 weights instead of once per output element. So the 236× in [my earlier comment](https://github.com/beehive-lab/TornadoVM/pull/1083#issuecomment-5677039382) was measuring a baseline penalised by a layout choice **I had imposed on it**, not the tile kernel's advantage. **32× is the honest figure** for the same work over the same bytes.

### Reading the native layout costs the tile kernel 44%

87 µs → 125 µs, for two reasons, both legitimate:

- a strided view carries no `assume_aligned` — correctly, since its rows are not 16-byte aligned — so the loads lose that hint;
- the nibble tile is 16 wide rather than 32, so a block is one `mma` on a concatenated tile instead of two on contiguous halves (4× `HMMA.16816.F32` rather than 8×).

That is a tradeoff, not a regression: **a preparation pass that splits the scales out is still worth ~1.4× on this kernel.** The difference is that this is now a measured choice rather than a limitation of the API — which is the outcome I was after when I went looking for the "missing" feature in the first place.

### One thing worth passing on

My first attempt at the indexing was wrong, and arithmetic caught it before the compiler did: for a tile of 32 output columns at one k-block, the blocks needed stride by `blocksPerColumn` rows, not by one, so no single rank-2 view can gather them as a contiguous row range. What works is folding the k-block into the **offset** (a runtime value) and letting the row pitch step one output column at a time:

```java
PartitionView scaleView = tc.partition(
        tc.viewStrided(packedHalves, blockStep * (BLOCK_BYTES / 2), n, 1, blocksPerColumn * (BLOCK_BYTES / 2)), TILE, 1);
PartitionView nibbleView = tc.partition(
        tc.viewStrided(packedBytes, 2 + blockStep * BLOCK_BYTES, n, NIBBLE_BYTES, blocksPerColumn * BLOCK_BYTES), TILE, NIBBLE_BYTES);
```

Anyone porting a blocked quantised format will hit the same shape: the pitch steps the dimension you tile over, and the block index belongs in the offset.

cc @orionpapadakis @kotselidis — this supersedes the 236× figure in the comment you were tagged on.

mikepapadim:
## Java in, CUDA Tile C++ out — three kernels side by side

Three kernels now carry their own `--printKernel` output in a javadoc comment next to the Java
that produced it, so the lowering can be read without running anything. All snippets below are
verbatim, elided only where noted.

### 1. Row softmax — `TestTileRowKernels#softmaxRow`

```java
PartitionView iv = tc.partition(tc.view(in, rows, n), 1, ROW_TILE);
PartitionView ov = tc.partition(tc.view(out, rows, n), 1, ROW_TILE);
int row = tc.bidX();

Tile x = iv.loadMasked(row, 0);
Tile shifted = tc.sub(x, tc.max(x, 1));
Tile weights = tc.exp(shifted);
ov.storeMasked(tc.div(weights, tc.sum(weights, 1)), row, 0);
```

```cpp
extern "C" __tile_global__ void softmaxRow(<4 reserved>, unsigned char *arg1,
                                           unsigned char *arg2, int arg3, int arg4)
{
  unsigned long long ul_0, ul_1;
  int i_4;

  // BLOCK 0
  ul_0  =  (unsigned long long) arg1;
  ul_1  =  (unsigned long long) arg2;
  auto tview_2 = ct::partition_view{ct::tensor_span{ct::assume_aligned(
      reinterpret_cast<float *>(ul_0 + 16), 16_ic), ct::extents{256, 256}}, ct::shape{1_ic, 2048_ic}};
  auto tview_3 = ct::partition_view{ct::tensor_span{ct::assume_aligned(
      reinterpret_cast<float *>(ul_1 + 16), 16_ic), ct::extents{256, 256}}, ct::shape{1_ic, 2048_ic}};
  i_4 = ct::bid().x;
  auto tile_5 = tview_2.load_masked(i_4, 0);
  auto tile_6 = ct::reduce_max(tile_5, 1_ic);
  auto tile_7 = tile_5 - tile_6;
  auto tile_8 = ct::exp(tile_7);
  auto tile_9 = ct::sum(tile_8, 1_ic);
  auto tile_10 = tile_8 / tile_9;
  tview_3.store_masked(tile_10, i_4, 0);
  return;
}
```

Worth reading off it: the reductions land as `ct::reduce_max` / `ct::sum` with the axis as an
`_ic` compile-time literal, and their `[1, 1]` results are then used **directly** in
`tile_5 - tile_6` and `tile_8 / tile_9` — CUDA Tile broadcasts a reduced dimension back inside
an operator, so `sub` and `div` need no explicit broadcast. The `16` added to each buffer base
is the header offset, and `assume_aligned` on it is what lets the tile compiler consider TMA.
No inline PTX and no `mma.sync` anywhere, which is the constraint this whole path exists to
honour.

### 2. `matmul` beside `mma` — `TestTileOpLevel#opMatmulHalf`

```java
ov.store(tc.cast(tc.matmul(av.load(0, 0), bv.load(0, 0)), DType.F32), 0, 0);
```

```cpp
auto tile_6 = tview_3.load(0, 0);
auto tile_7 = tview_4.load(0, 0);
auto tile_8 = ct::matmul(tile_6, tile_7);
auto tile_9 = ct::element_cast<float>(tile_8);
tview_5.store(tile_9, 0, 0);
```

The corresponding two lines from `opMma`, same operand shapes and types:

```cpp
ct::tile<float, ct::shape<32, 32>> tile_8 = ct::full<ct::tile<float, ct::shape<32, 32>>>(0.0f);
auto tile_9 = ct::mma(tile_6, tile_7, tile_8);
```

That contrast is the reason `matmul` was worth adding as a separate operation rather than
leaving callers to use `mma` against a zero tile. For `mma` the accumulator is a declared fp32
tile and it is what fixes the result type. For `matmul` there is no accumulator, the result type
is inferred from the operands — fp16 here — and the `ct::element_cast<float>` is the kernel's
own `cast`, which is the proof in the emitted source that the product really was the narrow
type and had to be widened on the way out. `testMatmulHalf` asserts exactly that from the Java
side: rounding each result back to fp16 has to be a no-op.

### 3. A loop-carried tile — `TestTileRecurrence#gatedDeltaRule`

TileGym's `recurrent_gated_delta_rule`, at 32 steps. Seven view constructions and the reserved
ABI parameters elided:

```cpp
extern "C" __tile_global__ void gatedDeltaRule(<4 reserved>, <7 buffers>, int arg8)
{
  unsigned long long ul_0, ul_1, ul_2, ul_3, ul_4, ul_5, ul_6;
  int i_16, i_40;
  ct::tile<float, ct::shape<32, 32>> ul_15;        // <- the loop-carried state

  // BLOCK 0
  ... seven partition_view constructions ...
  ct::tile<float, ct::shape<32, 32>> tile_14 = ct::full<ct::tile<float, ct::shape<32, 32>>>(0.0f);
  // BLOCK 1 MERGES [0 2 ]
  ul_15  =  tile_14;
  i_16  =  0;
  for(;i_16 < 32;)
  {
    // BLOCK 2
    auto tile_17 = tview_10.load(i_16, 0);                 // the gate, a [1, 1] tile
    auto tile_18 = ct::exp(tile_17);                       // exp stays on the device
    auto tile_19 = ct::broadcast(tile_18, ct::shape<32, 32>{});
    auto tile_20 = ul_15 * tile_19;                        // state = state * decay
    auto tile_21 = tview_7.load(i_16, 0);                  // a key row
    auto tile_22 = ct::reshape(tile_21, ct::shape<32, 1>{});
    auto tile_23 = ct::broadcast(tile_22, ct::shape<32, 32>{});
    auto tile_24 = tile_20 * tile_23;
    auto tile_25 = ct::sum(tile_24, 0_ic);                 // what the state predicts
    auto tile_26 = tview_11.load(i_16, 0);                 // beta
    auto tile_27 = ct::broadcast(tile_26, ct::shape<1, 32>{});
    auto tile_28 = tview_9.load(i_16, 0);                  // the value row
    auto tile_29 = tile_28 - tile_25;
    auto tile_30 = tile_29 * tile_27;                      // delta
    auto tile_31 = ct::broadcast(tile_22, ct::shape<32, 32>{});
    auto tile_32 = ct::broadcast(tile_30, ct::shape<32, 32>{});
    auto tile_33 = tile_31 * tile_32;                      // the rank-1 update
    auto tile_34 = tile_20 + tile_33;                      // the new state
    auto tile_35 = tview_8.load(i_16, 0);                  // a query row
    auto tile_36 = ct::reshape(tile_35, ct::shape<32, 1>{});
    auto tile_37 = ct::broadcast(tile_36, ct::shape<32, 32>{});
    auto tile_38 = tile_34 * tile_37;
    auto tile_39 = ct::sum(tile_38, 0_ic);                 // the readout
    tview_12.store(tile_39, i_16, 0);
    i_40  =  i_16 + 1;
    ul_15  =  tile_34;                                     // <- the state travels round
    i_16  =  i_40;
  }  // B1
  // BLOCK 3
  tview_13.store(ul_15, 0, 0);
  return;
}
```

The time loop survives as a real backward branch, and the line to look at is the declaration of
`ul_15`. Every other tile in the body is an `auto` bound once, because the emitter is walking
SSA values. The state is the one value that merges from two predecessors, so it becomes a named,
explicitly typed `ct::tile<float, ct::shape<32, 32>>` at function scope with an assignment at
the latch. **That declaration is the phi node**, and it is why a loop-carried tile needs the
declared-type side table on the compilation result rather than `auto` — the part of the backend
this kernel exercises that no straight-line kernel does.

Two smaller things: `steps` has been constant-folded into both the loop bound and the view
extents, so a different step count is a different compilation. And `ct::broadcast(tile_22, ...)`
is emitted twice, at `tile_23` and `tile_31` — the same Java expression appears twice in the
source and nothing is common-subexpression-eliminated at this level, which leaves the tile
compiler to decide whether that is one broadcast or two.

---

Also in this push: `ct::matmul` is now exposed (`TileContext.matmul`), which takes coverage of
CUDA 13.3's 84 `__tile_builtin__` operations from 70 to 71. The remaining 13 are the ten
`_masked` atomic variants, `atomic_compare_exchange` in both forms, and `permute`.

kotselidis (approved):

PR #1091, opened 2026-09-16
Fix signed integer conversion of unsigned intermediates

### LINKED ISSUE #1087: [CUDA] Integer subtraction after masked byte shift wraps before float conversion
### Describe the bug

For `int q = (packed.get(i) & 0xFF) >> 4 & 0xF`, storing `q - 8` into a `FloatArray` yields a large positive value for nibbles below eight. The generated unsigned temporary changes Java signed arithmetic semantics.

Probe exits 1:
```text
java.lang.AssertionError: nibble 0: expected -8.0, actual 4.2949673E9
```
Control converts to float before subtracting, exits 0, and prints `PASS` for all 16 values.

Generated CUDA:
```c
unsigned int ui_10, ui_11, ui_12;
ui_10 = ch_6 & 255U;
ui_11 = ui_10 >> 4;
ui_12 = ui_11 + -8;
f_13 = (float) ui_12;
```
The unsigned subtraction wraps before conversion to float.

### How To Reproduce

```bash
cd tornadovm-repro
```

**Compile and run**

Save the complete source below as `UnsignedNibbleRecentringRepro.java` in any empty directory. Set `JAVA_HOME` to JDK 21 and `TORNADOVM_HOME` to the built CUDA SDK (the directory containing `bin/tornado`). Run from the directory containing the saved file:

```bash
repro_dir=$(mktemp -d)
  -cp "$TORNADOVM_HOME/share/java/tornado/*" -d "$repro_dir" UnsignedNibbleRecentringRepro.java

  --classpath "$repro_dir" UnsignedNibbleRecentringRepro

  --classpath "$repro_dir" UnsignedNibbleRecentringRepro control
```

```java
import uk.ac.manchester.tornado.api.types.arrays.ByteArray;


public class UnsignedNibbleRecentringRepro {


    public static void recentre(KernelContext ctx, ByteArray packed, FloatArray out) {
        int i = ctx.globalIdx;
        int q = (packed.get(i) & 0xFF) >> 4 & 0xF;
        out.set(i, q - 8);
    }

    public static void recentreAsFloat(KernelContext ctx, ByteArray packed, FloatArray out) {
        int i = ctx.globalIdx;
        int q = (packed.get(i) & 0xFF) >> 4 & 0xF;
        out.set(i, (float) q - 8.0f);
    }


        boolean control = args.length > 0 && args[0].equals("control");
        ByteArray packed = new ByteArray(16);
        for (int i = 0; i < 16; i++) {
            packed.set(i, (byte) (i << 4));
        }
        FloatArray out = new FloatArray(16);
        out.init(0.0f);

        TaskGraph graph =
                new TaskGraph("nibble")
                        .transferToDevice(DataTransferMode.EVERY_EXECUTION, packed, out)
                        .task(
                                "k",
                                control ? UnsignedNibbleRecentringRepro::recentreAsFloat : UnsignedNibbleRecentringRepro::recentre,
                                new KernelContext(),
                                packed,
                                out)
                        .transferToHost(DataTransferMode.EVERY_EXECUTION, out);
        WorkerGrid1D worker = new WorkerGrid1D(16);
        worker.setLocalWork(16, 1, 1);
        scheduler.addWorkerGrid("nibble.k", worker);
        }

        for (int i = 0; i < 16; i++) {
            assertEquals("nibble " + i, (float) (i - 8), out.get(i), 0.0f);
        }
        System.out.println("PASS");
    }
    private static void assertEquals(String label, float expected, float actual, float tolerance) {
        if (!Float.isFinite(actual) || Math.abs(expected - actual) > tolerance) {
            throw new AssertionError(label + ": expected " + expected + ", actual " + actual);
        }
    }
}
```

### Expected behavior

For high nibbles 0 through 15, the output should be exactly -8 through 7. All eight values below eight must remain negative.

**Acceptance criteria**

- The unchanged probe and floating-subtraction control both exit 0 for every nibble 0 through 15, producing exactly -8 through 7.
- Generated code must preserve signed subtraction before float conversion; an unsigned nonnegative intermediate must not cause wraparound in the result.
- Include regression variants for high/low nibble extraction and conversion across zero. Do not force every safe unsigned intermediate to signed merely to hide this one failure.

### Computing system setup

### Additional context

**Requested change**

Fix CUDA signed arithmetic/conversion lowering for the masked-byte expression in the reproducer. Preserve the original Java expression in the regression test; moving subtraction after float conversion only works around the compiler defect.

Write `(float) q - 8.0f` as a workaround. Preserve Java signed subtraction and conversion semantics in lowering. Nonnegative intermediates may use unsigned representations, but that representation must not alter a later result that crosses zero.

orionpapadakis:
#### Description

Fixes #1087.

`emitZeroExtend` gives the result of a mask such as `packed.get(i) & 0xFF` an **unsigned** LIR kind, which is a fine representation for the non-negative intermediate itself. The problem is that the unsignedness survives the signed Java arithmetic applied afterwards: `q - 8` keeps the `unsigned int` kind while the bit pattern goes negative, and the conversion that consumes it then reads `0xfffffff8` as an unsigned quantity.

```c
ui_10 = ch_6 & 255U;
ui_11 = ui_10 >> 4;
ui_12 = ui_11 + -8;
f_13  = (float) ui_12;   // 4.2949673E9 instead of -8.0
```

`+`, `-` and `*` produce the same bit pattern whether the operands are signed or unsigned, so the value held in `ui_12` is already correct — only the operations whose C semantics depend on signedness misread it. This patch fixes those consumers rather than the representation:

* `CastNode.generate` reinterprets the operand through the signed C type of the Java source width for the signed integer-to-floating-point conversions (`I2F`/`I2D` through `(int)`, `L2F`/`L2D` through `(long)`).
* `emitSignExtend` (and the `I2D` case of `emitFloatConvert`) does the same before widening, so `long r = (b & 0xFF) - 128` no longer yields `4294967168`.

Operands that already carry a signed kind are emitted unchanged, so safe unsigned intermediates keep their unsigned representation. `ui_10` and `ui_11` above are untouched and only the conversion gains a cast:

```c
ui_10 = ch_6 & 255U;
ui_11 = ui_10 >> 4;
ui_12 = ui_11 + -8;
f_13  = (float) (int) ui_12;
```

The same defect exists verbatim in the OpenCL and Metal backends, so the fix is mirrored there.

**Regression tests.** Five cases added to `numpromotion.TestZeroExtend`, each covering all 256 byte patterns and validated against the sequential Java reference:

* `testHighNibbleRecentre` — the reporter's expression `((a & 0xFF) >> 4 & 0xF) - 8` stored into a `FloatArray`.
* `testLowNibbleRecentre` — the low-nibble variant.
* `testHighNibbleRecentreAsFloat` — the report's `(float) q - 8.0f` workaround, so it keeps working.
* `testByteRecentreToDouble` / `testByteRecentreToLong` — conversion across zero to `double` and to `long`.

Three of the five fail on `develop` and pass with this patch. The existing `testByte`/`testShort`/`testInt` cases guard the unsigned intermediates that must stay unsigned.

#### Problem description

For `int q = (packed.get(i) & 0xFF) >> 4 & 0xF`, storing `q - 8` into a `FloatArray` produced a large positive value for every nibble below eight, because the unsigned temporary reached the `(float)` cast unchanged. The reproducer in #1087 fails with `nibble 0: expected -8.0, actual 4.2949673E9`; it now exits 0 for both the probe and the control.

#### Backend/s tested

- [x] OpenCL
- [x] CUDA
- [ ] Metal (change mirrored, but no macOS hardware available here)

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=opencl,cuda
tornado-test uk.ac.manchester.tornado.unittests.numpromotion.TestZeroExtend -V
```

Validated on an RTX 5070 Ti (driver 580.142, CUDA 12.0, JDK 21.0.2). `tornado-test --quickPass` was run before and after the patch on both a CUDA-only and an OpenCL+CUDA build: the failure set is identical apart from the three new tests flipping to pass (the remaining failures are pre-existing on this machine).


tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/graal/lir/MetalArithmeticTool.java
```diff
@@ -146,7 +146,7 @@ public Value emitFloatConvert(FloatConvert floatConvert, Value input) {
         Logger.traceBuildLIR(Logger.BACKEND.Metal, "emitFloatConvert: (%s) %s", floatConvert, input);
         switch (floatConvert) {
             case I2D:
-                return emitUnaryAssign(MetalUnaryOp.CAST_TO_DOUBLE, LIRKind.value(MetalKind.DOUBLE), input);
+                return emitUnaryAssign(MetalUnaryOp.CAST_TO_DOUBLE, LIRKind.value(MetalKind.DOUBLE), asSignedOperand(input, 32));
```
mikepapadim: why this magic number 32?

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDAArithmeticTool.java
```diff
@@ -144,7 +144,7 @@ public Value emitFloatConvert(FloatConvert floatConvert, Value input) {
         Logger.traceBuildLIR(Logger.BACKEND.OpenCL, "emitFloatConvert: (%s) %s", floatConvert, input);
         switch (floatConvert) {
             case I2D:
-                return emitUnaryAssign(CUDAUnaryOp.CAST_TO_DOUBLE, LIRKind.value(CUDAKind.DOUBLE), input);
+                return emitUnaryAssign(CUDAUnaryOp.CAST_TO_DOUBLE, LIRKind.value(CUDAKind.DOUBLE), asSignedOperand(input, 32));
```
mikepapadim: same magic number constant

tornado-drivers/metal/src/main/java/uk/ac/manchester/tornado/drivers/metal/graal/lir/MetalArithmeticTool.java
orionpapadakis: its indeed magic number. now fixed. thanks

mikepapadim (approved):

stratika (approved):

PR #1093, opened 2026-09-16
[cuda] Include cuda_fp16.h for kernels whose only fp16 use is a shared half tile

### LINKED ISSUE #1089: [CUDA] Bare shared half declarations omit cuda_fp16.h and fail NVRTC compilation
### Describe the bug

A kernel using `allocateHalfFloatLocalArray` and a runtime-valued swizzled half store emits bare `half` declarations/casts without `cuda_fp16.h`. NVRTC rejects it. Adding an otherwise unrelated loaded-half-to-float conversion triggers the header and makes the kernel compile.

Probe exits 1 with NVRTC diagnostics:
```text
error: identifier "half" is undefined
    __shared__ half half_4[128];
error: identifier "half" is undefined
    ((half *) half_4)[__bo >> 1] = f_28;
```
Control exits 0, checks all outputs, and prints `with fp16 spelling: out[0] = 9.0`.

At the tested revision, the source gate checks `__half`, `half2`, or `2half`; bare `half` does not match. The control emits `__half2float`, which matches. The probe deliberately uses a runtime-computed value: a constant can emit `__float2half` and hide the bug.

### How To Reproduce

```bash
cd tornadovm-repro
```

**Compile and run**

Save the complete source below as `MissingFp16IncludeRepro.java` in any empty directory. Set `JAVA_HOME` to JDK 21 and `TORNADOVM_HOME` to the built CUDA SDK (the directory containing `bin/tornado`). Run from the directory containing the saved file:

```bash
repro_dir=$(mktemp -d)
  -cp "$TORNADOVM_HOME/share/java/tornado/*" -d "$repro_dir" MissingFp16IncludeRepro.java

  --classpath "$repro_dir" MissingFp16IncludeRepro

  --classpath "$repro_dir" MissingFp16IncludeRepro trigger
```

```java
import uk.ac.manchester.tornado.api.enums.MMAShape;
import uk.ac.manchester.tornado.api.types.HalfFloat;
import uk.ac.manchester.tornado.api.types.arrays.HalfFloatArray;


public final class MissingFp16IncludeRepro {

    private static final int M = 16;
    private static final int N = 8;
    private static final int K = 16;
    private static final int LOCAL = 32;

    private MissingFp16IncludeRepro() {}


    public static void halfTileNoSpelling(
            KernelContext ctx, HalfFloatArray a, FloatArray scale, FloatArray out) {
        int lane = ctx.localIdx;
        int[] aTile = ctx.allocateIntLocalArray(M * K / 2);
        HalfFloat[] bTile = ctx.allocateHalfFloatLocalArray(N * K);

        // A: filled by cp.async, so no half is ever read into a register here.
        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * LOCAL;
            int row = i >>> 3;
            int kk = (i & 7) << 1;
            ctx.asyncCopyToLocal(aTile, i, a, row * K + kk);
        }

        // B: written from a value computed at run time, so the store lowers to a bare
        // ((half *) tile)[...] = <float> assignment. A compile-time constant would instead emit
        // __float2half(...), which is one of the spellings the include scan looks for -- that is
        // the difference between reproducing and not.
        for (int t = 0; t < 4; t++) {
            int row = (lane & 3) * 4 + t;
            int col = lane >> 2;
            ctx.mmaStoreBSwizzled(bTile, row, col, N, new HalfFloat(scale.get(0) + t), 0);
        }

        ctx.asyncCopyCommit();
        ctx.asyncCopyWaitGroup(0);
        ctx.localBarrier();

        float[] acc = ctx.mmaFragment(0.0f);
        acc =
                ctx.mma(
                        ctx.mmaLoadA(aTile, K, 0),
                        ctx.mmaLoadBSwizzled(bTile, K, 0),
                        acc,
                        MMAShape.M16N8K16);
        ctx.mmaStore(acc, out, 0, 0, N);
    }


    public static void halfTileWithSpelling(
            KernelContext ctx, HalfFloatArray a, FloatArray scale, HalfFloatArray weights,
            FloatArray out) {
        int lane = ctx.localIdx;
        int[] aTile = ctx.allocateIntLocalArray(M * K / 2);
        HalfFloat[] bTile = ctx.allocateHalfFloatLocalArray(N * K);

        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * LOCAL;
            int row = i >>> 3;
            int kk = (i & 7) << 1;
            ctx.asyncCopyToLocal(aTile, i, a, row * K + kk);
        }
        for (int t = 0; t < 4; t++) {
            int row = (lane & 3) * 4 + t;
            int col = lane >> 2;
            ctx.mmaStoreBSwizzled(bTile, row, col, N, new HalfFloat(scale.get(0) + t), 0);
        }

        ctx.asyncCopyCommit();
        ctx.asyncCopyWaitGroup(0);
        ctx.localBarrier();

        float[] acc = ctx.mmaFragment(0.0f);
        acc =
                ctx.mma(
                        ctx.mmaLoadA(aTile, K, 0),
                        ctx.mmaLoadBSwizzled(bTile, K, 0),
                        acc,
                        MMAShape.M16N8K16);
        ctx.mmaStore(acc, out, 0, 0, N);
        if (lane == 0) {
            out.set(0, out.get(0) + weights.get(0).getFloat32());
        }
    }

        boolean trigger = args.length > 0 && "trigger".equals(args[0]);

        HalfFloatArray a = new HalfFloatArray(M * K);
        for (int i = 0; i < M * K; i++) {
            a.set(i, new HalfFloat(0.25f));
        }
        HalfFloatArray weights = new HalfFloatArray(2);
        weights.set(0, new HalfFloat(1.0f));
        weights.set(1, new HalfFloat(1.0f));
        FloatArray scale = new FloatArray(1);
        scale.set(0, 0.5f);
        FloatArray out = new FloatArray(M * N);
        out.init(0.0f);

        TaskGraph graph = new TaskGraph("repro");
        graph.transferToDevice(DataTransferMode.EVERY_EXECUTION, a, weights, scale, out);
        if (trigger) {
            graph.task(
                    "mma",
                    MissingFp16IncludeRepro::halfTileWithSpelling,
                    new KernelContext(),
                    a,
                    scale,
                    weights,
                    out);
        } else {
            graph.task(
                    "mma",
                    MissingFp16IncludeRepro::halfTileNoSpelling,
                    new KernelContext(),
                    a,
                    scale,
                    out);
        }
        graph.transferToHost(DataTransferMode.EVERY_EXECUTION, out);

        WorkerGrid1D worker = new WorkerGrid1D(LOCAL);
        worker.setLocalWork(LOCAL, 1, 1);
        scheduler.addWorkerGrid("repro.mma", worker);

        }
        for (int i = 0; i < M * N; i++) {
            float expected = trigger && i == 0 ? 9.0f : 8.0f;
            if (out.get(i) != expected) {
                throw new AssertionError("output " + i + ": expected " + expected + ", actual " + out.get(i));
            }
        }
        System.out.printf(
                "%s: out[0] = %s%n", trigger ? "with fp16 spelling" : "no fp16 spelling", out.get(0));
    }
}
```

### Expected behavior

Both kernels should compile. The probe should produce `8.0f` for all 128 outputs. The control adds `1.0f` to element zero, so it should produce `9.0f` there and `8.0f` elsewhere.

**Acceptance criteria**

- The unchanged probe compiles and produces `8.0f` in all 128 outputs. The control compiles and produces `9.0f` at index zero and `8.0f` elsewhere.
- Generated kernels that need FP16 declarations include the required header once, including the bare-half case that currently misses the text gate.
- Add regression coverage for the runtime-valued bare-half path and the existing intrinsic-trigger path. Preserve the conditional-header compatibility rationale for kernels that do not require FP16.

### Computing system setup

### Additional context

**Requested change**

Fix FP16 header selection for CUDA emitted types and intrinsics, including bare `half` shared arrays and casts. The kernel must compile without an unrelated half-to-float operation being added to the application.

An additional half-to-float conversion avoids the trigger but does not repair header selection. Track whether generated code requires FP16 declarations/intrinsics and include the preamble accordingly. Keep unrelated kernels independent of this header: some toolkits' on-disk `cuda_fp16.hpp` references `NV_IF_ELSE_TARGET` from `<nv/target>`, which can be unavailable under `__CUDACC_RTC__`. This is the documented reason the existing preamble is conditional.

The relevant gate in `CUDACompilationResultBuilder.finish` is:

```java
if ((source.contains("__half") || source.contains("half2") || source.contains("2half")) && !source.contains("cuda_fp16.h")) {
    source = CUDAPreamble.PREAMBLE + source;
}
```

`CUDAPreamble.PREAMBLE` is `"#include <cuda_fp16.h>\n"`. Bare `half` emitted by the backend itself does not satisfy this condition. This source evidence and the passing control isolate header selection from missing SDK headers on the test machine.

mikepapadim:
Fixes #1089.

### The defect

`CUDACompilationResultBuilder#finish` decides whether to prepend `#include <cuda_fp16.h>` by scanning the generated source. The test was a substring search:

```java
if ((source.contains("__half") || source.contains("half2") || source.contains("2half")) && !source.contains("cuda_fp16.h"))
```

A kernel that allocates a `__shared__` fp16 tile and fills it from a run-time value never materialises a `__half` register. Its only fp16 constructs are the tile declaration and the pointer casts the swizzled store goes through, and the backend spelt both with cuda_fp16.h's unprefixed `half` alias — which none of those three substrings match. The kernel reached NVRTC without the header:

```text
error: identifier "half" is undefined
    __shared__ half half_4[128];
error: identifier "half" is undefined
    ((half *) half_4)[__bo >> 1] = f_28;
```

Adding any unrelated half-to-float conversion emitted `__half2float`, which the scan did match, which is why the same tile setup compiled in other kernels. A compile-time constant store hides the bug the same way (it folds to `__float2half`), so the reproducer stores a value computed at run time.

### The fix

- The scan matches **whole identifiers** instead of substrings, so the unprefixed `half` / `half2` aliases count. It moves into `CUDAPreamble`, next to the headers it gates, together with the equivalent fp8 and bf16 predicates that `finish` was open-coding.
- The emitted fp16 type is normalised to `__half` (`__shared__ __half tile[N]`, `((__half *) tile)[...]`), so generated code no longer depends on a C++-only typedef. The scan still accepts the unprefixed aliases, so the emitter and the gate cannot drift apart again.
- The include stays conditional. Some toolkits' on-disk `cuda_fp16.hpp` references `NV_IF_ELSE_TARGET` from `<nv/target>`, which is excluded under `__CUDACC_RTC__`; only kernels that genuinely need fp16 should depend on the header resolving. That rationale is preserved in `CUDAPreamble`'s javadoc.

### Tests

**`CUDAPreambleTest`** (new, host-only, runs under `mvn test` with no GPU) covers the bare shared-tile shape, the conversion-intrinsic shape that already worked, the negative cases (a kernel with no fp16; identifiers such as `half_4` and `halfway` that merely contain "half"), no double-prepend, and the fp8/bf16 gates.

**`TestMatrixMultiplicationMMACpAsync`** gains the report's two device kernels: `sharedHalfTileWithoutFp16Spelling` (the probe) and `sharedHalfTileWithFp16Spelling` (the control), checked against a host oracle.

### Validation

Before the change, on this branch's tests:

```text
Running test: testSharedHalfTileWithoutFp16Spelling ... [FAIL]
    tornado_kernel.cu(92): error: expected an expression
          ((half *) half_4)[__bo >> 1] = f_37;
Running test: testSharedHalfTileWithFp16Spelling    ... [PASS]
Test ran: 8, Failed: 1
```

After:

```text
Test ran: 8, Failed: 0, Unsupported: 0      TestMatrixMultiplicationMMACpAsync
Tests run: 6, Failures: 0, Errors: 0        CUDAPreambleTest
```

No regressions in the neighbouring suites (same before/after, including the four pre-existing unsupported swizzle configurations):

| Suite | Result |
| --- | --- |
| `TestMatrixMultiplicationMMA` | 14 passed |
| `TestMatrixMultiplicationMMABF16` | 7 passed |
| `TestMatrixMultiplicationMMAFP8` | 8 passed |
| `TestMatrixMultiplicationMMAInt8` | 9 passed |
| `test-native.sh` (incl. `TestHalf2PackedCodegen`) | Test success |


mikepapadim:
Full unit-test suite on this branch (RTX 5070 Ti, sm_120, driver 580.142, CUDA 13.0 NVRTC, JDK 25, CUDA-only SDK): 199 suites, ~20 min.

The suites that report failures do so identically on a pristine `develop` (`ec970e26d`) build:

| --- | --- | --- |
| `TestProfiler` | 3 failed (`testProfilerFromExecutionPlan`, `testProfilerEnabled`, `testProfilerOnAndOff`) | same 3 |
| `TestMemoryLeak` | 3 failed | 2 failed (the third passes on re-run; order-dependent) |
| `TestCuBlasLt` | 1 failed in-suite, 0 on re-run | 9 failed on re-run, 0 in-suite |

None are in the code this change touches, and no suite fails here that passes on develop. Note that `tornado-test` exits 0 and writes `OK` to `.unittestingStatus` even with these failures, so the per-suite lines are what the table is built from.

The fp16 and MMA suites in the PR description are the directly relevant evidence, and those are green.

tornado-drivers/cuda/src/test/java/uk/ac/manchester/tornado/drivers/cuda/tests/CUDAPreambleTest.java
```diff
@@ -0,0 +1,111 @@
...
+ * Header selection for generated CUDA C.
+ *
+ * <p>These are the kernel shapes that decide whether {@code CUDACompilationResultBuilder#finish}
+ * prepends an optional include. The case that matters most is a kernel whose only fp16
+ * construct is a {@code __shared__} tile plus the pointer casts around it: it carries no
+ * {@code __half}-typed operand and no conversion intrinsic, so an include scan that looks for
+ * {@code __half} / {@code half2} / {@code 2half} as substrings misses it and NVRTC rejects the
+ * kernel with {@code identifier "half" is undefined}.
+ */
+public class CUDAPreambleTest {
```
stratika: why is this test class in the backend?

mikepapadim: Fair point — moved it out of `graal.backend` in eb8dc59. It now sits in `tornado-drivers/cuda/src/test/java/.../drivers/cuda/tests/CUDAPreambleTest.java`, the `tests` package name this backend already uses for its other test classes.

On why it stays in the driver module rather than `tornado-unittests`: it asserts on `CUDAPreamble` directly, and `tornado-unittests` has no dependency on the driver modules — no test there imports a driver class — so moving it would mean adding one. The module already carries host-side unit tests (`CUDAGlobalLoadBatchingTest` and its invariants test, added with the load-batching pass) and surefire runs them under `mvn test`, so this introduces nothing new. It needs no GPU, which is the point: it pins the include-scan behaviour in CI on machines with no CUDA device, while `TestMatrixMultiplicationMMACpAsync` covers the same fix end-to-end on hardware.

Happy to move it again if you would rather it lived somewhere else — or drop it and rely on the device tests alone.

mikepapadim: Reworked: the test is gone rather than relocated, and the branch is squashed back to a single commit.

Putting it in `tornado-unittests` is the right instinct and I tried that first — it does not work as written. The test asserts on `CUDAPreamble`, `tornado-unittests` has no dependency on the driver modules (nothing there imports a driver class today), and `tornado-drivers/cuda` is built only under the `cuda-backend` profile, so adding that dependency would break an OpenCL-only or Metal-only build. The alternative is profile-conditional source roots for one class, which is more machinery than the test is worth.

So coverage for this fix now lives entirely in `tornado-unittests`: `TestMatrixMultiplicationMMACpAsync` carries the report's probe kernel — which fails to compile without the change — and its control. What is lost is the guard on a machine with no CUDA device. If you would like that back, say so and I will add the profile plumbing to host an equivalent test in `tornado-unittests`.

stratika (approved):

tornado-drivers/cuda/src/test/java/uk/ac/manchester/tornado/drivers/cuda/tests/CUDAPreambleTest.java
mikepapadim: Better answer than my last one, thanks for pushing on it.

`TestHalf2PackedCodegen` — in this backend's `tests` package, run by `bin/test-native.sh` — already compiles kernels through the real backend and asserts on the generated CUDA C, `#include <cuda_fp16.h>` included. So the check now lives there as one more kernel, with no new test class anywhere: a `__shared__` fp16 tile written from a run-time value, which is exactly the shape the include scan missed.

It is a real guard, not a restatement. With the four production files reverted to `develop`:

```text
[FAIL] kernel sharedHalfTileFromComputedValue is missing "#include <cuda_fp16.h>"
[FAIL] kernel sharedHalfTileFromComputedValue is missing "__shared__ __half"
[FAIL] kernel sharedHalfTileFromComputedValue is missing "((__half *)"
Test failed
```

and `Test success` with them.

It needs a CUDA device like the rest of that file, so the no-GPU coverage the deleted unit test gave is still gone — but the assertion on the generated source is back, in a place that already does exactly this.

orionpapadakis (approved):

orionpapadakis (approved):
Reviewed the compiler changes and found no blocking issues. The shared FP16 declarations and swizzled pointer casts consistently use __half, and the header predicates recognize the emitted FP16, FP8, and BF16 constructs while preserving conditional inclusion.

Validation used an isolated local CUDA SDK with this revision's changed compiler classes and tests, on an RTX 5090 with driver 580.142, CUDA 13.0, and JDK 21; CPU fallback was disabled.

- GPU suites: TestMatrixMultiplicationMMACpAsync (including both new shared-tile cases), TestMatrixMultiplicationMMABF16, TestMatrixMultiplicationMMAFP8, TestHalf2Packed, TestHalfFloatLocalArray, and TestSwizzledLocalArrays. Across these suites, 33 tests passed and 4 explicitly rejected the CUDA backend through existing unsupported-test guards. JUnit reports those unsupported cases as failures; there were no unexpected failures.
- Updated TestHalf2PackedCodegen: passed, including the new shared-tile header, declaration, and cast assertions.
- Additional local host checks: 43 header-predicate assertions passed, covering aliases, intrinsics, identifier boundaries, existing includes, and FP8/BF16 gates.

PR #1094, opened 2026-09-16
[cuda] Support constant-index reads of an MMA accumulator fragment

### LINKED ISSUE #1090: [CUDA] Constant-index reads of an FP16 MMA accumulator cause an internal compiler error
### Describe the bug

`KernelContext.mma(...)` returns a `float[]`, but reading its four elements at constant indices fails in CUDA address lowering. The same tile setup succeeds when the fragment is consumed through `mmaStore`. This report demonstrates FP16 MMA result reads only, not integer MMA, dynamic indexing, or fragment writes.

Probe exits 1:
```text
TornadoInternalError: unimplemented: address origin unimplemented: uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAMMAComputeNode
```
Control uses `mmaStore`, exits 0, and prints `PASS`. It checks the unscaled matrix product in row-major order; the probe checks twice that product in lane-fragment order. The control establishes that the shared-tile setup and MMA operation themselves work.

### How To Reproduce

```bash
cd tornadovm-repro
```

**Compile and run**

Save the complete source below as `MmaFragmentAccessRepro.java` in any empty directory. Set `JAVA_HOME` to JDK 21 and `TORNADOVM_HOME` to the built CUDA SDK (the directory containing `bin/tornado`). Run from the directory containing the saved file:

```bash
repro_dir=$(mktemp -d)
  -cp "$TORNADOVM_HOME/share/java/tornado/*" -d "$repro_dir" MmaFragmentAccessRepro.java

  --classpath "$repro_dir" MmaFragmentAccessRepro

  --classpath "$repro_dir" MmaFragmentAccessRepro control
```

```java
import uk.ac.manchester.tornado.api.enums.MMAShape;
import uk.ac.manchester.tornado.api.types.HalfFloat;
import uk.ac.manchester.tornado.api.types.arrays.HalfFloatArray;


public class MmaFragmentAccessRepro {

    private static final int M = 16;
    private static final int N = 8;
    private static final int K = 16;


    public static void scaleFragment(
            KernelContext ctx, HalfFloatArray a, HalfFloatArray b, FloatArray out) {
        int lane = ctx.localIdx;
        int[] aTile = ctx.allocateIntLocalArray(M * K / 2);
        HalfFloat[] bTile = ctx.allocateHalfFloatLocalArray(N * K);

        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * 32;
            int row = i >>> 3;
            int kk = (i & 7) << 1;
            int base = row * K + kk;
            aTile[i] =
                    (a.get(base).getHalfFloatValue() & 0xFFFF)
                            | ((a.get(base + 1).getHalfFloatValue() & 0xFFFF) << 16);
        }
        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * 32;
            int col = i >>> 4;
            int kk = i & 15;
            ctx.swizzleStoreFp16Stride32(bTile, kk, col, N, b.get(col * K + kk));
        }
        ctx.localBarrier();

        float[] acc = ctx.mmaFragment(0.0f);
        acc = ctx.mma(ctx.mmaLoadA(aTile, K), ctx.mmaLoadBSwizzled(bTile, K), acc, MMAShape.M16N8K16);

        // THE PROBE: constant-index READS of the accumulator, written to plain global memory.
        out.set(lane * 4, acc[0] * 2.0f);
        out.set(lane * 4 + 1, acc[1] * 2.0f);
        out.set(lane * 4 + 2, acc[2] * 2.0f);
        out.set(lane * 4 + 3, acc[3] * 2.0f);
    }

    public static void storeFragment(
            KernelContext ctx, HalfFloatArray a, HalfFloatArray b, FloatArray out) {
        int lane = ctx.localIdx;
        int[] aTile = ctx.allocateIntLocalArray(M * K / 2);
        HalfFloat[] bTile = ctx.allocateHalfFloatLocalArray(N * K);

        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * 32;
            int row = i >>> 3;
            int kk = (i & 7) << 1;
            int base = row * K + kk;
            aTile[i] =
                    (a.get(base).getHalfFloatValue() & 0xFFFF)
                            | ((a.get(base + 1).getHalfFloatValue() & 0xFFFF) << 16);
        }
        for (int slot = 0; slot < 4; slot++) {
            int i = lane + slot * 32;
            int col = i >>> 4;
            int kk = i & 15;
            ctx.swizzleStoreFp16Stride32(bTile, kk, col, N, b.get(col * K + kk));
        }
        ctx.localBarrier();

        float[] acc = ctx.mmaFragment(0.0f);
        acc = ctx.mma(ctx.mmaLoadA(aTile, K), ctx.mmaLoadBSwizzled(bTile, K), acc, MMAShape.M16N8K16);

        ctx.mmaStore(acc, out, 0, 0, N);
    }

        boolean control = args.length > 0 && args[0].equals("control");
        HalfFloatArray a = new HalfFloatArray(M * K);
        HalfFloatArray b = new HalfFloatArray(N * K);
        float[] ah = new float[M * K];
        float[] bh = new float[N * K];
        for (int i = 0; i < M * K; i++) {
            ah[i] = ((i * 7) % 13 - 6) * 0.25f;
            a.set(i, new HalfFloat(ah[i]));
        }
        for (int i = 0; i < N * K; i++) {
            bh[i] = ((i * 5) % 11 - 5) * 0.5f;
            b.set(i, new HalfFloat(bh[i]));
        }
        FloatArray out = new FloatArray(M * N);
        out.init(0.0f);

        KernelContext ctx = new KernelContext();
        TaskGraph graph =
                new TaskGraph("probe")
                        .transferToDevice(DataTransferMode.EVERY_EXECUTION, a, b, out)
                        .task("k", control ? MmaFragmentAccessRepro::storeFragment : MmaFragmentAccessRepro::scaleFragment, ctx, a, b, out)
                        .transferToHost(DataTransferMode.EVERY_EXECUTION, out);

        WorkerGrid1D worker = new WorkerGrid1D(32);
        worker.setLocalWork(32, 1, 1);
        scheduler.addWorkerGrid("probe.k", worker);
        }

        for (int lane = 0; lane < 32; lane++) {
            for (int i = 0; i < 4; i++) {
                int m = lane / 4 + 8 * (i / 2);
                int n = (lane % 4) * 2 + (i % 2);
                float expected = 0;
                for (int k = 0; k < K; k++) {
                    expected += ah[m * K + k] * bh[n * K + k];
                }
                assertEquals(
                        "lane " + lane + " elem " + i,
                        (control ? 1.0f : 2.0f) * expected,
                        out.get(control ? m * N + n : lane * 4 + i),
                        1e-2f);
            }
        }
        System.out.println("PASS");
    }
    private static void assertEquals(String label, float expected, float actual, float tolerance) {
        if (!Float.isFinite(actual) || Math.abs(expected - actual) > tolerance) {
            throw new AssertionError(label + ": expected " + expected + ", actual " + actual);
        }
    }
}
```

### Expected behavior

Either support constant-index fragment reads with documented lane mapping, or reject them with an actionable unsupported-operation diagnostic. If supported, all 128 outputs should match the CPU matrix product multiplied by two.

**Acceptance criteria**

- Preferred functional fix: the unchanged probe exits 0 and all 128 lane-fragment outputs match twice the CPU matrix product; the `mmaStore` control still passes its unscaled row-major oracle.
- Preserve the fragment's lane mapping and uniform warp participation. Add regression coverage for the four constant indices and existing `mmaStore` behavior.
- If fragment indexing is intentionally unsupported, provide an explicit API restriction and actionable compilation diagnostic with a regression test. Report that result as a diagnostic fix, not as implemented fragment access.

### Computing system setup

### Additional context

**Requested change**

Support constant-index reads of the FP16 MMA accumulator returned by `KernelContext.mma`, or establish and clearly report an intentional API restriction. The demonstrated failure concerns reads of four constant indices; integer fragments, writes, and dynamic indices are outside the demonstrated scope.

Use `mmaStore` to materialize the fragment before subsequent processing. This is a functional alternative, not a measured performance recommendation. Consider lowering constant-index accesses to register-element operations or documenting a typed fragment API; preserve warp-uniform participation requirements. Integer fragment conversion and block-wise rescaling would require separate tests.

The first relevant compiler frames are `CUDAAddressLowering.lower`, called from `AddressLoweringByNodePhase.run`. The error occurs while treating the reported value node as an address origin; this identifies the failing phase, not a proven root-cause patch location.

mikepapadim:
Fixes #1090.

### The defect

`KernelContext.mma(...)` hands the accumulator back as a `float[]` (an `int[]` on the int8/s32 path), but reading an element of it died in address lowering:

```text
TornadoInternalError: unimplemented: address origin unimplemented:
    uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAMMAComputeNode
```

A fragment is the lane's own register tuple — the code generator emits it as a C array local to the thread (`float mma_acc_f_127[4]`). It has no base pointer, so lowering `acc[0]` through the ordinary array path builds an address whose origin `CUDAAddressLowering` cannot name. A kernel that only wanted to scale the result had to push the whole fragment through `mmaStore` and read it back from memory.

### The fix

`acc[i]` is recognised while the `LoadIndexed` is lowered and emitted as the element read it actually is. For the report's probe the generated CUDA C is now:

```c
float mma_acc_f_127[4];
asm volatile("mma.sync.aligned.m16n8k16.row.col.f32.f16.f16.f32 ..." : "=f"(mma_acc_f_127[0]), ...);
f_128 = mma_acc_f_127[0];
```

The fragment-producing nodes carry a `MarkMMAFragment` marker, so the lowering sees through the Pi, proxy and phi nodes an accumulator picks up when it crosses a null check, a loop exit or a K-loop back-edge — the usual GEMM shape, where the accumulator is created before the loop, replaced by `mma` inside it, and read after it.

Three shapes cannot be served, and are now reported where the kernel source that caused them is still identifiable rather than dying later as an unnameable address origin:

| Shape | Message |
| --- | --- |
| `acc[i]` with a run-time `i` | *"An MMA accumulator fragment can only be indexed with a compile-time constant: the fragment lives in registers, and a run-time index would force it into local memory. Unroll the access, or write the fragment out with KernelContext.mmaStore and index the result."* |
| `fragA[i]` on an A/B operand fragment | *"Only the accumulator fragment returned by KernelContext.mma\*/mmaFragment\* can be indexed; the A and B fragments hold packed b32 operand lanes, whose elements do not correspond to the Java array's elements. Consume them with KernelContext.mma instead."* |
| `acc[i] = v` | *"An MMA fragment cannot be written through an index: it is a per-lane register tuple, not an array in memory. Build the accumulator with KernelContext.mmaFragment/mma, or write it out with KernelContext.mmaStore and modify the result there."* |

`KernelContext.mmaFragment` now documents the layout the reads expose: for lane `L`, element `i` is row `L / 4 + 8 * (i / 2)`, column `(L % 4) * 2 + i % 2` of the tile — the same mapping `mmaStore` uses to scatter the fragment to memory — along with the warp-uniformity requirement the fragment inherits from the `mma` that produced it.

### Tests

- **`TestMatrixMultiplicationMMA`**: `testFragmentElementReads` checks all four constant reads against a host oracle in lane-fragment order (so it pins the documented mapping, not just the values); `testFragmentStoreStillMatchesItsOracle` runs the same tiles through `mmaStore` against its row-major oracle as the control; `testDynamicFragmentIndexIsRejected` pins the run-time-index diagnostic.
- **`TestMatrixMultiplicationMMAInt8`**: `testIntFragmentElementReads` covers the s32 accumulator, `testOperandFragmentIndexIsRejected` the operand-fragment restriction.

### Validation

The report's `MmaFragmentAccessRepro`, unchanged, on an SDK built from this branch:

```text
$ tornado --jvm="-Dtornado.recover.bailout=False" --classpath . MmaFragmentAccessRepro
PASS
$ tornado --jvm="-Dtornado.recover.bailout=False" --classpath . MmaFragmentAccessRepro control
PASS
```

Suites:

| Suite | Result |
| --- | --- |
| `TestMatrixMultiplicationMMA` | 17 passed (3 new) |
| `TestMatrixMultiplicationMMAInt8` | 11 passed (2 new) |
| `TestMatrixMultiplicationMMAFP8` | 8 passed |
| `TestMatrixMultiplicationMMABF16` | 7 passed |
| `TestMatrixMultiplicationMMACpAsync` | 6 passed |
| `test-native.sh` | Test success |

### Out of scope

Indexing a `HalfFloat[]` A/B fragment still fails earlier, in `TornadoHalfFloatReplacement`, with its own message rather than the one above; that path is untouched here. Fragment writes and block-wise rescaling would need an API that can express the lane mapping, and are reported as a restriction rather than implemented.


mikepapadim:
Full unit-test suite on this branch (RTX 5070 Ti, sm_120, driver 580.142, CUDA 13.0 NVRTC, JDK 25, CUDA-only SDK):

```
$ tornado-test
Total Time(s): 2063.15
$ echo $?; cat .unittestingStatus
0
OK
```

No failures anywhere in the suite, not just the MMA and fp16 ones listed above.

mikepapadim:
Correction to my earlier comment: the full-suite run does contain failures. I read them from the exit code and `.unittestingStatus` (both "pass"), and that run's per-suite output had been truncated — `tornado-test` reports `OK` and exits 0 even when individual suites record failures, so neither signal says what I took it to say.

Here is the accurate picture. Full suite on this branch, RTX 5070 Ti (sm_120), driver 580.142, CUDA 13.0 NVRTC, JDK 25, CUDA-only SDK — 199 suites, of which these report failures:

| --- | --- | --- |
| `TestProfiler` | 3 failed (`testProfilerFromExecutionPlan`, `testProfilerEnabled`, `testProfilerOnAndOff`) | same 3 |
| `TestMemoryLeak` | 3 failed | 2 failed (the third passes on re-run; order-dependent) |
| `TestCuBlasLt` | 1 failed in-suite, 0 on re-run | 9 failed on re-run, 0 in-suite |

Every one of them reproduces on a pristine `develop` build with the same tests and the same counts; `TestMemoryLeak` and `TestCuBlasLt` swing between runs on both. None are in the code these changes touch, and no suite fails on this branch that passes on develop.

The MMA, fp16 and swizzle suites listed in the PR description remain the directly relevant evidence, and those are green.

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/CUDAMMAFragmentElementNode.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public void generate(NodeLIRBuilderTool gen) {
+        LIRGeneratorTool tool = gen.getLIRGeneratorTool();
+        Value fragmentValue = gen.operand(fragment);
+        if (!(fragmentValue.getPlatformKind() instanceof CUDAKind fragmentKind) || !fragmentKind.isMMAFragment()) {
+            throw new TornadoBailoutRuntimeException(
+                    "MMA fragment element read applied to a non-fragment value: " + fragmentValue);
+        }
+        Variable result = tool.newVariable(LIRKind.value(fragmentKind.getElementKind()));
```
orionpapadakis: [P2] Accept fragment phis before backend declaration recovery

A valid loop-carried accumulator reaches this point as an object-derived `ULONG` LIR variable, so this check rejects the Pi/proxy/phi shapes that `MarkMMAFragment` explicitly accepts. `CUDABackend.emitVariableDefs()` only recovers the fragment array declaration from incoming copies later. Both a float MMA K-loop and an int8 MMA K-loop fail here when their accumulator elements are read after the loop.

Lowering has already validated the fragment and constant index; use the element read's scalar stamp instead of the fragment operand's provisional platform kind:

```suggestion
        // Lowering has validated the fragment and index. Fragment phis retain an
        // object-derived ULONG kind until the backend recovers their declarations.
        // The read stamp already carries the scalar float/int result type.
        Variable result = tool.newVariable(tool.getLIRKind(stamp));
```

The now-unused `LIRKind`, `CUDAKind`, and `TornadoBailoutRuntimeException` imports can be removed. Add float and int tests that initialize the accumulator before a runtime-controlled K loop, update it with `mma`/`mmaInt8`, and read all four elements after the loop; cover zero and multiple iterations.

orionpapadakis (changes requested):
The direct fragment reads pass, but the advertised loop-carried accumulator support fails during LIR generation. Both float and int accumulators are affected. I reproduced the failure and tested the small fix suggested inline.

Validation used this PR's changed compiler classes and tests in an isolated CUDA SDK on an RTX 5090, driver 580.142, CUDA 13.0, JDK 21, with CPU fallback disabled.

- Existing five MMA suites: 49 tests passed.
- Added real-MMA K-loop regressions for float and int accumulator reads: both fail on the original PR with `MMA fragment element read applied to a non-fragment value: ...|ULONG`.
- After deriving the scalar result kind from the read stamp: all 51 tests pass. Each added regression checks all four elements in every lane at runtime-controlled iteration counts 0, 1, and 3.

Please include these loop-carried regression cases with the fix. There is also a small documentation correction: each lane holds four of the m16n8 tile's 128 elements, not a quarter of the tile.

orionpapadakis (approved):
The loop-carried fragment issue is resolved. The scalar result kind comes from the element stamp, while the backend retains the fragment's array representation. Commit bb7e2d381 adds float and int MMA regression coverage at runtime iteration counts 0, 1, and 3, preserving the author's preceding fix.

Validation: 51 CUDA tests passed on an NVIDIA GeForce RTX 5090 Laptop GPU with JDK 21. Approving the updated head.

orionpapadakis (commented):
The three failed OpenCL jobs at 400bbf0b7 share two non-whitelisted failures: TestMatrixMultiplicationMMA.testDynamicFragmentIndexIsRejected and TestMatrixMultiplicationMMAInt8.testOperandFragmentIndexIsRejected.

Both tests used @Test(expected = TornadoBailoutRuntimeException.class). On OpenCL, their backend guard throws TornadoVMOpenCLNotSupported; JUnit wraps that as an unexpected-exception assertion failure, so TornadoVM's runner can no longer classify it as unsupported. Metal has the same guard problem. My previous CUDA-only validation missed this test-harness regression.

Fixed in fd01ad8a9: explicit backend guards run before a scoped assertThrows. The CUDA compiler rejection remains required; these tests are not whitelisted or disabled.

Validation with the current PR's freshly compiled CUDA backend and test classes on JDK 21 / RTX 5090 Laptop GPU:
- Before: reproduced both OpenCL failures with TornadoTestRunner.
- After: OpenCL reports 30 unsupported, 0 failed; CUDA reports 30 passed, 0 failed, including both rejection tests.

All other failures in the three CI logs were already marked WHITELISTED by the runner. The CUDA CI legs on JDK 21/25/27 passed on the preceding head. The new commit is pushed; fresh CI results are pending.

Failure logs: [JDK 21](https://github.com/beehive-lab/TornadoVM/actions/runs/35387461843/job/105800359878), [JDK 25](https://github.com/beehive-lab/TornadoVM/actions/runs/35387461843/job/105800359907), [JDK 27](https://github.com/beehive-lab/TornadoVM/actions/runs/35387461843/job/105800359850).

Issue #1096, opened 2026-09-18
[CUDA] float-to-half conversion lowers only as a store: new HalfFloat(f).getHalfFloatValue() fails with "address origin unimplemented: CUDAConvertFloatToHalf"


mikepapadim:
### Summary

On the CUDA backend the float→half conversion can only be reached as a **store** into a `HalfFloatArray`. The moment the converted value is consumed as a value in the same kernel — `new HalfFloat(f).getHalfFloatValue()` — compilation aborts in `CUDAAddressLowering.lower`, which is handed the conversion node itself as an address origin:

```
uk.ac.manchester.tornado.api.exceptions.TornadoInternalError: unimplemented:
    address origin unimplemented: uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAConvertFloatToHalf
	at uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDAAddressLowering.lower(CUDAAddressLowering.java:56)
	at tornado.graal.compiler.phases.common.AddressLoweringByNodePhase.run(AddressLoweringByNodePhase.java:67)
	at uk.ac.manchester.tornado.drivers.cuda.graal.compiler.CUDACompiler.emitFrontEnd(CUDACompiler.java:185)
```

If any arithmetic precedes the conversion, the reported origin is that arithmetic instead — `address origin unimplemented: tornado.graal.compiler.nodes.calc.MulNode` — which is the same message as #1057.

### Minimal reproducer

Three tasks, no `ByteArray`, no MMA, no tiling, no shared memory. One case per JVM.

```java
import uk.ac.manchester.tornado.api.*;
import uk.ac.manchester.tornado.api.enums.DataTransferMode;
import uk.ac.manchester.tornado.api.types.HalfFloat;
import uk.ac.manchester.tornado.api.types.arrays.*;

public class HalfFloatValueRepro {

    static final int N = 256;

    /** WORKS: the conversion is the store itself. */
    public static void store(KernelContext ctx, FloatArray in, HalfFloatArray out) {
        int i = ctx.globalIdx;
        out.set(i, new HalfFloat(in.get(i) * 2.0f));
    }

    /** FAILS: address origin unimplemented: MulNode */
    public static void value(KernelContext ctx, FloatArray in, IntArray out) {
        int i = ctx.globalIdx;
        float v = in.get(i) * 2.0f;
        out.set(i, new HalfFloat(v).getHalfFloatValue() & 0xFFFF);
    }

[...]

Issue #1057, opened 2026-08-28
[CUDA] address origin unimplemented: MulNode — CUDAAddressLowering rejects a tensor-core GEMM over Q8_0 ByteArray weights (OpenCL/Metal compile the same kernel)


orionpapadakis:
### Summary

On the CUDA backend, `CUDAAddressLowering.lower` refuses a `MulNode` as an address origin and aborts compilation of a tensor-core GEMM kernel that uses `ByteArray` (Q8_0) weights:

```
Exception in thread "main" uk.ac.manchester.tornado.api.exceptions.TornadoInternalError:
    unimplemented: address origin unimplemented: tornado.graal.compiler.nodes.calc.MulNode
	at uk.ac.manchester.tornado.api.exceptions.TornadoInternalError.unimplemented(TornadoInternalError.java:41)
	at uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDAAddressLowering.lower(CUDAAddressLowering.java:56)
	at tornado.graal.compiler.phases.common.AddressLoweringByNodePhase.run(AddressLoweringByNodePhase.java:67)
	at tornado.graal.compiler.phases.BasePhase.apply(BasePhase.java:434)
	at tornado.graal.compiler.phases.PhaseSuite.run(PhaseSuite.java:390)
	at uk.ac.manchester.tornado.drivers.cuda.graal.compiler.CUDACompiler.emitFrontEnd(CUDACompiler.java:185)
	at uk.ac.manchester.tornado.drivers.cuda.graal.compiler.CUDACompiler.compile(CUDACompiler.java:125)
	at uk.ac.manchester.tornado.drivers.cuda.runtime.CUDATornadoDevice.compileTask(CUDATornadoDevice.java:271)
```

The same kernel compiles and runs correctly on the **OpenCL** and **Metal** backends, so this is CUDA-specific.

### Version

Reproduced on `develop` at **7fcd11b10** (current tip at the time of writing) and also at **803428c7c**. Built as:

```
mvn -Pjdk21,cuda-backend -Dtornado.backend=cuda install -DskipTests \
    -pl '!tornado-cudnn,!tornado-drivers/cudnn-jni'
```

### Environment

- NVIDIA GeForce RTX 5090 Laptop GPU, driver 580.142
- CUDA 13.1 (V13.1.115)
- JDK 21.0.2 (SDKMAN `21.0.2-open`), Linux x86_64

[...]

PR #1098, opened 2026-09-18
[cuda] Let a converted half be read back as a value, not only stored

mikepapadim:
Fixes #1096. Fixes #1057.

### One root cause, two reports

`HalfFloat.getHalfFloatValue()` and `getFloat32()` are intercepted by an invocation plugin for receivers that `TornadoHalfFloatReplacement` later strips down to a raw half value. The reason is written into `isSyntheticHalfReceiver`'s own javadoc: inlining their bytecode against such a receiver materialises a read of the `halfFloatValue` field over what becomes a half register, and no later pass repairs it.

A receiver that is a freshly constructed `new HalfFloat(x)` is exactly that case — the replacement phase dissolves the allocation into the value it wraps — but it was not in the list. So the call inlined, the phase rewired the allocation's usages to the converted value, and the field read reached address lowering with that value as its base:

```text
TornadoInternalError: unimplemented: address origin unimplemented:
    uk.ac.manchester.tornado.drivers.cuda.graal.nodes.CUDAConvertFloatToHalf
```

With any arithmetic in front of the conversion, the same refusal names the arithmetic instead (`...calc.MulNode`) — which is how #1057 reported it. Instrumenting `CUDAAddressLowering` on the reproducer from #1096 shows the two cases are one graph shape:

```text
value:       base=7|*                      usages=[17|OffsetAddress] offset=Constant(12)
valueOfLoad: base=11|CUDAConvertFloatToHalf usages=[16|OffsetAddress] offset=Constant(12)
```

Offset 12 is the `halfFloatValue` field. Storing the identical expression into a `HalfFloatArray` always worked, because the store never reads the object back — which is what made float-to-half reachable only as a store.

#1096 offered the shared-cause hypothesis for #1057 but could not confirm it against the source at that revision. It is confirmed: the kernel #1057 failed on had

```java
int lo = new HalfFloat(v0).getHalfFloatValue() & 0xFFFF;
int hi = new HalfFloat(v1).getHalfFloatValue() & 0xFFFF;
```

in `packQ8Halves`, replaced by a software conversion in GPULlama3.java commit `2b462d27` ("fix(cuda): compile Q8_0 tensor-core batch prefill on PTX"), whose message names the same construct and the same `MulNode` refusal. That expression is the `testPackHalfPair` case below.

### The fix

- **`CUDAHalfFloatPlugins`**: `isSyntheticHalfReceiver` also accepts a freshly constructed `new HalfFloat(...)` — both the allocation as the parser still holds it on the stack and the `NewHalfFloatInstance` the `<init>` node plugin substitutes for it. No field read is ever formed, so no address is either.

  Scoped to the **float** constructor. `new HalfFloat(short)` takes a raw bit pattern, for which this backend has no reinterpretation (`__ushort_as_half` is emitted nowhere), so intercepting it would hand the bits reader a plain short and silently produce a numerically converted value. Those receivers keep inlining exactly as before.

- **`TornadoHalfFloatReplacement`**: the constructor-argument handling converts any float-valued argument, not only a read. A computed argument used to stay in a float register — invisible while the only consumer is a store (the write casts the destination to `__half` and C++ converts on assignment), but on the read path it would reach `__half_as_ushort` as a float and depend on an implicit `cuda_fp16.h` conversion to mean anything.

Emitted CUDA C is now the two hardware instructions, in place of the ~40-operation software conversion #1057 had to fall back to in an inner K loop:

```c
half_10  =  __float2half(f_9);
ui_11 = (unsigned) __half_as_ushort(half_10);
```

and for `getFloat32()`, `__half2float(__float2half(f))`.

### Tests

`TestHalfFloatValueConversion` (new, registered in `tornado-test`): the bits of a computed value and of a loaded one — the two shapes the reports name — `getFloat32()` of a `new HalfFloat`, and two converted halves packed into one int, the tensor-core weight-staging shape from #1057. Oracles are `Float.floatToFloat16` / `Float.float16ToFloat`, exact.

All four fail on `develop` with precisely the reported errors:

```text
testBitsOfComputed        [FAILED] address origin unimplemented: ...calc.MulNode
testBitsOfLoad            [FAILED] address origin unimplemented: ...CUDAConvertFloatToHalf
testFloat32OfNewHalfFloat [FAILED]
testPackHalfPair          [FAILED] address origin unimplemented: ...calc.MulNode
Test ran: 4, Failed: 4
```

and pass with this change. The reproducer from #1096 also passes unchanged in all three of its cases (`store`, `value`, `valueOfLoad`).

### Validation

| Suite | Result |
| --- | --- |
| `TestHalfFloatValueConversion` | 4 passed (new) |
| `TestHalfFloatInlineWrite` | 3 passed |
| `TestHalfFloatInPlaceUpdate` | 4 passed |
| `foundation.TestHalfFloats` | 4 passed |
| `vectortypes.TestHalfFloats` | 26 passed, 18 unsupported (unchanged) |
| `TestHalf2Packed` | 7 passed |
| `TestHalfFloatDotProduct` | 1 passed |
| `TestHistogramHalfFloat` | 1 passed |
| `TestLocalMemoryReductionsHalfFloats` | 4 passed |
| `TestBitonicSortHalfFloat` | 1 passed |
| `TestMatrixMultiplicationMMA` / `MMACpAsync` / `MMABF16` | 14 / 6 / 7 passed |
| `codegen.TestSharedMemoryStaging` | 7 passed |
| `test-native.sh` | Test success |

A full `tornado-test` run on this branch is in progress; I will post the per-suite comparison against a pristine `develop` build as a comment. (`tornado-test` exits 0 and writes `OK` even when individual suites fail, so the comparison is built from the per-suite lines, not the exit code.)

### Scope

CUDA only. The OpenCL and Metal backends carry parallel copies of both files; #1057 reports the same kernel compiling on those backends, and I have no device here to test a change to them, so they are left alone.

**A separate defect found while testing this**, not fixed here: `new HalfFloat(short)` is miscompiled on the store path today, on `develop`, independently of this change. `out.set(i, new HalfFloat((short) bits))` writes `*(__half *) p = <short>`, which converts numerically rather than reinterpreting, so half bits `0xC800` (-8.0) are stored as -14336.0. Every element of a 32-element round trip mismatches. That needs a `__ushort_as_half` reinterpretation, which this backend does not emit anywhere; it is why the interception above is scoped to the float constructor rather than all `HalfFloat` allocations.


mairooni (approved):

mikepapadim:
Full `tornado-test` run on this branch, and the tests reworked to carry the reports' own kernels.

**Tests now cover the KernelContext form first-class.** Both reports hit this from KernelContext kernels, and that path lowers differently from an `@Parallel` loop, so the test class now runs the reproducers as they were written: `store` / `value` / `valueOfLoad` from #1096 with explicit `globalIdx` indexing, and — new — `packQ8Halves` from #1057 verbatim (34-byte Q8_0 block, fp16 block scale × int8 quant) staged through a `ctx.allocateIntLocalArray` tile, which is the tiling that report could not reduce away. The `@Parallel` variants are kept alongside.

That test is a direct in-tree reproduction of #1057. With the two production files reverted to `develop` and the tests kept:

```text
testStore                  [PASS]      <- the store path always worked; the control
testValue                  [FAILED]    address origin unimplemented: ...calc.MulNode
testValueOfLoad            [FAILED]    address origin unimplemented: ...CUDAConvertFloatToHalf
testFloat32OfNew           [FAILED]    CUDAConvertHalfToFloat: input was cleared by a preceding graph rewrite
testStageQ8Tile            [FAILED]    address origin unimplemented: ...calc.MulNode
testBitsOfComputedParallel [FAILED]    address origin unimplemented: ...calc.MulNode
testPackHalfPairParallel   [FAILED]    address origin unimplemented: ...calc.MulNode
Test ran: 7, Failed: 6
```

All 7 pass with this change.

**Full suite** (RTX 5070 Ti sm_120, driver 580.142, CUDA 13.0 NVRTC, JDK 25): 200 suites. Six report failures, and every one of them reproduces on a pristine `develop` (`ec970e26d`) build:

| --- | --- | --- |
| `TestProfiler` | 3 failed | same 3 |
| `TestReductionsFloats` | 1 failed (`testComputePi`) | same 1, 3 runs out of 3 |
| `TestCuBlasLt` | 1 failed in-suite | order-dependent on both (0 or 9 on re-run) |

No suite fails here that passes on develop. Note `tornado-test` exits 0 and writes `OK` to `.unittestingStatus` even with these failures, so the table is built from the per-suite lines.

mikepapadim:
The separate defect noted at the end of the description — `new HalfFloat(short)` storing a numerically converted value instead of reinterpreting the bits — is now fixed in #1099. That PR adds the `__ushort_as_half` reinterpretation this one's float-only scoping exists to avoid needing; once both land, the scoping in `isSyntheticHalfReceiver` can be dropped so `new HalfFloat(short).getHalfFloatValue()` compiles too. Both touch `getHalfFloatValue` in `TornadoHalfFloatReplacement`, so whichever merges second needs a trivial rebase — happy to do it.

mikepapadim:

| case | input | expected |
| --- | --- | --- |

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/plugins/CUDAHalfFloatPlugins.java
```diff
@@ -69,10 +72,54 @@ private static boolean isSyntheticHalfReceiver(ValueNode receiverValue) {
             return true;
         }
         ValueNode unproxified = receiverValue instanceof ValueProxyNode proxy ? proxy.value() : receiverValue;
+        // A freshly constructed `new HalfFloat(x)` - either the allocation as the parser still has it on
+        // the stack, or the NewHalfFloatInstance the <init> node plugin substitutes for it.
+        // TornadoHalfFloatReplacement dissolves that allocation into the half value it wraps, so a field
+        // read inlined against it ends up addressing a half register rather than an object: the
+        // "address origin unimplemented" refusal that made `new HalfFloat(f).getHalfFloatValue()`
+        // uncompilable while the same expression stored straight into a HalfFloatArray worked.
+        if (unproxified instanceof NewHalfFloatInstance || isHalfFloatAllocation(unproxified)) {
```
orionpapadakis, line 81: [P2] Preserve signed short semantics for constructor readbacks

Accepting these receivers routes getHalfFloatValue() through HalfFloatPlaceholder, which TornadoHalfFloatReplacement lowers to CUDAConvertHalfBitsToIntNode. That node emits `(unsigned) __half_as_ushort(...)`, but the Java accessor returns a signed short. Widening its result to int must sign-extend.

GPU reproducer, with `in[i] = -4.0f`:

```java
for (@Parallel int i = 0; i < out.getSize(); i++) {
    out.set(i, new HalfFloat(in.get(i) * 2.0f).getHalfFloatValue());
}
```

Java expects `-14336` (signed `0xC800`); the patched CUDA compiler returns `51200`. The same kernel with `& 0xFFFF` passes, which is why the new integer-output tests miss this.

Please sign-extend the accessor result before integer consumers and add an unmasked regression test with a negative half value. Explicitly masked packing should retain its current result.

orionpapadakis (changes requested):
The constructor readback fix passes all 8 tests in TestHalfFloatValueConversion, but an additional GPU reproducer exposes a signed-short correctness issue. Please preserve the Java return semantics of getHalfFloatValue() before merging.

Validation: PR compiler changes and regression tests were applied to an isolated local CUDA SDK and run on an RTX 5090 with driver 580.142, CUDA 13.0, and JDK 21. CPU fallback was disabled. The masked control passes; the unmasked integer result fails as described inline.

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/compiler/plugins/CUDAHalfFloatPlugins.java
orionpapadakis: Prepared and GPU-validated a fix for this finding. The additional tests refined the diagnosis: there are two paths to correct, plus a shared-consumer case.

1. `isWriteHalfFloat()` currently treats any `JavaWriteNode` carrying a `HalfFloatPlaceholder` as a half store. An unmasked `IntArray.set()` therefore becomes a 16-bit `__half` store, leaving the upper 16 bits unchanged. Restrict that replacement to `JavaKind.Short` destinations.
2. Other consumers need signed bits: give `CUDAConvertHalfBitsToIntNode` a Java short stamp and a CUDA `INT` result, and emit `(int) __half_as_short(...)`.
3. Replacing a 16-bit store must not replace every use of its placeholder. A shared `short bits` stored into both `ShortArray` and `IntArray` otherwise makes the integer consumer receive the numeric half value (`-8`) instead of its signed bit pattern (`-14336`). Preserve the placeholder for the remaining consumers.

Validation on the same RTX 5090 / CUDA 13.0 / JDK 21 setup, with CPU fallback disabled:
- Four added tests fail on the original PR: unmasked computed bits, unmasked loaded bits, widening/shift/comparison semantics, and shared short/int consumers. The original eight tests pass.
- After the fix, all **66 tests pass** across the conversion suite and nine neighboring half/MMA suites, including masked half packing and Q8 staging.
- Inputs cover positive/negative values, signed zeros, subnormals, maximum finite halves and infinities. Integer outputs are prefilled with a sentinel and transferred to the device to detect partial-width writes.

The complete compiler correction is below; add those four regression cases to `TestHalfFloatValueConversion`. The review remains Request changes until the fix lands.

```diff
diff --git a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRStmt.java b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRStmt.java
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRStmt.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDALIRStmt.java
@@ -2645,10 +2645,10 @@
 
         @Override
         public void emitCode(CUDACompilationResultBuilder crb, CUDAAssembler asm) {
-            // u32_result = (unsigned) __half_as_ushort(half_value);
+            // Preserve the signed short returned by HalfFloat.getHalfFloatValue().
             asm.indent();
             asm.emitValue(crb, result);
-            asm.emit(" = (unsigned) __half_as_ushort(");
+            asm.emit(" = (int) __half_as_short(");
             asm.emitValue(crb, halfValue);
             asm.emit(")");
             asm.delimiter();
diff --git a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/CUDAConvertHalfBitsToIntNode.java b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/CUDAConvertHalfBitsToIntNode.java
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/CUDAConvertHalfBitsToIntNode.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/nodes/CUDAConvertHalfBitsToIntNode.java
@@ -36,12 +36,12 @@
 import uk.ac.manchester.tornado.drivers.cuda.graal.lir.CUDALIRStmt;
 
 /**
- * Reinterprets the raw bit pattern of an f16 value as an unsigned 32-bit
- * integer (zero-extended). Used to pack two f16 values into a single int32
- * for ldmatrix-based shared-memory tiles.
+ * Reinterprets the raw bit pattern of an f16 value as a signed Java short,
+ * sign-extended to the int stack kind used by Java bytecode. Callers packing
+ * unsigned half bits can explicitly mask the result with {@code 0xFFFF}.
  *
  * <p>CUDA C emitted (via CUDALIRStmt.HalfBitsToIntStmt):
- * <pre>u32_result = (unsigned) __half_as_ushort(half_value);</pre>
+ * <pre>s32_result = (int) __half_as_short(half_value);</pre>
  */
 @NodeInfo
 public class CUDAConvertHalfBitsToIntNode extends ValueNode implements LIRLowerable {
@@ -53,14 +53,14 @@
     private ValueNode halfValueNode;
 
     public CUDAConvertHalfBitsToIntNode(ValueNode halfValueNode) {
-        super(TYPE, StampFactory.forKind(JavaKind.Int));
+        super(TYPE, StampFactory.forKind(JavaKind.Short));
         this.halfValueNode = halfValueNode;
     }
 
     @Override
     public void generate(NodeLIRBuilderTool generator) {
         LIRGeneratorTool tool = generator.getLIRGeneratorTool();
-        Variable result = tool.newVariable(LIRKind.value(CUDAKind.UINT));
+        Variable result = tool.newVariable(LIRKind.value(CUDAKind.INT));
         Value halfValue = generator.operand(halfValueNode);
         tool.append(new CUDALIRStmt.HalfBitsToIntStmt(result, halfValue));
         generator.setResult(this, result);
diff --git a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/phases/TornadoHalfFloatReplacement.java b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/phases/TornadoHalfFloatReplacement.java
--- a/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/phases/TornadoHalfFloatReplacement.java
+++ b/tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/phases/TornadoHalfFloatReplacement.java
@@ -337,10 +337,10 @@
     }
 
     private static boolean isWriteHalfFloat(JavaWriteNode javaWrite) {
-        if (javaWrite.value() instanceof HalfFloatPlaceholder) {
-            return true;
-        }
-        return false;
+        // A placeholder also represents getHalfFloatValue() used as a Java short.
+        // Only a 16-bit destination can be replaced by a half store; an int write
+        // must retain its width and receive the sign-extended bits instead.
+        return javaWrite.getWriteKind() == JavaKind.Short && javaWrite.value() instanceof HalfFloatPlaceholder;
     }
 
     private static void replaceFixed(Node n, Node other) {
@@ -507,13 +507,16 @@
                     // if the result of an operation or a stored value is written
                     writingValue = placeholder.getInput();
                 }
-                placeholder.replaceAtUsages(writingValue);
-                placeholder.safeDelete();
+                // Replace only this 16-bit store. Other consumers still need the
+                // placeholder to become signed bits (or a half-to-float conversion).
                 AddressNode writingAddress = javaWrite.getAddress();
                 WriteHalfFloatNode writeHalfFloatNode = new WriteHalfFloatNode(writingAddress, writingValue);
                 graph.addWithoutUnique(writeHalfFloatNode);
                 replaceFixed(javaWrite, writeHalfFloatNode);
                 deleteFixed(javaWrite);
+                if (placeholder.hasNoUsages()) {
+                    placeholder.safeDelete();
+                }
             }
         }
 
```

orionpapadakis (approved):
Fixed and pushed in f262ecf3a. HalfFloat.getHalfFloatValue() now preserves Java signed-short semantics when widened to int. Half-float write replacement is restricted to short stores, and replacing one short store no longer rewrites other consumers of the same placeholder.

Four new regressions failed before the fix and pass afterward: computed and loaded signed bits, widening/comparison/shift uses, and shared ShortArray/IntArray consumers. The full targeted CUDA run passed 66 tests on an NVIDIA GeForce RTX 5090 Laptop GPU with JDK 21. This resolves the earlier change request; approving the updated head.

PR #1099, opened 2026-09-18
[cuda] Reinterpret the bits a HalfFloat(short) carries, instead of converting them

mikepapadim:
Found while testing #1098; reported there and split out as its own change.

### The defect

`new HalfFloat(short)` takes the half's **raw bit pattern** — its javadoc says so, and `getHalfFloatValue()` hands the same bits back — but the CUDA backend let that short through as an integer. The store then assigned it to a `__half` destination and C++ converted it numerically:

```java
out.set(i, new HalfFloat((short) bits));   // bits of -8.0, i.e. 0xC800
```
```c
sh_11  =  (short) i_7;
i_12  =  (int) sh_11;
*(( __half *) ul_10)  =  i_12;             // stores -14336.0
```

Every element of a 32-element round trip came out as a different number, with nothing to indicate it. The constant form, `new HalfFloat((short) 0x3C00)`, does not compile at all.

This is independent of #1098 and reproduces on `develop` today.

### The fix

A `CUDAConvertIntBitsToHalfNode` — the inverse of the existing `CUDAConvertHalfBitsToIntNode` — emits the reinterpretation the constructor means:

```c
half_value = __ushort_as_half((unsigned short) bits);
```

The cast to `unsigned short` is what drops the sign extension a `short` picks up on the Java operand stack, so a half with the sign bit set keeps its bit pattern instead of wrapping to another value.

The replacement phase routes an integer-valued constructor argument through it. Values that already carry the synthetic `HalfFloatStamp` are left alone: those are halves in their own right (a read from a `HalfFloatArray`, the result of half arithmetic) and must not be reinterpreted again.

### Relationship to #1098

#1098 scopes its plugin interception to the float constructor precisely because this reinterpretation did not exist — intercepting the short form there would have turned a compile error into silently wrong numbers. Once both land, that scoping can be dropped so `new HalfFloat(short).getHalfFloatValue()` compiles too; it still fails with the address-origin refusal today. Both touch `getHalfFloatValue` in `TornadoHalfFloatReplacement` (each adds a branch to the same chain), so expect a trivial textual conflict whichever merges second — happy to rebase.

### Tests

`TestHalfFloatShortBits` (new, registered in `tornado-test`) builds halves from bit patterns of both signs, zero, a subnormal and the largest finite half, and checks them against `Float.float16ToFloat`:

- `testStoreBits` — the `@Parallel` loop form
- `testStoreBitsKernelContext` — the same with explicit `globalIdx` indexing
- `testStageBitsThroughLocalTile` — bits staged through a `ctx.allocateIntLocalArray` tile, the shape a kernel unpacking quantized weights uses
- `testRoundTripBits` — a half taken apart into its bits and put back together, an exact copy
- `testConstantBits` — `0x3C00` is half 1.0, not the number 15360

All five fail on `develop`:

```text
testStageBitsThroughLocalTile [FAILED] out[1] expected:<1.0> but was:<15360.0>
testStoreBits                 [FAILED] out[1] expected:<1.0> but was:<15360.0>
testRoundTripBits             [FAILED] out[1] expected:<1.0> but was:<15360.0>
testStoreBitsKernelContext    [FAILED] out[1] expected:<1.0> but was:<15360.0>
testConstantBits              [FAILED] (bailout: does not compile)
Test ran: 5, Failed: 5
```

and pass with this change.

### Validation

| Suite | Result |
| --- | --- |
| `TestHalfFloatShortBits` | 5 passed (new) |
| `TestHalfFloatInlineWrite` / `TestHalfFloatInPlaceUpdate` | 3 / 4 passed |
| `foundation.TestHalfFloats` | 4 passed |
| `vectortypes.TestHalfFloats` | 26 passed, 18 unsupported (unchanged) |
| `TestHalf2Packed` / `TestHalfFloatDotProduct` | 7 / 1 passed |
| `TestHistogramHalfFloat` / `TestLocalMemoryReductionsHalfFloats` | 1 / 4 passed |
| `TestBitonicSortHalfFloat` | 1 passed |
| `TestMatrixMultiplicationMMA` | 14 passed |
| `tile.TestTileRawBuffers` | 2 passed, 2 unsupported (unchanged) |

### Scope

CUDA only. The OpenCL and Metal backends carry parallel copies of the replacement phase and may well have the same defect, but I have no device here to test a change to them.


orionpapadakis (approved):
No blocking findings in the short-bits constructor/store lowering. The conversion reinterprets integer bits as half rather than numerically converting them, while existing HalfFloatStamp values retain their half representation.

Validation: 38 targeted CUDA regression tests passed, plus an independent exhaustive constructor/store check covering all 65,536 short bit patterns, including NaNs, on an NVIDIA GeForce RTX 5090 Laptop GPU with JDK 21. Approval covers this PR's constructor/store scope; the separate accessor readback fixes are in #1098.

PR #1100, opened 2026-09-18
[docs] Put the CUDA Tile path in the README

mikepapadim:
The `TileContext` path landed in #1083 with a full guide in `docs/source/tile-api.rst`, but nothing in the README says it exists. This adds it, and keeps it short.

### What was added

**A section in the NVIDIA part**, `🧩 CUDA Tile: a tile is the unit of work`, leading with the fact that distinguishes the path: a task whose kernel takes a `TileContext` is compiled through CUDA Tile (`nvcc -tilecubin --tile-only`) rather than the SIMT path, you write what **one tile block** does, and **nothing in the API names a thread, a warp or a fragment**.

Then the tile GEMM from `tornado-examples/.../tile/TileMatrixMultiply.java` — nine lines, `tc.mma` in a loop, `WorkerGrid2D(n / TILE, n / TILE)` annotated as tile blocks rather than threads — and four bullets a reader needs before trying it:

- it composes: a `@Parallel` kernel, a `KernelContext` kernel, a `TileContext` kernel and a native cuBLAS call in one `TaskGraph`, captured into a single CUDA Graph (links `TestTileChaining`);
- a tile kernel is also ordinary Java, because every operation has a JVM implementation;
- tile shapes are compile-time constants, extents are not;
- the toolkit / driver / compute-capability requirements, and that an unsupported combination fails with `TornadoDeviceTileNotSupported` naming what is missing, before code generation.

It closes with links to the guide and to the runnable examples.

**A row in the NVIDIA capability table**, next to the CUDA Graphs and Tensor Core rows, pointing at that section.

**A line in the NVIDIA FAQ entry**, which enumerates what the CUDA backend does and stopped at CUDA Graphs.

### Precision

- Every claim is checked against the code, not the PR description: `-tilecubin --tile-only` in `CUDATileCompiler`, the sm_80 / toolkit-13.3 gates and `TornadoDeviceTileNotSupported` in `CUDATileSupportPhase`, the one-graph chaining in `TestTileChaining#testWholeChainInOneCudaGraph`, and the worker-grid shape in `TileMatrixMultiply`.
- The code snippet compiles against `tornado-api` (checked with `javac` against the built jar), so it is not pseudocode.
- The feature is marked **"On `develop`, not in 6.1.0"**, because it is unreleased.
- The guide is linked in-repo (`docs/source/tile-api.rst`), not on readthedocs: `https://tornadovm.readthedocs.io/en/latest/tile-api.html` is a 404 today, since "latest" does not track `develop`. Worth switching to the readthedocs link once it publishes.

### Unrelated one-liner

The contributing section linked `tornadovm-skill.skill`, which does not exist in the repo; repointed at `.claude/skills`, which does. Every relative link in the README now resolves.


stratika:
I close this PR, since we have a new opened (#1111).

PR #1101, opened 2026-09-18
[cuda] In-kernel reads of int32 MMA accumulator elements and byte-offset int8 fragment loads

orionpapadakis:
#### Description

Two small additions to the CUDA backend's MMA support, needed for block-quantized integer projections (the llama.cpp "MMQ" pattern: int8 activations × int8-unpacked Q4_0 weights on `m16n8k32 s8.s8.s32`, each 32-wide K block's int32 dot products scaled by that block's own FP32 scales before they are accumulated):

1. **Constant-index reads of int32 MMA accumulator elements.** A `LoadIndexedNode` whose array is an MMA accumulator fragment (the result of `mmaFragmentInt`, of an int8 `mmaInt8` compute, or a phi of those — a loop-carried fragment, which the backend already emits as a C local array) is lowered to a new `CUDAMMAFragmentElementReadNode` and emitted as `result = frag[i]`, a register read. Previously such a read reached `CUDAAddressLowering`, which refused the fragment as an address base (`unimplemented: address origin unimplemented: CUDAMMAComputeNode`), so the only way to scale per block was `mmaStoreInt` to global memory and a separate pass. Float fragments, non-constant indices and indices outside 0..3 are refused with a clear message; ordinary array lowering is untouched. Reads only.

2. **Byte-offset overloads `mmaLoadAInt8(int[], int, int)` / `mmaLoadBInt8(int[], int, int)`.** The int8 fragment-load nodes already carried an optional byte offset; the `KernelContext` overloads and their invocation plugins were missing, so a kernel could only load the sub-tile at a shared tile's base. These are the int8 counterparts of the existing `mmaLoadA(int[], int, int)` / `mmaLoadB(int[], int, int)`.

Motivation and measurements: in jllm beehive-lab/jllm#163 an int8 block-scaled projection built on these two additions runs the complete quantize + decode + GEMM path 20–25% faster than the FP16 dequantize + GEMM pair at the production shapes, +22–26% end-to-end prefill throughput; the int8→int32 tensor path runs 3.3× the FP16→FP32 rate on that device.

#### Problem description

Not a bug fix. Without (1), `acc[i]` on a fragment throws in `CUDAAddressLowering.lower` at `AddressLoweringByNodePhase`; without (2), int8 A/B fragments cannot be loaded from an offset inside a shared array.

#### Backend/s tested

- [ ] OpenCL
- [x] CUDA
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

```bash
make BACKEND=cuda
source setvars.sh
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMMAInt8FragmentElementRead
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMAInt8
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMA
tornado-test -V uk.ac.manchester.tornado.unittests.kernelcontext.matrices.TestMatrixMultiplicationMMACpAsync
```

The new suite: element reads compared with `mmaStoreInt` on every element of a 16×8 tile (the documented lane/element → row/column mapping); two K blocks scaled by distinct FP32 factors before FP32 accumulation, with the fragment loop-carried (the phi path), exact against a host reference; and the second of two A and B tiles selected through the byte-offset overloads. Results on the setup above: 3/3, 9/9, 14/14, 6/6 pass. The new tests skip on OpenCL and Metal like the existing int8 MMA tests.


orionpapadakis (commented):
Fixed and pushed in 68c8d10d0. MMA fragment recognition did not unwrap PiNode and ValueProxyNode, so reading a loop-carried int accumulator after a runtime-controlled loop reached unsupported address lowering. Both recognition helpers now unwrap those nodes.

Added a regression that checks all four per-lane elements against a host dot-product reference after 0, 1, and 3 iterations. Also corrected the existing block-scaled test's description: it creates a fresh fragment per block and did not cover the loop-carried path.

PR #1102, opened 2026-09-18
[opencl] Let a converted half be stored and read back as a value

mikepapadim:
The OpenCL twin of #1098 — found by taking the open CUDA compiler-fix PRs and checking each one against the OpenCL backend on real hardware.

### What OpenCL does on `develop`

| Kernel expression | OpenCL result on develop |
| --- | --- |
| `out.set(i, new HalfFloat(f * 2.0f))` | `node is not LIRLowerable: NewHalfFloatInstance` |
| `new HalfFloat(f * 2.0f).getHalfFloatValue()` | `address origin unimplemented: ...calc.MulNode` |
| `new HalfFloat(in.get(i)).getHalfFloatValue()` | `address origin unimplemented: OCLConvertFloatToHalf` |
| `new HalfFloat(f * 2.0f).getFloat32()` | `node is not LIRLowerable: NewHalfFloatInstance` |

The middle two are the defect #1096 and #1057 report against CUDA, with `OCLConvertFloatToHalf` where the CUDA error names `CUDAConvertFloatToHalf`. The other two are a second gap: the inline-write fixes the CUDA phase already carries were never ported, so on OpenCL even *storing* a computed `new HalfFloat(...)` fails — which is why `TestHalfFloatInlineWrite` skips this backend.

### The fix — four pieces, each the twin of one the CUDA phase has

1. **`OCLHalfFloatPlugins`** intercepts `getHalfFloatValue()` / `getFloat32()` on a freshly constructed `new HalfFloat(float)` receiver, so the call is not inlined into a read of the `halfFloatValue` field over a value the replacement phase is about to dissolve into a half register. Scoped to the float constructor, for the same reason as #1098: `new HalfFloat(short)` carries raw bits this backend has no reinterpretation for.
2. **The phase converts any float-valued constructor argument**, not only a read.
3. **The phase gains the three inline-write fixes**: find the `NewHalfFloatInstance` by walking the fixed chain instead of requiring adjacency, match the write context through the placeholder's data input as well as its predecessor, and drop leftover dead `HalfFloat` allocations.
4. **A new `OCLConvertHalfBitsToIntNode`** gives `getHalfFloatValue()` something to lower to; without it the `HalfFloatPlaceholder` reached LIR.

On (4), one thing worth a reviewer's eye. The obvious spelling is `as_ushort(h)`, and NVIDIA's OpenCL rejects it:

```text
<kernel>:25:18: error: call to 'as_ushort' is ambiguous
  ui_11 = (uint) as_ushort(half_10);
cl_kernel.h:15077:25: note: candidate function
ushort __OVERLOADABLE__ as_ushort (short);
```

so the statement reinterprets through a pointer instead, which every implementation accepts:

```c
{ half __hbits_ui_11 = (half) f_9; ui_11 = (uint) *((__private ushort *) &__hbits_ui_11); }
```

The temporary also means the operand does not have to be a variable, and the enclosing block keeps the name from colliding.

### Tests

`TestHalfFloatValueConversionOpenCL` carries the same kernels as #1098's CUDA class: the `KernelContext` reproducers from #1096 (`store`, `value`, `valueOfLoad`, `getFloat32`) and `packQ8Halves` from #1057 staged through a shared tile, plus the `@Parallel` variants. All seven fail on `develop` with the errors in the table and pass with this change.

`TestHalfFloatInlineWrite` no longer skips OpenCL: its three cases pass there now (verified), where they were skipped as a CUDA-only fix.

### Validation

OpenCL on an RTX 5070 Ti (NVIDIA OpenCL 1.2 platform), JDK 25, OpenCL-only SDK. Every half-float suite, before and after, is identical apart from the new tests:

| Suite | develop | this branch |
| --- | --- | --- |
| `TestHalfFloatValueConversionOpenCL` | 7 failed | **7 passed** |
| `TestHalfFloatInlineWrite` | 3 skipped | **3 passed** |
| `TestHalfFloatInPlaceUpdate` | 4 ran, 4 unsupported | same |
| `foundation.TestHalfFloats` | 4 passed | same |
| `vectortypes.TestHalfFloats` | 26 ran, 23 unsupported | same |
| `TestHalf2Packed` | 7 ran, 1 unsupported | same |
| `TestHalfFloatDotProduct` / `TestHalfFloatLocalArray` | 1 / 1 passed | same |
| `TestHistogramHalfFloat` / `TestLocalMemoryReductionsHalfFloats` | 1 / 4 ran | same |
| `TestBitonicSortHalfFloat` / `tile.TestTileRawBuffers` | 1 / 2 ran | same |

### Relationship to the CUDA PRs

Independent of #1098 — different files, different backend — and either can merge first. The two test classes are near-duplicates by design (each scoped to its own backend so both stay green standalone); once both land they can be folded into one class that runs on both.

For the record, the other two open CUDA compiler fixes do **not** apply here: #1093 gates `cuda_fp16.h` by scanning the generated source, and OpenCL emits `#pragma OPENCL EXTENSION cl_khr_fp16` unconditionally; #1094 is MMA fragment access, and this backend rejects every `mma*` intrinsic by design.


mikepapadim:
Added a second commit: **the `HalfFloat(short)` reinterpretation** (the OpenCL twin of #1099), because leaving it out would have made this PR strictly worse in one corner.

The dead-allocation cleanup in the first commit lets `new HalfFloat((short) bits)` compile on OpenCL for the first time — on develop it fails with `Node implementing Lowerable not handled: NewInstance`. Compiling, it would then assign the short to a `half` destination and the implicit conversion would renumber it: the bits of -8.0 (`0xC800`) written as **-14336.0**, silently, for every element. Measured on this branch before the second commit: 31 of 32 elements wrong. Trading a compile error for a silent miscompile is worse than either, so the reinterpretation belongs here rather than in a follow-up.

`OCLConvertIntBitsToHalfNode` is the inverse of the node the first commit added, through a pointer for the same overload-resolution reason, with a `ushort` cast to drop the sign extension a `short` picks up on the Java operand stack.

Verified on the same OpenCL device — every half-float suite, plus both new classes:

| Suite | Result |
| --- | --- |
| `TestHalfFloatValueConversionOpenCL` | 7 passed |
| `TestHalfFloatShortBitsOpenCL` | 5 passed (new) |
| `TestHalfFloatInlineWrite` | 3 passed (was 3 skipped on develop) |
| `TestHalfFloatInPlaceUpdate` / `foundation.TestHalfFloats` | 4 / 4 — unchanged |
| `vectortypes.TestHalfFloats` / `TestHalf2Packed` | 26 / 7 — unchanged |
| `TestHalfFloatDotProduct` / `TestHalfFloatLocalArray` | 1 / 1 — unchanged |
| `TestHistogramHalfFloat` / `TestLocalMemoryReductionsHalfFloats` | 1 / 4 — unchanged |
| `TestBitonicSortHalfFloat` / `tile.TestTileRawBuffers` | 1 / 2 — unchanged |

The standalone short-ctor reproducer that mismatched on all 32 elements now passes.

mikepapadim:

| case | input | expected |
| --- | --- | --- |

Checked through **both** OpenCL implementations on this machine — NVIDIA (RTX 5070 Ti) and Intel (UHD 770) — and the `(half)` conversion matches on every case on both.

orionpapadakis (approved):
Fixed and pushed in 291b0dfe0. The OpenCL accessor must return signed short bits widened to Java int, not zero-extended unsigned bits. The fix also restricts half stores to short-width writes and preserves other consumers when replacing a short store, preventing partial IntArray writes and shared-consumer miscompilation.

All four added regressions failed before the fix. Afterward, 25 supported OpenCL half-float tests passed on an NVIDIA GeForce RTX 5090 Laptop GPU with JDK 21. An initial broader run also included TestHalfFloatDotProduct, which explicitly rejects OpenCL; that unsupported test was excluded from the clean run. Approving the updated head.

PR #1103, opened 2026-09-18
[metal] Let a converted half be read back as a value, not only stored (untested: needs a macOS run)

mikepapadim:
> ### ⚠️ This change has never been executed
>
> It was written on Linux, where the Metal backend compiles but cannot run. It is verified by reading the code and by its CUDA (#1098) and OpenCL (#1102) twins — both of which *are* verified on hardware — and nothing more. **Draft until someone runs `TestHalfFloatValueConversionMetal` on macOS.** That is the one command that settles it:
>
> ```bash
> tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestHalfFloatValueConversionMetal
> ```

The Metal third of the campaign that produced #1098 (CUDA) and #1102 (OpenCL): taking each open compiler-fix PR and checking whether the defect exists on the other backends.

### Why the defect is present here

- `MetalAddressLowering` has the same whitelist as the CUDA and OpenCL ones, so a read whose base is a half register is refused the same way: `address origin unimplemented: ...`.
- `MetalHalfFloatPlugins.isSyntheticHalfReceiver` has the same receiver test, so `getHalfFloatValue()` / `getFloat32()` on a freshly constructed `new HalfFloat(x)` is inlined into a read of the `halfFloatValue` field — over a value `TornadoHalfFloatReplacement` is about to dissolve into a half.

Two things are specific to Metal, and make it the weakest of the three backends here:

1. **There was no float-to-half conversion node at all.** `getHalfFloatValue` in the phase handled only `ConstantNode`, so a float constructor argument was carried on as a float register and only became a half implicitly, at the store.
2. **The leftover `HalfFloatPlaceholder` was replaced by the value it wrapped.** So `getHalfFloatValue()` handed a `half` straight to integer arithmetic, which MSL converts numerically instead of reinterpreting — the bits of half `1.0` (`0x3C00`) would read as `1`, not `15360` *or* the right answer.

### The change

Mirrors #1102 piece for piece, plus the two nodes Metal was missing:

```c
half_value = (half)(float_value);
u32_result = (uint) as_type<ushort>(half_value);
```

`MetalPreamble` already emits `as_type`, so the reinterpretation spelling is one this backend uses elsewhere. (OpenCL needed a different spelling for the same operation — NVIDIA's OpenCL rejects `as_ushort(half)` as ambiguous — which is why #1102 goes through a pointer instead.)

The plugin interception is scoped to the **float** constructor, as in the other two: `new HalfFloat(short)` carries raw bits this backend has no reinterpretation for on the store path, so intercepting it would turn a compile error into a silently wrong number.

### What a macOS run should check

- `TestHalfFloatValueConversionMetal` (added here) — the `KernelContext` reproducers from #1096 and the Q8_0 weight staging from #1057, with `@Parallel` variants. On the CUDA and OpenCL twins every one of these fails before the fix and passes after.
- The store path more broadly: 41 unit-test files build a `new HalfFloat(...)`, and the float-conversion piece changes how that value reaches the store (explicit `(half)` cast instead of an implicit conversion on assignment). Semantically the same, but it is the part of this change with existing coverage.
- No existing test exercises `getHalfFloatValue()` on anything but a CUDA-only path (`TestMatrixMultiplicationMMA`, `TestAsyncCopyToLocal`), so that half of the change has no prior Metal coverage to regress — and none to lean on either.

### Relationship to the other two

Independent of #1098 and #1102 — different backend, different files. Merge order does not matter.


mikepapadim:
One more thing for whoever runs this on macOS, found while finishing the OpenCL twin (#1102).

`new HalfFloat(short)` takes the half's **raw bit pattern**, and on CUDA and OpenCL the backend used to assign that short to a `half` destination, where the implicit conversion renumbers it — the bits of -8.0 (`0xC800`) stored as -14336.0. That is fixed for CUDA in #1099 and for OpenCL in #1102's second commit.

Metal looks like it has the same defect, and **this PR does not address it**: the phase change here wraps only *float*-valued constructor arguments, so the short path is untouched, and the plugin interception is scoped to the float constructor precisely so it cannot turn that miscompile into a reachable one. So this PR should not make the short case worse — but it probably does not make it right either.

I could not check: it needs a device. A macOS run of

```bash
tornado-test -V uk.ac.manchester.tornado.unittests.arrays.TestHalfFloatShortBitsOpenCL
```

(with its `assumeOpenCLBackend` switched to Metal) would settle it in one go. If it fails the way CUDA and OpenCL did, the fix is the Metal analogue of #1099: a node emitting `as_type<half>(ushort_bits)`, wired into `getHalfFloatValue` for integer-kind constructor arguments.

mikepapadim:

| case | input | expected |
| --- | --- | --- |

**Not measured here** — Metal needs macOS, so on this PR the test is the check a reviewer with a Mac should run, not a result. MSL specifies its float-to-half conversion as IEEE binary16, and both other backends pass every case on hardware, so the expectation is that it passes unchanged.

orionpapadakis (approved):

PR #1104, opened 2026-09-18
[backends] Emit Java's wrapping integer arithmetic with defined wrap-around

mikepapadim:
Java specifies `+`, `-`, `*` and `<<` on `int` and `long` to **wrap** on overflow. The backends emitted them as plain signed C arithmetic, whose overflow is **undefined**, and the device compilers act on that.

```
(a * a) >> 20    a = 46341    device  +2048     Java  -2048      the product overflows
a << 31          a = 1        device  0         Java  INT_MIN
```

The second is the instructive one. Graal canonicalises `a << 31` into `-(a << 31)` — the same value under wrap-around — and negating `INT_MIN` is undefined too, so the emitted `-i_9` let the compiler fold the whole expression to zero:

```c
i_9   =  i_7 << 31;
i_10  =  -i_9;          // device: 0
```

### How this surfaced

`tornado-unittests/.../fuzz/TestCudaCodegenRegressions` documents five findings from tornado-fuzz, each `@Ignore`d as *"Open CUDA backend bug"*. Two things turned out to be wrong with that framing:

1. **They are not CUDA bugs.** With the `@Ignore`s removed, four of the five fail *identically* on OpenCL — same values, same expressions. They are one shared defect: Java's wrapping arithmetic emitted as signed C arithmetic.
2. **The class is not in `tornado-test` at all**, so none of it has ever run in the suite. It is registered here.

They also are not recent: the same results come out of the 6.1.0-era dist.

### The fix

`ADD`, `SUB`, `MUL` and left shift on the signed integer kinds are emitted through the unsigned kind, which C defines as wrap-around:

```c
i_7  =  (int) ((unsigned int) i_6 + 15);
i_6  =  (int) ((unsigned int) i_5 << 4);
```

Casting the **left** operand is enough — the usual arithmetic conversions make the whole expression unsigned — and that matters for cost, see below. The narrowing back to the signed kind is *implementation-defined* rather than undefined, and every target TornadoVM runs on defines it as the two's-complement reinterpretation Java specifies. Unary minus is emitted the same way. Unsigned kinds and floats are untouched: those already wrap, or define overflow.

### Cost: none measured

Register-tiled SGEMM (64 threads/group, 32×32 tile, 4×4 per thread) and a tiled matmul, RTX 5070 Ti, interleaved before/after over four and three runs:

| Benchmark | before | after |
| --- | --- | --- |
| register-tiled SGEMM | 0.374 / 0.349 / 0.355 / 0.349 ms | 0.347 / 0.348 / 0.343 / 0.370 ms |
| tiled matmul 1024² | 0.704 / 0.708 / 0.701 ms | 0.705 / 0.700 / 0.704 ms |

Both inside run-to-run spread. Worth knowing for anyone revisiting this: casting **both** operands — the obvious first form — did cost about 2% on the SGEMM and doubled the noise in the emitted code, which is why only the left one is cast.

### Tests

The four integer-overflow tests are un-`@Ignore`d and pass on **both** backends:

```text
testShiftBy31          [PASS]      testIntMinMul           [PASS]
testSignBitNegate      [PASS]      testSignedShiftProduct  [PASS]
```

The fifth stays ignored: `(INT_MIN / (b|1)) / (i|1)` crashes a *shared Graal phase* with `unhandled node in reassociation with constants`, which is a separate defect and not undefined behaviour.

The class javadoc is corrected to say the findings were cross-backend rather than CUDA-specific.

### Validation

RTX 5070 Ti, JDK 25. Integer-heavy suites, both backends, all green and unchanged from develop:

| Suite | CUDA | OpenCL |
| --- | --- | --- |
| `foundation.TestIntegers` / `TestLong` | 7 / 2 | 7 / 2 |
| `numpromotion.Types` / `Inlining` | 9 / 2 | 9 / 2 |
| `logic.TestLogic` | 9 | 9 |
| `arrays.TestArrays` / `TestBasicOperations` | 23 / 24 | 23 / 24 |
| `math.TestMath` | 36 | 36 |
| `loops.TestLoops` | 26 | 26 |
| `branching.TestConditionals` | 20 | 20 |
| `reductions.TestReductionsIntegers` | 15 | 15 |
| `codegen.TestSharedMemoryStaging` | 7 | 7 |
| `bitsets.BitSetTests` | 1 | — |

A full `tornado-test` run on CUDA is in progress; I will post the per-suite comparison against develop as a comment.

### Metal

Same change, **compile-checked only** — there is no Metal device on this machine. The emission path is identical in shape to the other two, and a macOS run of `TestCudaCodegenRegressions` would confirm it.


mikepapadim:
Full `tornado-test` run on this branch: **201 suites**, RTX 5070 Ti / JDK 25.

Suites reporting failures, and what each one is:

| Suite | Status |
| --- | --- |
| `TestMatrixMultiplicationKernelContext`, `TestProfiler`, `TestDevices`, `ComputeTests`, `TestCuBlasLt` | pre-existing — all five reproduce on a pristine `develop` build, checked earlier in this campaign |
| `QuantizationTests` | **7/7 passes standalone on both builds** (with and without this change) — in-suite flake, not a regression. Checked deliberately, because integer packing is exactly what this PR touches |
| `TestMemoryLeak` (6 entries) | passes standalone on both builds; these swing between runs on this machine (0 in one earlier full run, 2 in another, 6 here) |

Nothing fails here that passes on develop.

Also pushed: the fifth finding in `TestCudaCodegenRegressions` stays `@Ignore`d, but its note was wrong in the same way the other four were. It is not a CUDA backend crash — `(INT_MIN / (b|1)) / (i|1)` throws from Graal's `ReassociationPhase` on CUDA **and** OpenCL, and `-Dgraal.ReassociateExpressions=false` compiles the kernel and produces the correct result, which pins the phase. That phase comes from the vendored Graal jar, so it cannot be fixed from this repository; the note now records that and the workaround.

orionpapadakis (approved):
The unsigned arithmetic intermediates address the signed-overflow code-generation issue. Boundary testing also reproduced an adjacent, pre-existing CUDA bug: negative and oversized left-shift distances did not follow Java's masking rules. Commit 571ac8783 masks distances with 31 for int and 63 for long in the CUDA, OpenCL, and Metal emitters, and adds int/long boundary tests for addition, subtraction, multiplication, negation, and left shift.

Validation: 85 tests passed on CUDA and 85 on OpenCL, including the added boundary coverage, on an NVIDIA GeForce RTX 5090 Laptop GPU with JDK 21. The existing ignored division-reassociation test remains ignored. The Metal Java backend compiles; Metal runtime behavior was not tested because no Metal device is available. Approving the updated head.

tornado-drivers/cuda/src/main/java/uk/ac/manchester/tornado/drivers/cuda/graal/lir/CUDABinary.java
```diff
@@ -101,7 +101,14 @@ private void emitWrapping(CUDACompilationResultBuilder crb, CUDAAssembler asm) {
             asm.space();
             asm.emit(opcode.toString());
             asm.space();
-            asm.emitValueOrOp(crb, y);
+            if (opcode == CUDAAssembler.CUDABinaryOp.BITWISE_LEFT_SHIFT) {
+                // Java masks shift distances; an out-of-range C shift is undefined.
+                asm.emit("(");
+                asm.emitValueOrOp(crb, y);
+                asm.emit(kind == CUDAKind.INT ? " & 31)" : " & 63)");
```
mikepapadim: harcoded values!

orionpapadakis: good catch! replaced with `Integer.SIZE - 1` / `Long.SIZE - 1`

PR #1106, opened 2026-09-18
[ci] Stop pull requests queueing behind each other on the self-hosted runners

mikepapadim:
## Critical — CI health. Please prioritise this review.

@orionpapadakis @stratika — tagging you both as reviewers because CI is currently not usable on any open pull request, including yours. Nothing is failing; everything is queued.

### What is happening

Measured on `beehive-lab/TornadoVM` at 2026-09-18 15:59Z:

- **32 unfinished workflow runs, 212 unfinished jobs**, against **3 self-hosted runners** that take one job at a time.
- Queue depth by runner label: **127** waiting on `self-hosted,macOS,arm64` (1 runner), **46** on `self-hosted,Linux,x64,sdk-builder` (1 runner, `cyclone`), **36** on `self-hosted,Linux,x64` (2 runners).
- The macOS runner was at that moment executing jobs from a run **created at 10:53 that morning** — GitHub's self-hosted queue is FIFO, so every new pull request lands behind roughly five hours of work.
- Worst case observed: `linux / test-jdk21-on-jdk21-opencl` on #1094 sat queued from 2026-09-16 17:09Z until it was cancelled by hand at 2026-09-18 12:31Z — **43 hours, never assigned a runner**.

The runners are online and healthy. This is entirely a queue problem, and the causes are in these YAML files.

The reason it is worth treating as critical rather than as slowness: a stalled check never resolves, so a pull request is never red and never mergeable, and no one gets a signal telling them to do anything. Recent pull requests (#1099–#1104) all show this — some builds green, the rest yellow forever.

### What this changes

**1. Concurrency groups on both pull-request workflows** (`build-test.yml`, `unit-tests-reflection.yml`)

There were none. Every push to a branch started a fresh 21-job run and left its predecessor in the queue. On the day this was measured, `fix/opencl-halffloat-value-conversion` had **five queued runs on five different SHAs** (11:32, 11:44, 12:49, 13:09, 14:18) — four of them dead weight blocking everyone else. `fix/backends-wrapping-integer-arithmetic` had five, `develop` four.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Pull requests cancel their predecessors; pushes to `develop` do not, since a merge is a result someone will want to look up later.

This is at the workflow level on the *caller*, which is deliberately not the thing the existing comment in `build-test-platform.yml` warns against — that comment is correctly about putting a concurrency group on the per-matrix test job, where GitHub would keep one pending job and silently drop the rest of the matrix. Cancelling the caller cancels the reusable child runs cleanly.

**2. macOS device tests move to the nightly and to `ci-full-jdk-sweep`**

This is the binding constraint. One runner serves 4 builds + 6 tests per run, strictly serialised — measured durations put that at roughly **50 minutes of that one machine per run**, against an arrival rate several times higher. It cannot drain, and no amount of tuning elsewhere changes that.

A pull request now gets the **four macOS builds** — which is what catches what a macOS-blind change actually breaks: the Metal and OpenCL native compile, the aarch64 toolchain, the SDK assembly. Linux keeps its **full device test matrix on every pull request**, on two runners, so neither OpenCL nor CUDA loses device coverage. Put `ci-full-jdk-sweep` on a pull request to get the macOS device tests before merging; the nightly runs them unconditionally.

This is the judgement call in the PR and the one I would most like your opinion on, @orionpapadakis / @stratika — particularly whether Metal needs per-PR device coverage badly enough to justify a second macOS runner instead.

**3. The downstream sample apps run once per (platform, backend), not per JDK**

`TornadoVM-Ray-Tracer` and `KFusion` were cloned and `mvn clean install`-ed from scratch on *every* test leg, inside the slot that holds the GPU while it does it. On a subset run that is three legs per backend rebuilding the same two applications against the same SDK.

What they guard is the API surface the shipped artifact exposes — kernel entries resolved through the compiler-generated `SerializedLambda` — which is a property of the artifact, not of the JDK underneath it. So they now run on the newest-JDK leg of each (platform, backend) on a subset run, and on every leg on a full sweep. The `mvn install-file` step that only exists to seed their dependencies is gated with them.

### Effect per pull request

| | before | after |
|---|---|---|
| jobs on the single macOS runner | 10 (~50 min serialised) | 4 (~12 min) |
| Linux test legs | 6 | 6 (unchanged) |
| Linux legs building the sample apps | 6 | 2 |
| superseded runs left in the queue | all of them | none |

Nightly and `ci-full-jdk-sweep` coverage is unchanged: 4 builds + 14 tests per platform, sample apps on every leg.

### Verification

- All three workflows parse (`yaml.safe_load`).
- The `plan` job's embedded matrix generator was extracted and run against all three event shapes; output is as intended:
  - `pull_request` → `test_macos_enabled=false`, 4 macOS build jobs, 6 Linux test legs of which 2 run the apps
  - `pull_request` + `ci-full-jdk-sweep` → `test_macos_enabled=true`, 4 + 14 per platform, apps on all 14
  - `schedule` → same as the label case
- `develop` has **no required status checks** configured, so making the macOS test jobs conditional cannot block a merge.

### Not included, worth a follow-up

- **`sdk-builder` collides with the bare Linux label.** `cyclone` carries both `[self-hosted, Linux, x64, sdk-builder]` and `[self-hosted, Linux, x64]`, so on top of every CUDA leg being pinned to it, it also absorbs OpenCL legs, `code-quality` and `plan`. The existing comment in `build-test.yml` already flags the role-label as fragile; the collision with the unqualified label is the part it does not mention. A machine-unique label would fix it.
- **The 32 runs currently queued still need clearing by hand** — this change stops new pile-ups but does not drain the existing one, which will otherwise drain FIFO for the rest of the day.
- **`unit-tests-reflection.yml` has no `branches:` filter** on `pull_request` while `build-test.yml` filters to `develop`. Harmless (GitHub-hosted), but inconsistent.
- **`27-ea` is a flake source**: `actions/setup-java` failed in 8s on #1100 while sibling runs at the same minute passed. No retry.

https://claude.ai/code/session_014xVfr1SjQPTruQos3HArW9


orionpapadakis (approved):
Fixed the review findings in 487cfd8bd:

- Unrelated label events previously shared the active PR concurrency group, so they could cancel real checks and then skip all replacement jobs. They now use an isolated group, and only adding ci-full-jdk-sweep passes the label guard.
- Non-PR runs now use distinct groups. cancel-in-progress: false alone does not preserve pending runs in a shared group, so merge, nightly, and manual runs could previously displace one another.
- Sample apps now run once per backend and SDK in subset runs. The preview JDK 21 SDK and jdk22plus SDK are different artifacts; testing only the latter left the former without downstream integration coverage.

Validation: actionlint passed for all three workflows; the embedded matrix generator passed PR, labeled PR, push, schedule, and both manual-dispatch modes. Concurrency/label scenarios also passed. Subset runs retain six Linux device-test legs, with apps on four of them; full sweeps retain fourteen legs per platform.

The macOS tradeoff remains explicit: builds on every PR, device tests nightly or with ci-full-jdk-sweep. Metal therefore has no default per-PR runtime coverage. I have not triggered or claimed a live CI run. Approving the updated head.

Issue #1105, opened 2026-09-18
CUDA Tile: FP8 elementwise and reduction ops fail to compile — no arithmetic operators for FP8 tile element types


mikepapadim:
## Summary

On `develop` at `8d592d6fbaafe42d66e1e22c27057cb1cdcf9097`, seven FP8 tests in
`TestTileOpLevel` fail on sm_120 because **CUDA Tile C++ 13.4 defines no arithmetic
operators for FP8 tile element types**, and `CUDATileStmt` emits the operator form
directly.

This is *not* the documented "fp8 tiles below compute capability 9.0" limitation
(`docs/source/tile-api.rst`) — that one is a `tileiras` rejection with
`unsupported type 'f8E4M3FN'`. The GPU here is CC 12.0, well above the gate, and the failure
is a C++ overload-resolution error at the nvcc stage.

## Failing tests

```
uk.ac.manchester.tornado.unittests.tile.TestTileOpLevel
  testFp8Add  testFp8Sub  testFp8Mul  testFp8Div  testFp8Maximum  testFp8Exp  testFp8Sum
```

Everything else in the tile suite is clean: **212 of 219 tests pass**, with all 7 failures in
this one class. `TestTileMatmul`, `TestTileGemmVariants`, `TestTileAttention`,
`TestTileAttentionVariants`, `TestTileRowKernels`, `TestTileMasking`, `TestTileAtomics`,
`TestTileDTypes`, `TestTileRecurrence` and the rest are 0-failure.

Each failure surfaces as:

```
Bailout is disabled.
Reason: Unable to compile task op.k - fp8Add
The internal error is: [ERROR] The CUDA Tile compiler failed.
```

## Minimal reproduction, without TornadoVM

The generated translation unit is what nvcc rejects, so this reduces to plain CUDA Tile C++:

```cpp
#include <cuda_fp8.h>
#include "cuda_tile.h"
namespace ct = cuda::tiles;
using namespace ct::literals;

extern "C" __tile_global__ void fp8Add(__nv_fp8_e4m3 *a, __nv_fp8_e4m3 *b,
                                       __nv_fp8_e4m3 *c, int n) {
    auto av = ct::partition_view{ct::tensor_span{a, ct::extents{n}}, ct::shape<128>{}};
    auto bv = ct::partition_view{ct::tensor_span{b, ct::extents{n}}, ct::shape<128>{}};
[...]

PR #1108, opened 2026-09-18
[cuda] Compile FP8 tile arithmetic through FP32

orionpapadakis:
#### Description

FP8 CUDA Tile arithmetic now compiles by converting operands to FP32, performing the operation, and converting the result back to the original FP8 format. This covers binary arithmetic, math calls, reductions/scans, scaling and FMA for E4M3 and E5M2.

Conversion back happens at each operation boundary, preserving the API's result type and rounding in chained or loop-carried computations. Reductions/scans use FP32 intermediates and convert their final outputs to FP8. Callers wanting FP32 results across operations can explicitly cast their input tiles first. Loads, stores, shape operations and matrix accumulation retain their existing behavior.

Adds exact rounding tests for both FP8 formats and registers them in `tornado-test`. Coverage includes direct FP8 stores, signed/subnormal inputs, halfway rounding, reductions/scans, and runtime-controlled loops with 0, 1 and 3 iterations. The expected results use an independent nearest-even reference over representable FP8 values. Documents the arithmetic precision contract.

Fixes #1105.

#### Problem description

CUDA Tile C++ 13.4 has no direct FP8 arithmetic overloads. `TestTileOpLevel` fails to compile FP8 add, subtract, multiply, divide, maximum, exponential and sum kernels even on sm_120. Emitting the same operator/call with FP32 operands resolves overload selection; converting back preserves the FP8 tile type expected by stores and loop declarations.

#### Backend/s tested

- [ ] OpenCL
- [x] CUDA
- [ ] Metal

#### OS tested

- [x] Linux
- [ ] OSx
- [ ] Windows

Use a CUDA Tile-capable toolchain and driver, with compute capability 9.0 or newer for FP8:

```bash
make jdk21 BACKEND=cuda
source setvars.sh
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileOpLevel
tornado-test -V uk.ac.manchester.tornado.unittests.tile.TestTileFP8Arithmetic
```

Validation on NVIDIA GeForce RTX 5090 Laptop GPU (sm_120), driver 580.142, CUDA Tile/nvcc 13.4.92 and JDK 21.0.2:

- Before: the seven reported tests and both new format regressions fail compilation (9 failures among 108 tests).
- After: all **230 tests across 20 registered Tile suites pass**, including the original failures and exact FP8 regressions.
- Checkstyle passes for the changed compiler and test files; `git diff --check` passes.

GPU validation used the freshly compiled develop CUDA backend plus this patch and test classes in an isolated JDK 21 SDK. This was not a full SDK rebuild. No OpenCL or Metal runtime testing was needed for this CUDA-only emitter change.


mikepapadim (approved):

mikepapadim (commented):
test is not added in the tornado-test.py

